PR #6, opened 2021-02-01
Prepare testing for Ubuntu flavors by adding some mate images

ItzSwirlz:
I didn't do rst... I think we may be able to use the same URL. If not, something to come back to later


oSoMoN:
Thanks for this @ItzSwirlz. The assets look good to my untrained eye, and the location makes sense, but I'm not sure how to test loading the "mate" variant. According to [the flutter documentation](https://flutter.dev/docs/development/ui/assets-and-images#asset-variants), at the moment only resolution-aware variants are taken into account:

    « Flutter uses asset variants when choosing resolution-appropriate images.
      In the future, this mechanism might be extended to include variants for
      different locales or regions, reading directions, and so on. »

ItzSwirlz:
Assuming we don't want to have massive patches, I was thinking this, which I don't know Dart/Flutter well enough to do:

1) Establish a connection between OS and Installer
2) If DE == mate, load mate assets

oSoMoN:
That's sensible, and for ease of testing this should probably check the value of an environment variable to determine the DE (maybe `$XDG_CURRENT_DESKTOP`), so it can easily be overridden.
It's unclear to me whether flutter currently supports specifying a preferred variant though, it will require some investigation.

jpnurmi:
Perhaps we could abuse Flutter's asset selection from packages? If we bundle DE-specific images into separate packages, we should be able to load them like this:

```dart
Column(
  children: [
    Image.asset('assets/foo.png', package: 'mate_icons'),
    SvgPicture.asset('assets/bar.svg', package: 'mate_icons'),
  ],
),
```

https://api.flutter.dev/flutter/painting/AssetImage-class.html#:~:text=_imageInfo%3F.scale%20%3F%3F%201.0%2C%0A%20%20%20%20)%3B%0A%20%20%7D%0A%7D-,Assets%20in%20packages,-To%20fetch%20an

ItzSwirlz:
Sorry I didn't see that. I have lots of other things to do with school, and plus this PR is old. Let's get the core Ubuntu down before being fancy :)

jpnurmi:
@ItzSwirlz There are still bits and pieces missing but preliminary support was added in https://github.com/canonical/ubuntu-desktop-installer/pull/478.

jpnurmi:
Also, https://github.com/canonical/ubuntu-desktop-installer/pull/485 added `RELEASE` variables instead of hardcoding "Ubuntu" in translation strings. The next step is to pass `Flavor.of(context).name` to fill the variables with the correct flavor name.

PR #14, opened 2021-02-02
i18n: Initial pt_BR translations

Lennoard:
Although [mentioned here](https://flutter.dev/docs/development/accessibility-and-localization/internationalization) I could not find a way to specify a country code when it comes to the `.arb` file. I hope the current set up is enough.


GiovanniSM20:
Dude try to add `console:pt:Português Brasil` on **languagelist** inside this path: `ubuntu-desktop-installer/subiquity_client/subiquity`

oSoMoN:
> Although [mentioned here](https://flutter.dev/docs/development/accessibility-and-localization/internationalization) I could not find a way to specify a country code when it comes to the `.arb` file. I hope the current set up is enough.

This requires adding a ​"@@locale"​ entry to your arb file (the value would be "pt_BR"). However I'm not sure this will work as expected because of the way `UbuntuLocalizations.supportedLocales` is defined (it's a list of language codes instead of a list of locales).

GiovanniSM20:
could someone close this pr because it has already been accepted...

oSoMoN:
Not sure what happened to the source branch of this PR, I tried pushing changes to update it after #24 landed, but the remote rejected them.
I went ahead and merged this PR + my changes into main. Thanks @Lennoard for the translations!

PR #15, opened 2021-02-03
Added cla-check, copied from snapcraft

kenvandine:


tools/cla_check.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
```
MarcusTomlinson, line 1: Putting this in a `tools` directory at the root implies it's a tool for development when it's really only for our GitHub Actions.

How about we put this in `.github/workflows/tools/cla_check.py` rather.

MarcusTomlinson (changes requested):

tools/cla_check.py
kenvandine: Makes sense, i'll make it so!

MarcusTomlinson (approved):
Love it! Thanks Ken :)

PR #21, opened 2021-02-06
Added navigator.pop

Abhishek01039:


oSoMoN:
There is no page to pop here, the empty body for the function was unintentionally added in revision 53e96278b02281439b402a6d8b05852595594d45, I reverted it in revision 220bdff6e4b61ec4def92f1b3ebc9f2ef15c47f6.

Issue #8, opened 2021-02-02
Contribution Guide!


ArishSultan:
Hi, I want to contribute to this project, is there any guide or a roadmap?

kenvandine:
Right now the designs aren't in a public place, we'll work to get them published in a place that makes it easier to collaborate. Your contributions are welcome, and appreciated, just give us a little time to get prepared.

iapicca:
@kenvandine 
I'd like to sign the [cla](https://ubuntu.com/legal/contributors/agreement) but I'm not sure
what to insert at `Please add the Canonical Project Manager or contact`

kenvandine:
> @kenvandine
> I'd like to sign the [cla](https://ubuntu.com/legal/contributors/agreement) but I'm not sure
> what to insert at `Please add the Canonical Project Manager or contact`

You can reference me as your contact.  Thanks!

ArishSultan:
The UI designs provided in the latest issues (all issues marked as specification).

Are those final or still being worked on? if yes than, when will the development start?

MarcusTomlinson:
@ArishSultan those designs may be tweaked a bit here and there, but for the most part they are final yes. Development has started.

ArishSultan:
Is there anything that i can work on?

oSoMoN:
I added contributor guidelines in https://github.com/canonical/ubuntu-desktop-installer/pull/60.

oSoMoN:
> Is there anything that i can work on?

To answer your question Arish, yes, there's a lot that you can work on, contributions very welcome!
UI and UX specifications are published as [Github issues](https://github.com/canonical/ubuntu-desktop-installer/issues?q=is%3Aissue+label%3Aspecification+sort%3Acreated-asc). Any specification issue without an assignee is up for grabs, if you start work on one of them please make sure to comment on the issue to say so, a project administrator will assign it to you.

oSoMoN:
[Contributor guidelines](https://github.com/canonical/ubuntu-desktop-installer/blob/main/CONTRIBUTING.md) are now merged.

PR #24, opened 2021-02-15
Localization upgrade

ArishSultan:
Implement new Localization pattern as described in version 1.22


ArishSultan:
How to fix the cla issue?

iapicca:
@ArishSultan 
https://github.com/canonical/ubuntu-desktop-installer/issues/8#issuecomment-773510447

ArishSultan:
I have signed the CLA, should I add a new committ?

MarcusTomlinson:
> I have signed the CLA, should I add a new committ?

Hi @ArishSultan, unfortunately our CLA check only works for those who've signed the CLA via Launchpad, and not GitHub as yet.

I've checked with Legal and they've confirmed you have signed the CLA, you (and whoever eventually reviews this) may ignore the cla-check status.

oSoMoN:
Some functional regressions I noted when testing your branch:

 - On the welcome screen, the current locale is not selected by default (I use French on my desktop, if the French translations are available they should be loaded and the corresponding entry in the language list should be selected)
 
 - The "You may wish to read the release notes." string doesn't appear translated after manually selecting a language for which it is translated
 
- On the keyboard layout selection screen, the language for the current locale should be selected by default, if available

ArishSultan:
Ok, i will fix these

MarcusTomlinson:
> > I have signed the CLA, should I add a new committ?
> 
> Hi @ArishSultan, unfortunately our CLA check only works for those who've signed the CLA via Launchpad, and not GitHub as yet.
> 
> I've checked with Legal and they've confirmed you have signed the CLA, you (and whoever eventually reviews this) may ignore the cla-check status.

Mejans:
Hello,

why would the Occitan be ignored o_Ô ?

ArishSultan:
Its becuase the global material localization has not implementation for occitan,

You can see from the doxumwntation that the material package has implementation for 70 languages occitan is not included in those languages,

So, i ignored it for some time until someone provides its complete transalation.

oSoMoN:
> Its becuase the global material localization has no implementation for occitan,

This shouldn't matter, we're not using Material translations in the app, are we? We shouldn't be tied to Flutter's list of translations. Ubuntu as a whole has translations for [244 languages](https://translations.launchpad.net/ubuntu/+translations) (not all of them are complete or of high quality though).

Can we make sure that all the existing translations are usable?

Mejans:
From what I saw, Material translations is only about dates, months, and days things...
Doesn't seem to be used here yeah.

ArishSultan:
Material Translation is used in whole application, most of the widgets inside the `material.dart` depend upon those translations i.e. AppBar.

For now I have implemented a dummy translation which will use `en` for all unknown translations.

ArishSultan:
Why is this `subiquity_client` test failing?

Mejans:
Thanks @ArishSultan it's appreciated

ArishSultan:
anybody here?

MarcusTomlinson:
> anybody here?

Sure. Don’t know why it’s failing. Hmm. I just ran CI against main again and passed fine. I’ll kick your build again and see if that resolves it. Otherwise I’ll have a look in Monday.

MarcusTomlinson:
Subiquity tests are green again - interesting. Looks like there’s something up with the installer tests now though. Looks like a legit issue at least.

ArishSultan:
Idk, test is passing on my side.

ubuntu_desktop_installer/lib/welcomepage.dart
```diff
@@ -35,10 +35,10 @@ class _WelcomePageState extends State<WelcomePage> {
     _languageListScrollController = AutoScrollController();
 
     widget.client
-        .fetchKeyboardLayouts(kbdAssetName, Locale(Intl.defaultLocale));
+        .fetchKeyboardLayouts(kbdAssetName, UbuntuDesktopInstallerApp.locale);
     final locale = Intl.defaultLocale;
     for (var i = 0; i < widget.client.languagelist.length; ++i) {
-      if (widget.client.languagelist[i].item1.languageCode == locale) {
+      if (locale.contains(widget.client.languagelist[i].item1.languageCode)) {
```
MarcusTomlinson, line 41: This is where the test is failing (fails locally for me as well). It appears that `locale` is null at this point.

MarcusTomlinson (changes requested):

ubuntu_desktop_installer/lib/welcomepage.dart
ArishSultan: I figured it out, `setupAppLocalizations` was not called in test initializations.

So, the `Intl.defaultLocale` gave a null value which lead to the `locale` variable being null too.

oSoMoN:
The "translations" section of README.md probably needs to be updated to reflect the new workflow.

oSoMoN:
I'm seeing a couple of functional regressions:
 - when I start the app with French as the default locale, French is correctly selected by default in the language list, but the French translations aren't loaded (the UI is in English)
 - the continue button on the "try or install" screen should be disabled by default, until one of the three options is selected

oSoMoN:
Note: I just pushed a couple of commits that address the two functional regressions I was mentioning in my previous comment. I'm continuing with the review.

ubuntu_desktop_installer/lib/l10n/app_localizations.dart
```diff
@@ -0,0 +1,327 @@
...
+        shortMonthDayFormat: DateFormat('MMM d'),
+      ),
+    );
+  }
+
+  @override
+  bool shouldReload(_UbuntuLocalizationsDelegate old) => false;
+}
+
+class GlobalUbuntuLocalizations extends MaterialLocalizationEn {
```
oSoMoN, line 300: This approach doesn't look right: some code appears to be copy/pasted from an implementation of the 'nn' locale (Norwegian Nynorsk), and extending `MaterialLocalizationEn` [isn't recommended](https://api.flutter.dev/flutter/flutter_localizations/GlobalMaterialLocalizations-class.html#overriding-translations).

For occitan support, maybe we can extend [`MaterialLocalizationCa`](https://api.flutter.dev/flutter/flutter_localizations/MaterialLocalizationCa-class.html)? @Mejans what do you think? Would that be close enough with regards to date/time and number formats?

oSoMoN (changes requested):

ubuntu_desktop_installer/lib/l10n/app_localizations.dart
Mejans: yes, linguistically one language, politically two :) So yes it would be fine.

ArishSultan: No problem, As I said it was a dummy implementation (a quick fix).

MarcusTomlinson:
So sorry for the CI noise. The CLA check was going bonkers. Looks to be fixed now.

oSoMoN:
> The "translations" section of README.md probably needs to be updated to reflect the new workflow.

While doing that, I've been wondering how to automate the extraction of strings from source code to the translation template. I haven't seen anything related to that mentioned in the documentation. Not a blocker, so I'm going to go ahead and merge this PR now, but it would be a nice improvement for a future PR.

PR #56, opened 2021-03-21
Follow light vs. dark system theme

jpnurmi:
| Before | After, light | After, dark |
|---|---|---|
| Screenshot from 2021-03-21 19-09-00 | Screenshot from 2021-03-21 19-11-08 | Screenshot from 2021-03-21 19-12-42 |
| Screenshot from 2021-03-21 19-09-07 | Screenshot from 2021-03-21 19-11-12 | Screenshot from 2021-03-21 19-12-46 |
| Screenshot from 2021-03-21 19-09-38 | Screenshot from 2021-03-21 19-11-17 | Screenshot from 2021-03-21 19-12-52 |
| Screenshot from 2021-03-21 19-09-41 | Screenshot from 2021-03-21 19-11-40 | Screenshot from 2021-03-21 19-12-56 |

### DIFF
--- a/ubuntu_desktop_installer/lib/main.dart
+++ b/ubuntu_desktop_installer/lib/main.dart
@@ -42,6 +42,8 @@ class UbuntuDesktopInstallerApp extends StatelessWidget {
         locale: value,
         onGenerateTitle: (context) => AppLocalizations.of(context).appTitle,
         theme: yaruLightTheme,
+        darkTheme: yaruDarkTheme,
+        themeMode: ThemeMode.system,
         debugShowCheckedModeBanner: false,
         localizationsDelegates: [
           ...AppLocalizations.localizationsDelegates,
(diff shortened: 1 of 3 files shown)


jpnurmi:
There's a tiny difference in the option card's body text color on the "try or install" page:
image

The reason I used translucency instead is that the primary variant color didn't work in dark theme:
Screenshot from 2021-03-21 19-41-37

ubuntu_desktop_installer/lib/main.dart
```diff
@@ -42,6 +42,8 @@ class UbuntuDesktopInstallerApp extends StatelessWidget {
         locale: value,
         onGenerateTitle: (context) => AppLocalizations.of(context).appTitle,
         theme: yaruLightTheme,
+        darkTheme: yaruDarkTheme,
+        themeMode: ThemeMode.system,
```
oSoMoN, line 46: According to [the documentation](https://api.flutter.dev/flutter/material/MaterialApp/themeMode.html), `ThemeMode.system` is the default value, so it's probably not necessary to set it here?

oSoMoN (commented):
LGTM, thanks for this.
Just one minor nit, see inline.

ubuntu_desktop_installer/lib/main.dart
jpnurmi: Thanks, good catch :)

PR #57, opened 2021-03-26
import yaru as yaru and change theme names

Feichtmeier:
Due to a recent quiet [big change in the yaru theme organisation](https://github.com/ubuntu/yaru.dart/pull/29) it is needed for the ubuntu installer to change the theme name vars which is done with this PR.

Tested locally:

https://user-images.githubusercontent.com/15329494/112620834-4fe2d080-8e29-11eb-9217-c981ca6f9b12.mp4

But a review is much appreciated!

CC @MarcusTomlinson


MarcusTomlinson:
Thanks!

PR #61, opened 2021-04-07
Move the keyboard model and associated logic to a separate class,

oSoMoN:
outside of SubiquityClient (where it didn't belong in the first place).

Implement default keyboard layout/variant selection using debconf.


oSoMoN:
Note that this isn't implementation-complete, but is already an improvement over the current implementation (and contributes to cleaning up `subiquity_client`), so I think this is worth merging as-is while I continue working on the keyboard layout selection screen.

jpnurmi:
Nice to see the remaining `dart:ui` dependencies gone from `subiquity_client`. Now it could be made a pure Dart package. :wink: 

I had been kind of eyeing the same earlier. I ended up with [`keyboard_info`](https://pub.dev/packages/keyboard_info) that I was planning to propose for querying the keyboard layout & variant. It's querying the info using GSettings and/or various files depending on the desktop environment.

jpnurmi:
Hey, does debconf return the original installation language? How does it compare to `/etc/default/keyboard`?

P.S. I left a very confusing message the other night that I decided to delete the next morning. :smile: What I meant to say is that I had been briefly looking at the same area, but mostly focusing on the keyboard layout and variant querying. I had ended up with [keyboard_info](https://pub.dev/packages/keyboard_info), which I had been testing on a bunch of DE's and the results were pretty promising. On all DE's that store the current KB layout into Dconf, the currently selected KB layout could be fetched via GSettings. The only exception was KDE that doesn't seem to store the currently selected KB layout anywhere accessible right away but only at logout time. Nevertheless, it would still result in the correct KB layout for most users that don't constantly switch between KB layouts. Finally, the installation time KB layout and variant stored in `/etc/default/keyboard` would be used as a fallback if none of the other methods above succeed.

MarcusTomlinson:
> Note that this isn't implementation-complete, but is already an improvement over the current implementation (and contributes to cleaning up `subiquity_client`), so I think this is worth merging as-is while I continue working on the keyboard layout selection screen.

@oSoMoN sure, looks good to me, let's merge. In case you missed it though, @jpnurmi had some interesting comments above. Nothing to hold up the PR for, but worth pondering.

oSoMoN:
@jpnurmi @MarcusTomlinson I didn't miss the comments (I even read and was going to reply to the original message that was later deleted). [keyboard_info](https://pub.dev/packages/keyboard_info) looks great and I'd love to see it used here, PR welcome!

My initial implementation calls into debconf because that's what ubiquity does [in the `console_setup` plugin](https://git.launchpad.net/ubiquity/tree/ubiquity/plugins/ubi-console-setup.py#n493). And I assumed that was the correct way to get the default layout and variant for the selected language in the context of the installer being run from an ISO, without an existing desktop environment (in this case ubiquity deletes `/etc/default/keyboard`, so there's not even that fallback option). However after a few tests in a VM I'm not sure this is working as intended, and it's only one side of the installer story − when run from an existing DE we definitely want to detect the current layout and variant by querying its settings.

PR #65, opened 2021-04-15
Use Provider for providing SubiquityClient

jpnurmi:
Nice for lazy initialization and to avoid having to pass the client instance down to constructors.


oSoMoN:
Very nice, thanks!

Issue #30, opened 2021-03-08
Screen 5 - Connect to Internet


seb128:
The design for screen n°5, Connect to Internet

![internet](https://people.canonical.com/~platform/desktop/installer/5.%20Connect%20to%20Internet/preview/ubuntu-flutter-installer---material-icons-connect-to-internet---drivers-installed---wi-fi-selected---multi-adapter---expanded.png)

Design details https://people.canonical.com/~platform/desktop/installer/5.%20Connect%20to%20Internet/

fitojb:
The density of the network list could be improved (well, the Yaru theme should be a lot denser than it is). I don’t think it’s terribly useful to display only one and a half items at any time inside that scrollable box.

Mannshoch:
My W-Lan Passkey is way too long to enter it with Keyboard. I like the QR-code solution like I use also on my Mobil phone.
![grafik](https://user-images.githubusercontent.com/7795891/111597517-86d43900-87ce-11eb-9e21-51640b93c673.png)

elbeicktalat:
@Mannshoch 
Hi bro,
it will be so helpful, but how you can scan the QR code?

Mannshoch:
On my laptop I have a Webcam included. On my desktop I attached an USB Webcam. (I use it in Homeoffice)

May you wish a creativ solution: https://randomblog.hu/using-your-android-phone-as-a-webcam-for-ubuntu-20-04-with-droidcam/
And if you wish a strange solution: Use KDEconnect and aus a Mobilephone cam. (but you need a kdeconnect on phone and also pair with it)

jpnurmi:
Hi, feel free to assign this to me if nobody is working on it yet.

jpnurmi:
Since the available vertical space for the Wi-Fi network list is limited, would it be a completely crazy idea to make the whole page scrollable, and expand the current connection type selection like this?

![connect-to-internet](https://user-images.githubusercontent.com/140617/114602105-85474500-9c96-11eb-9e44-a9e435a9b581.gif)

Feichtmeier:
I have the feeling that the whole UI is somehow crowded on the left side of the screen.
I thought about maybe using rather a Tabbed UI here instead of the radio button selection expanding those panels
@jpnurmi

MarcusTomlinson:
> Since the available vertical space for the Wi-Fi network list is limited, would it be a completely crazy idea to make the whole page scrollable, and expand the current connection type selection like this?
> 
[...]

PR #68, opened 2021-04-21
Project structure proposal

jpnurmi:
The proposal looks like this:
```
packages
├── subiquity_client
└── ubuntu_desktop_installer
    └── lib
        ├── app.dart
        ├── l10n
        ├── main.dart
        ├── pages
        |   └── foo_page.dart
        ├── routes.dart
        ├── widgets
        |   └── foo_view.dart
        └── widgets.dart
```
In the future, one would likely end up having typical `models/` and `services/` subdirectories too. Furthermore, more complex pages could also be organized into separate subdirectories once they start having view models and other page-specific files.

This change is partly motivated by #30 and would help to organize the files. :slightly_smiling_face:

P.S. `keyboard_model.dart` looks like a potential candidate for a small separate package so I left it in place for now.


oSoMoN:
README.md also needs to be updated to reflect the new path to `subiquity_client`. This otherwise looks good to me.

jpnurmi:
In the end, the structure is very much a matter of taste, of course, and the internets are full of various Flutter project structuring tips and opinions. :slightly_smiling_face: @ArishSultan, do you have any thoughts on this? This change would affect your ongoing tasks to some degree.

From what I've seen, the most common structures are either a) grouping by modules/features, or b) grouping by types (all services together, all repositories together, etc.). This proposal is sort of a mixture of the two in the sense that screens/pages would be independent modules that would not be aware of each other, but only known on the application level where the routes are defined and navigation is handled. And any data models and services that would be used across the modules would be grouped on the application level, whereas any module-specific models/services/repositories etc. would be inside the respective module.

Feichtmeier:
Even if I haven't be asked: I think grouping by types makes sense if you have such things as model classes, service classes and view classes. The view dir could be further divided into pages where you really route with the navigator to and small views which are put into the widget tree via bottom navbar / rails.

Feichtmeier:
Yeah :) we are on the same line. I didn't feel excluded, it was a bit humor, and I should improve at this field 👍 😆 
Breaking up the pages creates a lot diff, so it's understandable that you did not make this all at once

PR #70, opened 2021-04-22
Add CI support for code coverage.

oSoMoN:


oSoMoN:
There's some missing setup required by codecov.io to integrate with github CI, which requires action from an owner of the repository. I'm guessing this is the reason why the CI check currently doesn't show up here?

.github/workflows/flutter-ci.yml
```diff
@@ -38,7 +38,13 @@ jobs:
     
     - name: Run tests on subiquity_client
       working-directory: ./packages/subiquity_client
-      run: dart test
+      run: flutter test --coverage
```
MarcusTomlinson, line 41: This looks like a mistake. Should it not be `dart test --coverage`?

MarcusTomlinson (changes requested):

MarcusTomlinson:
I assume uploading multiple sets of coverage in CI is common practice? I don’t know enough about this. Does the report get generated as a combination of the two coverage results?

oSoMoN:
> I assume uploading multiple sets of coverage in CI is common practice? I don’t know enough about this. Does the report get generated as a combination of the two coverage results?

I lack experience with this, but [the documentation](https://docs.codecov.io/docs/merging-reports) seems to imply that reports will be merged automatically. Let's see if it happens in practice…

MarcusTomlinson:
Ok looks alright then. Let’s see how it goes.

PR #77, opened 2021-04-29
Screen 5 - Connect to Internet

jpnurmi:
An initial version without hidden Wi-Fi.

Ref: #30


jpnurmi:
Note: this is still a draft, but early feedback is already welcome. Also, the PR grew quite big so if things look good in general, I will start splitting off parts and send separate smaller PRs paired with some tests as appropriate. :smile:

jpnurmi:
Looks like there's currently a little formatting difference between Flutter master & dev, but it has been already reverted:
- change: https://github.com/dart-lang/dart_style/pull/1012
- revert: https://github.com/dart-lang/dart_style/pull/1019

```diff
diff --git a/packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/connect_to_internet_page.dart b/packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/connect_to_internet_page.dart
index d87a8ed..4cf612a 100644
--- a/packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/connect_to_internet_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/connect_to_internet_page.dart
@@ -200,8 +200,8 @@ class _ConnectToInternetPageState extends State<ConnectToInternetPage> {
                           onPressed: model.canContinue
                               ? _continue
                               : model.canConnect(name.text)
-                              ? _connect
-                              : null,
+                                  ? _connect
+                                  : null,
                         );
                       },
                     ),
```

oSoMoN:
That looks very promising, thanks for working on this!

Instead of fake devices and access points (which would still be useful as mocks for widget tests), can you try and hook it up to the [nm](https://pub.dev/packages/nm) package, and see if it works as expected?

+1 for smaller PRs + tests 😉

jpnurmi:
I tried `nm` earlier with a small separate test app earlier. Looks otherwise promising but it's still missing some features like adding connections.

oSoMoN:
> I tried `nm` with a small separate test app earlier. Looks otherwise good but it's still missing some features like adding connections.

Mind [filing issues](https://github.com/canonical/nm.dart/issues) for the missing features?

jpnurmi:
> Mind [filing issues](https://github.com/canonical/nm.dart/issues) for the missing features?

Apparently, addConnection() was already added: https://github.com/canonical/nm.dart/pull/24 :tada:

jpnurmi:
Note to self: radio tiles' content padding went missing while simplifying things :smile:

oSoMoN:
There are some trivial conflicts to resolve in `packages/ubuntu_desktop_installer/pubspec.yaml`.

Some functional considerations:
 - no option is initially selected on the screen, we should probably try to select a default if available (i.e. if there is an ethernet adapter, select "use wired connection" by default, otherwise if there is a wifi adapter and at least one network is available, select it, and if not default to "I don't want to connect to the internet just now")
 - If I turn off wifi on my device, and then select the second option, the name of the wifi adapter is shown but there is no list of network under it. Maybe it should be greyed out to make it clear that it's not usable until turned on?
 - With my wifi adapter turned off, I'm seeing the following error: `Unhandled Exception: org.freedesktop.NetworkManager.Device.NotAllowed: Scanning not allowed while unavailable` with a stack trace. This could be logged as an info/warning, without the stack trace, maybe?
 - When turning on and off the wifi adapter while on that screen, I'm seeing a bunch of `Another exception was thrown: A WifiModel was used after being disposed` being logged.
 - If I enter a wrong password for an existing wifi network, I am getting an error and stack trace, but no feedback in the UI other than not moving on to the next page. The network I selected remains selected, giving me the false impression that I am connected.
 - There's an encoding problem with network names, my neighbour has an access point with the "é" character in its name and it isn't being displayed correctly (it shows "Ã©" instead)

Feichtmeier:
@jpnurmi you could also exchange

https://github.com/canonical/ubuntu-desktop-installer/blob/9e8b53ade4530a997a1f46c91a48c1967e9b9364/packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_view.dart#L35

with 

`YaruIcons.window_close`
grafik

and
https://github.com/canonical/ubuntu-desktop-installer/blob/9e8b53ade4530a997a1f46c91a48c1967e9b9364/packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_view.dart#L173

with 
`YaruIcons.refresh`
grafik

The check iconlooks a bit better with the material icons than the `YaruIcons.ok_filled` and `YaruIcons.ok`
grafik

@Jupi007 could you add a regular check mark without the circle or a box? Similar to
grafik

Jupi007:
> @jpnurmi you could also exchange
> 
> https://github.com/canonical/ubuntu-desktop-installer/blob/9e8b53ade4530a997a1f46c91a48c1967e9b9364/packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_view.dart#L35
> 
> with
> 
> `YaruIcons.window_close` grafik
> 
> and
> 
> https://github.com/canonical/ubuntu-desktop-installer/blob/9e8b53ade4530a997a1f46c91a48c1967e9b9364/packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_view.dart#L173
> 
> with `YaruIcons.refresh` grafik
> 
> The check iconlooks a bit better with the material icons than the `YaruIcons.ok_filled` and `YaruIcons.ok` grafik
> 
> @Jupi007 could you add a regular check mark without the circle or a box? Similar to grafik

Of course :)

jpnurmi:
Thanks! I'll go through the icons.

Jupi007:
@jpnurmi I just published the `0.0.8` version of `yaru_icons` with new _ok_simple_ icon :)

jpnurmi:
My notes claim that I've marked this as ready for review 2 weeks ago. I guess I just forgot to press the button. :)

Feichtmeier:
Design nit-pick:

I think you could reduce the contrast between unselected wifi tiles and the refresh IconButton here

grafik

https://github.com/canonical/ubuntu-desktop-installer/blob/1488b1f48dc7994b754436f11e243cf0e6f16435/packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_view.dart#L174

Maybe reduce the strength of the black refresh button a bit 

:)

jpnurmi:
Thanks for testing! It was the previous NM lazy-init commit that broke NM changes.

It's a bit confusing but apparently one cannot subscribe to NetworkManagerClient.propertiesChanged before NetworkManagerClient.connect() has returned, because it returns a different empty stream instance until connected.

jpnurmi:
@Feichtmeier Thanks for the feedback! I removed the button because there's no such thing in the design. I used it for manual testing and forgot it there...

oSoMoN:
Thanks for the prompt fix, I can confirm this use case now works as expected.
I'm currently reviewing the code, so far so good, but I still have a good chunk to read yet.

packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_model.dart
```diff
@@ -0,0 +1,314 @@
...
+import 'package:collection/collection.dart';
+import 'package:dbus/dbus.dart';
+import 'package:flutter/material.dart';
+
+import '../../services.dart';
+import 'connect_model.dart';
+import 'network_device.dart';
+import 'property_stream_notifier.dart';
+
+// TODO: appropriate timeout?
```
oSoMoN, line 13: If it's any reference, gnome-control-center's Wi-Fi panel triggers a periodic scan every 15 seconds, which lasts for 3 seconds.

packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_model.dart
```diff
@@ -0,0 +1,314 @@
...
+    }
+    _devices = null;
+  }
+
+  void _updateDevices() {
+    _resetDevices();
+    notifyListeners();
+  }
+
+  WifiDevice? _selected;
```
oSoMoN, line 112: There are a bunch of references to `_selected` in getters and methods which don't change its value, for consistency and improved readability they should probably be replaced by `selectedDevice`.

packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_model.dart
```diff
@@ -0,0 +1,314 @@
...
+  WifiDevice? get selectedDevice => _selected;
+  bool isSelectedDevice(WifiDevice device) => device == _selected;
+  void selectDevice(WifiDevice? device) {
```
oSoMoN, line 115: Should there be a check here to make the method a no-op if `device == selectedDevice` ?

packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_model.dart
```diff
@@ -0,0 +1,314 @@
...
+
+class WifiModel extends PropertyStreamNotifier implements ConnectModel {
+  WifiModel(this._service, [this._udev]) {
+    addProperties(_service.propertiesChanged);
+    addPropertyListener('Devices', _updateDevices);
+    addPropertyListener('WirelessEnabled', notifyListeners);
+  }
+
+  @override
+  bool get canConnect => _selected?.canConnect == false;
```
oSoMoN, line 21: Is this right?

packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_model.dart
```diff
@@ -0,0 +1,314 @@
...
+    return accessPoints.firstWhereOrNull((ap) {
+      return ap.accessPoint == _wireless.activeAccessPoint;
+    });
+  }
+
+  bool isActiveAccessPoint(AccessPoint accessPoint) {
+    return accessPoint._accessPoint == _wireless.activeAccessPoint;
+  }
+
+  AccessPoint? _selected;
```
oSoMoN, line 219: Same remark as for `WifiModel`: there are a bunch of references to `_selected` in getters and methods which don't change its value, for consistency and improved readability they should probably be replaced by `selectedAccessPoint`.

packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_model.dart
```diff
@@ -0,0 +1,314 @@
...
+
+  final NetworkManagerDeviceWireless _wireless;
+
+  List<AccessPoint>? _accessPoints;
+  List<AccessPoint> get accessPoints => _accessPoints ??= _getAccessPoints();
+
+  List<AccessPoint> _getAccessPoints() {
+    final ssids = <List<int>>[];
+
+    bool hasSsid(List<int> ssid) {
```
oSoMoN, line 174: That's probably a change for the `nm` package, but I'll suggest it here while I'm at it: would it make sense to have a `SSID` type/class to hide away implementation details (like comparing two lists of integers, or converting from/to a string)?

packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_model.dart
```diff
@@ -0,0 +1,314 @@
...
+    return accessPoint._accessPoint == _selected?._accessPoint;
+  }
+
+  void selectAccessPoint(AccessPoint? accessPoint) {
+    _selected = accessPoint;
+    notifyListeners();
+  }
+
+  Future<List<int>?> getSsid(
+      NetworkManagerSettingsConnection connection) async {
```
oSoMoN, line 231: Does this method need to be public?

packages/ubuntu_desktop_installer/test/connect_to_internet_page_test.dart
```diff
@@ -0,0 +1,193 @@
...
+    final wifiDevice = MockWifiDevice();
+    when(wifiDevice.model).thenReturn('model');
+    when(wifiDevice.isAvailable).thenReturn(true);
+    when(wifiDevice.isBusy).thenReturn(false);
+    when(wifiDevice.scanning).thenReturn(false);
+    when(wifiDevice.accessPoints).thenReturn([]);
+
+    final wifiModel = MockWifiModel();
+    when(wifiModel.connectMode).thenReturn(ConnectMode.wifi);
+    when(wifiModel.devices).thenReturn([]);
```
oSoMoN, line 60: `wifiModel.devices` is later mocked to return `[wifiDevice]`, so this line is useless, right?

packages/ubuntu_desktop_installer/test/connect_to_internet_page_test.dart
```diff
@@ -0,0 +1,193 @@
...
+    );
+
+    final page = find.byType(ConnectToInternetPage);
+    expect(page, findsOneWidget);
+
+    final context = tester.element(page);
+    expect(
+        Provider.of<ConnectToInternetModel>(context, listen: false), isNotNull);
+    expect(
+        Provider.of<ConnectToInternetModel>(context, listen: false), isNotNull);
```
oSoMoN, line 189: This line is duplicated, I suppose it should test the existence of a `NoConnectModel` instance instead?

packages/ubuntu_desktop_installer/test/ethernet_model_test.dart
```diff
@@ -0,0 +1,114 @@
...
+
+    expect(model.canConnect, isFalse);
+    expect(model.canContinue, isFalse);
+    expect(model.isBusy, isFalse);
+    expect(model.isActive, isFalse);
+
+    expect(model.canConnect, isFalse);
+    expect(model.canContinue, isFalse);
+    expect(model.isBusy, isFalse);
+    expect(model.isActive, isFalse);
```
oSoMoN, line 72: These 4 lines are duplicated.

oSoMoN (commented):
Looks good and works as intended, thanks! See minor comments inline.

packages/ubuntu_desktop_installer/lib/pages/connect_to_internet/wifi_model.dart
jpnurmi: Thanks! I've added the same

oSoMoN:
Just found another functional issue while testing unusual user interaction: if I select an access point that's not my default one, the next button becomes "Connect", and if I then turn off wifi in network manager, I end up with no radio option selected, yet the next button still reads "Connect" and is enabled, and when clicked I get an unhandled exception:

    [ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: org.freedesktop.NetworkManager.Device.SpecificObjectNotFound: The access point /org/freedesktop/NetworkManager/AccessPoint/1 was not in the scan list.
    #0      DBusClient._callMethod (package:dbus/src/dbus_client.dart:1135:9)
    <asynchronous suspension>
    #1      DBusClient.callMethod (package:dbus/src/dbus_client.dart:626:12)
    <asynchronous suspension>
    #2      NetworkManagerClient.addAndActivateConnection (package:nm/src/network_manager_client.dart:2491:18)
    <asynchronous suspension>


Capture d’écran du 2021-12-02 16-44-04

jpnurmi:
I'm still working on the last remaining issue that the buttons sometimes end up in a wrong state.

oSoMoN:
All the changes since my last review look good. I'm still seeing one minor functional issue though: if I enable the wireless connection from the UI, then disable it from the nm applet, I end up in a state where no radio button is selected. I think the selection should default back to the most sensible option, in this case wired connection:

Capture d’écran du 2021-12-07 16-02-51

oSoMoN:
LGTM now, thanks!

oSoMoN (approved):

jpnurmi:
Yay, big thanks!

PR #78, opened 2021-04-30
Split RoundedContainer & RoundedListView

jpnurmi:
Originates from #77

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/widgets/rounded_container.dart
+++ b/packages/ubuntu_desktop_installer/lib/widgets/rounded_container.dart
@@ -1,0 +1,64 @@
+import 'package:flutter/material.dart';
+
+const double _kDefaultBorderWidth = 1.0;
+const double _kDefaultBorderRadius = 5.0;
+
+/// A container with a rounded border.
+class RoundedContainer extends StatelessWidget {
+  /// Creates a container with a rounded border.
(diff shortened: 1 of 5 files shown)


packages/ubuntu_desktop_installer/lib/widgets/rounded_container.dart
```diff
@@ -0,0 +1,64 @@
...
+    this.child,
+    this.width,
+    this.height,
+    this.borderColor,
+    this.borderWidth,
+    this.borderRadius,
+    this.constraints,
+  }) : super(key: key);
+
+  /// See [Container.child]
```
jpnurmi, line 20: I cheated a bit here with the docs. :) Anyway, is the long-term plan to have everything documented, or could `public_member_api_docs` be turned off?

oSoMoN: I think we should try and document all public members where it really adds value (for new contributors and maintainers, present and future). In this case referencing properties in already documented classes seems perfectly reasonable.
And since https://github.com/canonical/ubuntu-desktop-installer/commit/64cde9cbb0982d1a8ad8f31bd8e3759bada806a1 the analyzer job won't fail in CI when encountering undocumented public members.

oSoMoN:
LGTM, thanks!

PR #79, opened 2021-05-01
Move OptionCard to widgets & add tests

jpnurmi:
This change makes OptionCard a standalone and easy test widget. Furthermore, TryOrInstallPage becomes a bit simpler because an inherited widget is no longer necessary.


oSoMoN:
LGTM, thanks!

PR #86, opened 2021-05-06
WelcomePage: add missing setState()

jpnurmi:
Trigger a rebuild when the language list has been (asynchronously) loaded.


oSoMoN:
Good catch, thanks!

PR #87, opened 2021-05-06
Add web support for PR previews etc.

jpnurmi:
With these minor changes + the generated app runner for web, it would be possible to run the installer app in a browser. One can, for example, provide a clickable preview of PR changes, or let someone who doesn't have Flutter installed play with the installer: https://jpnurmi.kapsi.fi/ubuntu-desktop-installer/.

Web is a special platform and web support may easily break (no file system access, FFI, D-Bus, ...). Therefore, it may take a little extra effort to maintain web compatibility in the future, but it's certainly cool and useful to have until it may no longer be feasible. What do you think?

To deploy, run `flutter build web --release` and copy the contents of `build/web/` somewhere. Notice the comment in `index.html` that the base href may need to be adjusted if the web app is placed in a subdirectory.

```
  <!--
    If you are serving your web app in a path other than the root, change the
    href value below to reflect the base path you are serving from.

    The path provided below has to start and end with a slash "/" in order for
    it to work correctly.

    For more details:
    * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base
  -->
  <base href="/">
```


jpnurmi:
I'm already having second thoughts about this. :smile: Presumably, there will be so many Linux/desktop-specific dependencies with indirect web-incompatible dependencies (dart:io, D-Bus, FFI, ...) that maintaining web compatibility is probably not worth it...

PR #89, opened 2021-05-12
Subiquity deps

kenvandine:
Build curtin and probert as parts needed by subiquity and stage some subiquity runtime deps.  Classic fixes from seb128.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -66,7 +66,7 @@ parts:
     override-build: |
       set -eux
       cd packages/ubuntu_desktop_installer
-      flutter channel master
+      flutter channel dev
       flutter upgrade
       flutter config --enable-linux-desktop
       flutter doctor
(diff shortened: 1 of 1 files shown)


seb128:
Great, the snap builds wit those changes is working as expected and not hitting dri drivers errors anymore.

Our project instructions state to use flutter dev and there you change to master, is that an inconsistency we should fix or do we expect to use dev for work and roll out on stable (but then we don't test what we ship to users?)

The flutter CI is also having problems but I don't know if that's an issue? I see #93 from Marcus so probably expected and ok to ignore there? If so +1 from me for merging (depending of the flutter channel question reply though)

kenvandine:
> Great, the snap builds wit those changes is working as expected and not hitting dri drivers errors anymore.
> 
> Our project instructions state to use flutter dev and there you change to master, is that an inconsistency we should fix or do we expect to use dev for work and roll out on stable (but then we don't test what we ship to users?)
> 
> The flutter CI is also having problems but I don't know if that's an issue? I see #93 from Marcus so probably expected and ok to ignore there? If so +1 from me for merging (depending of the flutter channel question reply though)

I've switched it to use the dev channel.  The CI failures are unrelated and @MarcusTomlinson is working on fixing that.

seb128 (approved):
Thanks, seems fine to me, let's land it!

PR #92, opened 2021-05-13
Add WizardPage to reduce boilerplate & ensure consistent page structure

jpnurmi:
`WizardPage` provides the appropriate layout and the common building blocks for installation wizard pages.

```
 __________________
|                  |
| title (app bar)  |
|__________________|
|                  |
| header           |
|__________________|
|                  |
|                  |
|     content      |
|                  |
|__________________|
|                  |
| (footer) buttons |
|__________________|

```

In the current design, all other pages follow this structure except the last two pages (installation slides & installation complete).

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/wizard_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/wizard_page.dart
@@ -1,0 +1,150 @@
+import 'package:flutter/material.dart';
+
+const _kButtonSpacing = 8.0;
+const _kContentSpacing = 20.0;
+const _kContentPadding = EdgeInsets.symmetric(horizontal: 24);
+const _kHeaderPadding = EdgeInsets.fromLTRB(24, 24, 24, 0);
+const _kFooterPadding = EdgeInsets.fromLTRB(24, 0, 24, 24);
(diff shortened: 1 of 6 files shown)


packages/ubuntu_desktop_installer/lib/pages/wizard_page.dart
```diff
@@ -0,0 +1,147 @@
...
+  /// Continues to the next page.
+  next,
+}
+
+/// The base for wizard pages in the installer.
+///
+/// Provides the appropriate layout and the common building blocks for
+/// installation wizard pages.
+class WizardPage extends StatelessWidget {
+  /// Creates the wizard page. The [titleText] must be specified.
```
oSoMoN, line 27: I don't see any other reference to `[titleText]`.

oSoMoN (commented):
LGTM, just one minor comment.

jpnurmi:
What do you think about specifying the button labels outside `WizardPage`? It's slightly less convenient but then we could get rid of the `LocalizedView` instance in `WizardPage`, which is there just to get the default button labels.
```
WizardPage {
    // These would need to repeated for every page even though most of them have _Back_ & _Continue_.
    backLabel: lang.foo,
    nextLabel: lang.bar,
}
```

PR #93, opened 2021-05-13
Fix CI

MarcusTomlinson:


packages/subiquity_client/lib/src/http_unix_client.dart
```diff
@@ -86,7 +94,10 @@ class HttpUnixClient extends BaseClient {
 
   @override
   void close() {
-    _socket.close();
+    if (_socket != null) {
+      _socket?.close();
```
jpnurmi, line 98: `Socket.close()` is asynchronous. Would await help?

Side note: is there a chance there might be buffered data? Should the socket be flushed before closing?

packages/subiquity_client/lib/test/test_server.dart
```diff
@@ -42,7 +42,8 @@ class TestServer {
     return socketPath;
   }
 
-  void stop() {
+  Future<void> stop() async {
     _serverProcess.kill();
```
jpnurmi, line 46: Pass `ProcessSignal.sigkill` to make sure it dies? :)

MarcusTomlinson: Thanks for the suggestions, not sure yet what to do here. The subiquity server appears to freeze now and then upon receiving an invalid function parameter. Trying to figure out if it's Dart-side or Python-side at the moment. I may just disable that test for now...

PR #97, opened 2021-05-17
.gitignore: add .vscode/ for workspace settings

jpnurmi:
For example, [peacock](https://www.peacockcode.dev/) is an awesome VS Code extension that stores settings per workspace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 coverage/
 .idea/
 *.iml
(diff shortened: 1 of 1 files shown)


MarcusTomlinson:
Yeah, we don’t have any plans to include any vscode settings at this time so all good. If we do in the future we can just change this to exclude specifics from within .vscode

PR #106, opened 2021-06-02
Include missing stage packages and define pythonpath

seb128:
In case you wonder, the probert stage-package got split because snapcraft isn't pulling the depends (expected or bug?). The libglib dev is staged because flutter tries to load libgio-2.0.so and not .so.0.

A locally built snap now gives a working subiquity service on a new installation and the gobject types errors are resolved on focal


MarcusTomlinson (approved):
Works for me!

PR #107, opened 2021-06-03
Stage python3-requests-unixsocket

seb128:
It was pulled it on the default installation by software-properties until hirsute so testing on focal didn't show it was needed, but it isn't anymore since https://launchpad.net/ubuntu/+source/software-properties/0.99.6


kenvandine (approved):
LGTM

PR #112, opened 2021-06-16
Update instruction with missing package

didrocks:
Add missing liblzma package needed for ubuntu-desktop-installer


MarcusTomlinson:
This should actually go into the upstream flutter docs. I’ll handle it.

PR #113, opened 2021-06-18
Update probert, and tweak other depends

dbungert:
Make the subiquity depends look more like the subiqity snapcraft
definition.  This is primarly to pick up the desired verison of probert.


MarcusTomlinson (approved):
Looks good. Thanks @dbungert!

PR #114, opened 2021-06-21
Update yaru dependency to 0.0.7

Feichtmeier:
Updating yaru dependency to 0.0.7 since many changes happend since 0.0.2:
https://github.com/ubuntu/yaru.dart/blob/main/CHANGELOG.md

Especially replacing all aubergine elements with orange elements syncing with the canonical design team rebranding


MarcusTomlinson (approved):
Thanks for the update @Feichtmeier!

Issue #43, opened 2021-03-08
Screen 17 - Installation Complete


seb128:
The design for screen n°17, Installation Complete

![installed](https://user-images.githubusercontent.com/10974330/110315051-fac15500-8008-11eb-92d9-77fc25af2dab.png)

Design details https://people.canonical.com/~platform/desktop/installer/17.%20Installation%20Complete/

HaroonAwan:
I would like to work on this screen.

oSoMoN:
@HaroonAwan you stated your interest in implementing this screen on April 21st, but haven't followed up since then. Is this something you're still interested in? I'm asking because @charafau has proposed an initial implementation of it, so in order not to duplicate work, you might want to unassign yourself and pick something else?

oSoMoN:
I'd be interested in getting more details from design on the format of the image asset: will it be provided already cropped, or is the expectation that the asset is square, and the circular cropping happens dynamically when rendering it in the application?

long-chung:
@oSoMoN It would be great if the cropping can happen dynamically.
That way we can swap out the asset any time without waiting for a designer to supply a new asset that has been cropped and scaled each time.

oSoMoN:
That was my reasoning, thanks for confirming @long-chung.

madsrh:
The content on this screen takes up very little of window size. Would it be possible to add a faded mascot image in the background? (I apologize for the pixelated mockup below)

![image](https://user-images.githubusercontent.com/3986894/124367907-e0685280-dc5b-11eb-9e7c-650ebfd5f764.png)

long-chung:
Not keen on this. There's a lot of noise and repetition on this screen.

oSoMoN:
This has been implemented by @charafau, thanks!

PR #119, opened 2021-06-29
Screen 17 - Installation Complete

charafau:
Implemented screen 17 design

Screenshot from 2021-06-29 03-07-06

issue: https://github.com/canonical/ubuntu-desktop-installer/issues/43


packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete_model.dart
```diff
@@ -0,0 +1,38 @@
+const _boldMarkdown = '**';
+const _boldMarkdownLength = _boldMarkdown.length;
+
+class TextUtils {
```
charafau, line 4: this is simple text parsing for markdown like bold text - not sure how text from design will look like in other languages, it might be on the front, end or in the middle so this supports that text spans just need to be in markdown format

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete.dart
```diff
@@ -0,0 +1,83 @@
...
+                child: Text(lang.shutdown),
+              ),
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+
+  ButtonStyle? _buttonStyle(BuildContext context, {bool? highlighted}) {
```
charafau, line 72: I copied this style from WizardPage widget, it is probably better to make this theme external (same with colors), I'm open to discussion

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete.dart
```diff
@@ -0,0 +1,83 @@
...
+        .copyWith(fontWeight: FontWeight.bold);
+    return LocalizedView(
+      builder: (context, lang) => WizardPage(
+        title: Text(lang.installationCompleteTitle),
+        content: Column(
+          children: [
+            Padding(
+              padding: const EdgeInsets.only(top: 64, bottom: 32),
+              child: Image.asset(
+                'assets/version.png',
```
charafau, line 29: I put image into assets, it is probably better to decide some path in the system and load it from file (for different versions)

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete.dart
```diff
@@ -0,0 +1,83 @@
...
+                width: 160,
+                height: 160,
+              ),
+            ),
+            RichText(
+              text: TextSpan(
+                  style: Theme.of(context).textTheme.bodyText1,
+                  children: TextUtils.parseText(lang.readyToUse('Ubuntu 20.04'))
```
charafau, line 37: This should be loaded from external file, but I'm not sure where, again, open for discussion

charafau (commented):
I added few comments to code so it should be clear

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete.dart
oSoMoN: I think an asset is fine, at least for the time being.

Design commented on the issue (https://github.com/canonical/ubuntu-desktop-installer/issues/43#issuecomment-870672818) stating that it'd be great if the circular cropping could happen dynamically. Can you look into implementing that?

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete.dart
oSoMoN: For ISOs, there is a file on disk (`/cdrom/.disk/info`) that contains the version and codename of the product being installed, so you could probably extract the required information from there. Something similar is done for the release notes URL, see the `_releaseNotesURL()` function in `lib/pages/try_or_install_page.dart`.

The contents of that file for a 20.04.2 release ISO downloaded from ubuntu.com is:

    Ubuntu 20.04.2.0 LTS "Focal Fossa" - Release amd64 (20210209.1)

So some careful parsing will be required to extract the relevant string.

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete.dart
oSoMoN: Yes, it'd be better to avoid duplicating code. Can you extract this `_buttonStyle()` function into a shared piece of code?

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete_model.dart
oSoMoN: This is not very flexible: it assumes there's exactly one bold token in the input string, so we won't be able to re-use it if in the future we want to render strings with more than one bold token, or e.g. italic script.
I would say that either we go with a more generic implementation (probably using regular expressions) or we use an existing markdown package [from pub.dev](https://pub.dev/packages?q=markdown).
If we go with the first solution (custom implementation), please extract it to a place where it can be re-used by all pages.

oSoMoN (changes requested):
Thank you for working on this @charafau!

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete_model.dart
charafau: My first idea was to use markdown plugin, not sure if it's worth it at current stage to import whole library for one string in the whole application? For me it's easier to import :)  will use markdown then

charafau:
@oSoMoN should be good now, I didn't know what to display when file is not there so I'm returning a dummy string, let me know if you want me to change it

packages/ubuntu_desktop_installer/lib/pages/installation_complete/version_extractor.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:io';
+
+class VersionExtractor {
+  String _dummy() {
+    return 'Ubuntu 20.04';
```
oSoMoN, line 5: Instead of a hardcoded string, you should test for the existence of `/etc/os-release` and use the value of the `PRETTY_NAME` key, if it's present. As a last resort, I would just hardcode "Ubuntu", without a version number.

packages/ubuntu_desktop_installer/lib/pages/installation_complete/version_extractor.dart
```diff
@@ -0,0 +1,31 @@
...
+      }
+    }
+    return _dummy();
+  }
+
+  String extractVersion(String content) {
+    final chunks = content.split(' ');
+    if (chunks.length > 1) {
+      return '${chunks[0]} ${chunks[1]}';
+    }
```
oSoMoN, line 28: I'm not sure whether there's a specification for the contents of that `info` file, but it looks to me like the format is roughly:

    PRODUCTNAME VERSION "CODENAME" - OTHERINFO

and what we want is PRODUCTNAME and VERSION. I can imagine PRODUCTNAME may contain whitespaces (e.g. "Ubuntu MATE"), and so may VERSION (e.g. "20.04.2.0 LTS"), so splitting on whitespaces isn't going to handle all cases nicely.

I'd recommend a regexp that extracts those components and keeps only PRODUCTNAME and VERSION.

packages/ubuntu_desktop_installer/test/version_extractor_test.dart
```diff
@@ -0,0 +1,31 @@
...
+
+      final version = extractor.extractVersion('Ubuntu 21.04.10');
+
+      expect(version, 'Ubuntu 21.04.10');
+    });
+
+    test('should return empty string when version too short', () {
+      final extractor = VersionExtractor();
+
+      final version = extractor.extractVersion('Ubuntu');
```
oSoMoN, line 26: 'Ubuntu' would be a malformed input string, but I would say that as a best-effort fallback this should return 'Ubuntu', not an empty string.

oSoMoN (changes requested):
Thanks! I have a few more comments. Also, is there a good reason the image asset became a jpg?

charafau:
> Thanks! I have a few more comments. Also, is there a good reason the image asset became a jpg?

I created new image from wallpaper to see if squared images will be rounded and saved it as jpg as accident (plus I'm adding frame dynamically). Can  save it as png if that's necessary

oSoMoN:
> Can save it as png if that's necessary

Yes please, we want lossless compression where possible.

packages/ubuntu_desktop_installer/lib/pages/installation_complete/version_extractor.dart
charafau: I checked few isos and here are my findings:
Xubuntu 20.04.2.0 LTS "Focal Fossa" - Release amd64 (20210209.1)
Ubuntu 20.04.2.0 LTS "Focal Fossa" - Release amd64 (20210209.1)
Kubuntu 21.04 "Hirsute Hippo" - Release amd64 (20210420)
Kubuntu 20.04.2.0 LTS "Focal Fossa" - Release amd64 (20210209.1)
Ubuntu-MATE 21.04 "Hirsute Hippo" - Release amd64 (20210420)
Ubuntu-Budgie 21.04 "Hirsute Hippo" - Release amd64 (20210420)

seems like instead of some regex which would introduce more problems probably (insert xkcd comic here) it would be the best to just substring to first `"`

charafau:
@oSoMoN  I fixed that changes, and merge current main because of conflicts for some reason new subiquity client tests has failed I would like to merge this one so I can move to another screen if possible

oSoMoN:
Your subiquity submodule is outdated. Please update it locally with `git submodule update --recursive` and push.

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space_page.dart
```diff
@@ -126,8 +126,12 @@ class _AllocateDiskSpacePageState extends State<AllocateDiskSpacePage> {
...
                         await client.setGuidedStorage(choice);
 
-                    Navigator.pushNamed(context, Routes.writeChangesToDisk,
-                        arguments: storageResponse.config);
+                    await client.confirm('/dev/tty1');
+
+                    // Navigator.pushNamed(context, Routes.writeChangesToDisk,
+                    //     arguments: storageResponse.config);
+                    Navigator.of(context)
+                        .pushNamed(Routes.installationComplete);
```
oSoMoN, line 134: That's not right. The new page should go at the end of the current workflow, which is after the "Choose your look" page.

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete.dart
```diff
@@ -0,0 +1,78 @@
...
+                  ),
+                ],
+              ),
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+}
```
oSoMoN, line 78: For consistency, that file should be called `installation_complete_page.dart`.

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
```diff
@@ -0,0 +1,65 @@
+import 'dart:io';
+
+class VersionExtractor {
```
oSoMoN, line 3: The name of the class isn't really self-explanatory (it extracts more than just a version number). What about `ProductInfoExtractor`? Also, it could really use documentation to explain what it does and for its public members.

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
```diff
@@ -0,0 +1,65 @@
...
+  static VersionExtractor get instance {
+    if (_instance == null) {
+      _instance = VersionExtractor._();
+    }
+    return _instance!;
+  }
+
+  VersionExtractor._();
+
+  String loadedVersion = '';
```
oSoMoN, line 15: What about a static `_cachedProductInfo` attribute, instead of making the class a singleton?

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
```diff
@@ -0,0 +1,65 @@
...
+
+  String _dummy() {
```
oSoMoN, line 17: I would call it `_fallback`, rather than `_dummy`.

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
```diff
@@ -0,0 +1,65 @@
...
+    return 'Ubuntu';
+  }
+
+  String extractLocalVersion(String content) {
+    return content.substring(content.indexOf("\"") + 1, content.length - 1);
+  }
+
+  String getVersion({
```
oSoMoN, line 25: In line with my previous suggestion, I would rename that method to something like `getProductInfo()`.

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
```diff
@@ -0,0 +1,65 @@
...
+    String path = '/cdrom/.disk/info',
```
oSoMoN, line 26: `isoPath` would be more explicit.

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
```diff
@@ -0,0 +1,65 @@
...
+    String localPath = '/etc/os-release',
+  }) {
+    if (loadedVersion.isNotEmpty) {
+      return loadedVersion;
+    }
+
+    final fileOnCdrom = File(path);
+    final localVersionFile = File(localPath);
```
oSoMoN, line 34: No need to instantiate that File object until after failing to extract the info from the ISO.

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
```diff
@@ -0,0 +1,65 @@
...
+        //TODO: configure logger?
+      }
+    } else {
+      loadedVersion = _dummy();
+    }
+
+    return loadedVersion;
+  }
+
+  String extractVersion(String content) {
```
oSoMoN, line 62: This method doesn't need to be public, does it?

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
```diff
@@ -0,0 +1,65 @@
...
+
+        loadedVersion = extractVersion(content);
+      } catch (e) {
+        //TODO: configure logger?
+      }
+    } else if (localVersionFile.existsSync()) {
+      try {
+        final content = localVersionFile
+            .readAsLinesSync()
+            .firstWhere((line) => line.contains('PRETTY_NAME'));
```
oSoMoN, line 49: line.startsWith('PRETTY_NAME') would be more correct

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
```diff
@@ -0,0 +1,65 @@
...
+  VersionExtractor._();
+
+  String loadedVersion = '';
+
+  String _dummy() {
+    return 'Ubuntu';
+  }
+
+  String extractLocalVersion(String content) {
+    return content.substring(content.indexOf("\"") + 1, content.length - 1);
```
oSoMoN, line 22: This method doesn't need to be public, I think.
And the parsing makes unnecessary assumptions (that the closing double quote is the last character on the line). You could use `String.lastIndexOf` or `String.split`.

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
```diff
@@ -0,0 +1,65 @@
...
+    final localVersionFile = File(localPath);
+    if (fileOnCdrom.existsSync()) {
+      try {
+        final content = fileOnCdrom
+            .readAsLinesSync()
+            .firstWhere((line) => line.trim().isNotEmpty);
+
+        loadedVersion = extractVersion(content);
+      } catch (e) {
+        //TODO: configure logger?
```
oSoMoN, line 43: This doesn't do what you want: when catching an exception here (and lower down in the same method), the code should move on to trying the next extraction method for the product info, and as a last resort use the fallback hardcoded string. Here it will simply return an empty `loadedVersion`.

I don't think there's a need for logging here.

The tests for the class should try and test all these cases.

oSoMoN (changes requested):

packages/ubuntu_desktop_installer/lib/utils/version_extractor.dart
charafau: @oSoMoN I made those method public because I wanted to test just them. If we want to keep the test, it might be good to extract parsing logic to serparate class, what would you prefer?

oSoMoN: That's not the correct approach: tests should exercise (thoroughly) public methods because they are public, the rest is implementation details. In fact I'd argue that these methods shouldn't even be separate methods, since they're used in only one place their implementation should be inlined.

charafau: I agree, still it would be nice to have test for those parsers. Seems like file package ( https://stackoverflow.com/questions/61504873/testing-mocking-file-io-in-flutter ) allows for in memory file mocking. Should we move to this approach?

oSoMoN: Yes, we could do that. If I understand correctly, that would require passing a `FileSystem` instance to the `VersionExtractor` constructor, which exposes an implementation detail, but that's an acceptable trade-off to achieve good test coverage I suppose.

charafau: Sounds good, I will play with it and report my findings or just implement that if it's straighforward :+1:

packages/ubuntu_desktop_installer/lib/utils/product_info_extractor.dart
```diff
@@ -0,0 +1,64 @@
...
+            content.indexOf("\"") + 1, content.lastIndexOf("\""));
+      } on Exception catch (e) {
+        _cachedProductInfo = 'Ubuntu';
+      }
+    } else {
+      _cachedProductInfo = 'Ubuntu';
+    }
+  }
+
+  @visibleForTesting
```
charafau, line 60: I added this because there's no other way to reset the state and I don't want to flaky tests to happen

charafau:
@oSoMoN  ok, I think I applied changed to every change request, tried to fix that submodule but your command doesn't help. Can you commit to my branch to help to fix that?

packages/ubuntu_desktop_installer/lib/widgets/highlighted_button.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:flutter/material.dart';
+
+const _kHighlightBackground = Color(0xFF0e8420);
+const _kHighlightForeground = Colors.white;
+
+class HighlightedButton extends StatelessWidget {
```
oSoMoN, line 6: https://github.com/canonical/ubuntu-desktop-installer/commit/67f6241d88ef6f1842586ab9a5edc9ce6371e19f exposed the style for highlighted buttons, rendering this class unnecessary. You should now use a regular `OutlinedButton` with `style: Theme.of(context).highlightedButtonTheme.style`.

packages/ubuntu_desktop_installer/lib/utils/product_info_extractor.dart
```diff
@@ -0,0 +1,64 @@
+import 'dart:io';
+
+import 'package:file/file.dart';
+import 'package:file/local.dart';
+import 'package:flutter/foundation.dart';
+
+class ProductInfoExtractor {
+  static String _cachedProductInfo = '';
+
+  final FileSystem fileSystem;
```
oSoMoN, line 10: Can `fileSystem` be made private?

packages/ubuntu_desktop_installer/lib/utils/product_info_extractor.dart
```diff
@@ -0,0 +1,64 @@
+import 'dart:io';
+
+import 'package:file/file.dart';
+import 'package:file/local.dart';
+import 'package:flutter/foundation.dart';
+
+class ProductInfoExtractor {
```
oSoMoN, line 7: Can you please add documentation for the class and its one public method?

packages/ubuntu_desktop_installer/lib/utils/product_info_extractor.dart
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    if (fileSystem.file(isoPath).existsSync()) {
+      try {
+        final content = fileSystem
+            .file(isoPath)
+            .readAsLinesSync()
+            .firstWhere((line) => line.trim().isNotEmpty);
+
+        _cachedProductInfo = content.substring(0, content.indexOf('"') - 1);
```
oSoMoN, line 31: A comment to explain the expected format (and why we're breaking on the first occurrence of ") would be helpful.

packages/ubuntu_desktop_installer/lib/utils/product_info_extractor.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  final FileSystem fileSystem;
+
+  ProductInfoExtractor({
+    this.fileSystem = const LocalFileSystem(),
+  });
+
+  String getProductInfo({
+    String isoPath = '/cdrom/.disk/info',
+    String localPath = '/etc/os-release',
```
oSoMoN, line 18: As far as I can tell, `isoPath` and `localPath` are never overridden, not even in tests, so there is no need for these hardcoded file paths to be parameters, they can be made private constants, or local variables in the scope of the method.

packages/ubuntu_desktop_installer/lib/utils/product_info_extractor.dart
oSoMoN: Unfortunately this annotation only provides an info-level message from the analyzer, it doesn't forbid calling the method in production code, so it feels hackish. Maybe add a `force_reload` (or `ignore_cached`) parameter to `getProductInfo()`?

packages/ubuntu_desktop_installer/lib/utils/product_info_extractor.dart
```diff
@@ -0,0 +1,64 @@
+import 'dart:io';
```
oSoMoN, line 1: That import is unused now.

There's also a number of new analyzer warnings in this code that can be easily addressed.

packages/ubuntu_desktop_installer/test/product_info_extractor_test.dart
```diff
@@ -0,0 +1,105 @@
+import 'package:file/file.dart';
+import 'package:file/memory.dart';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:ubuntu_desktop_installer/utils/product_info_extractor.dart';
+
+void main() {
+  group('ProductInfoExtractor', () {
```
oSoMoN, line 8: All the tests are in the same group, so this is an unnecessary level.

packages/ubuntu_desktop_installer/test/product_info_extractor_test.dart
```diff
@@ -0,0 +1,105 @@
...
+    tearDown(() {
+      productInfoExtractor?.resetState();
+    });
+
+    test('should return product info when iso file exists', () async {
+      final isoPath = '/cdrom/.disk/info';
+
+      await fileSystem.file(isoPath).create(recursive: true).then((File f) {
+        f.writeAsString(
+            'Ubuntu 21.04 "Hirsute Hippo" - Release amd64 (20210420)');
```
oSoMoN, line 26: It would be useful to test with the contents of various ISOs (re [your comment](https://github.com/canonical/ubuntu-desktop-installer/pull/119#discussion_r665750514)).

oSoMoN (changes requested):

packages/ubuntu_desktop_installer/lib/widgets/highlighted_button.dart
oSoMoN: And since https://github.com/canonical/ubuntu-desktop-installer/commit/ca251e60236838839e46506e4fa6beb79b651975, you should simply use an `ElevatedButton` and not worry about the style, Yaru will handle it for you.

packages/ubuntu_desktop_installer/test/product_info_extractor_test.dart
charafau: `group` is necessary for `setUp()` method. it's also good convention

oSoMoN: It's not necessary. See https://pub.dev/packages/test :

> You can use the setUp() and tearDown() functions to share code between tests. The setUp() callback will run before every test _**in a group or test suite**_, and tearDown() will run after.

It doesn't hurt though, feel free to keep it.

charafau:
@oSoMoN another try, I think I applied to all comments

oSoMoN:
@charafau I resolved the merge conflicts and rebased on the latest changes in main in my personal fork at [oSoMoN:installation-complete](https://github.com/oSoMoN/ubuntu-desktop-installer/commits/installation-complete). Please review and pull from that branch if that looks correct to you, and then we can proceed with merging your branch into main.
Thanks again for your work on this and for your patience!

charafau:
@oSoMoN  thank you for taking time and looking into this, I took a look and your changes are good and I'm happy when you guys are happy :) 

I merged them into my branch and pushed the code so it should be good to go I guess?

PR #124, opened 2021-07-02
Present subiquity-loadkeys util

dbungert:
subiquity now expects to find loadkeys in path somewhere.
So offer that.

### DIFF
--- a/snap/local/subiquity-server
+++ b/snap/local/subiquity-server
@@ -3,7 +3,10 @@
 SCRIPT_DIR=`dirname $0`
 
 export PYTHONIOENCODING=utf-8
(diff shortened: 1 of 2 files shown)


oSoMoN:
Thanks @dbungert. That's better in the context of the ubuntu-desktop-installer snap, but when I run it from a checkout in my development environment, the subiquity server complains because it doesn't find `subiquity-loadkeys` in the path. Could subiquity fall back to trying to run `./bin/subiquity-loadkeys` if the binary isn't in the path, maybe?

oSoMoN:
Let's merge this as it already makes things better (at least in the snap case).
@dbungert please follow up with a fix for the local development environment case. Thanks!

PR #130, opened 2021-07-06
Use our own python interpreter to ensure the right version

seb128:
Otherwise we are getting error due to mismatching version depending on the distribution serie in use


MarcusTomlinson (approved):
Looks good. Just added some comments to the subiquity-server script to demystify the $SNAP override there.

PR #142, opened 2021-07-08
Make WizardPage's constants and themes available

jpnurmi:


oSoMoN:
LGTM, thanks!

Feichtmeier:
@jpnurmi the availible sizes look great, but I don't think it is a good idea to use installer specific hard coded colors. 
Can't you get those from Theme.of(context) .... ? Yaru.dart should give you all you need. This reduces the chance of having stuff not dark mode compatible

jpnurmi:
Sounds good! I didn't make any changes to any of the values in this one.

PR #144, opened 2021-07-10
Who are you page

### LINKED ISSUE #39: Screen 14 - Who Are You
The design for screen n°14, Who Are You

who

Design details https://people.canonical.com/~platform/desktop/installer/14.%20Who%20Are%20You/

### LINKED ISSUE #40: Screen 14 - Who Are You
The design for screen n°14, Who Are You

who

Design details https://people.canonical.com/~platform/desktop/installer/14.%20Who%20Are%20You/

Feichtmeier:
(click to enlarge)
whoareyou

TODO:

- [x] Decide on the validator regexes
- [x] Exchange the password :heavy_check_mark: with red / yellow / green text
- [x] do not remove password fields on radio toggle, I misunderstood this
- [x] use localized strings
- [x] Document all  public members
- [x] Add a test

@long-chung
- What validation regex do you want to have for "name" (currently only not allowed to be empty) "computer name" (currently not empty and two digits), "username" (currently only not allowed to be empty and two digits)?
- How many levels of password strengths do you have in mind and with what regex should they trigger? Currently three
- I think that the "The name it uses when it talks to other computers" is not really helping much, it draws the attention of the eyes to a place which actually does not do much - shouldn't it be clear what "computer name" is? :) I left an option for an alternative place (inside the textfield) - or we could just remove it?
- subiquity currently has no support for configuring the login strategy (auto login, require password). also, currently in ubuntu if you select this, you can login without a password but then immediately after login the popup comes up with "enter password for the keyring". IMO those radios are not helping much. isn't it also a bit insecure?

Edit: I've settled it down to three password strength for now. If we feel we need more, we can improve this later on

Closes #40 
Thanks again @jpnurmi for the help with the ValueListenableBuilder


packages/ubuntu_desktop_installer/lib/app.dart
```diff
@@ -54,6 +55,7 @@ class UbuntuDesktopInstallerApp extends StatelessWidget {
           Routes.allocateDiskSpace: (context) => AllocateDiskSpacePage(),
           Routes.writeChangesToDisk: (context) => WriteChangesToDiskPage(),
           Routes.chooseYourLook: (context) => ChooseYourLookPage(),
+          Routes.whoAreYouPage: (context) => WhoAreYouPage()
```
jpnurmi, line 58: could you follow the `static Widget WhoAreYouPage.create()` pattern shown in the _After_ section of #147, please?

packages/ubuntu_desktop_installer/pubspec.yaml
```diff
@@ -39,6 +39,8 @@ dependencies:
   tuple: ^2.0.0
   url_launcher: ^6.0.3
   yaru: ^0.0.8
+  form_field_validator: ^1.1.0
```
jpnurmi, line 42: can we keep the dependencies in alphabetical order?

P.S. are you, by any chance, using vscode? if so, check [Pubspec Assist](https://marketplace.visualstudio.com/items?itemName=jeroen-meijer.pubspec-assist) out! it's really nice for managing dependencies, but it places the list of dependencies in alphabetical order (besides flutter sdk dependencies at the top).

packages/ubuntu_desktop_installer/lib/widgets/validated_input.dart
```diff
@@ -0,0 +1,72 @@
...
+      : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Padding(
+      padding: const EdgeInsets.only(top: 10, bottom: 10),
+      child: Row(
+        children: [
+          SizedBox(
+            width: MediaQuery.of(context).size.width / 1.6,
```
jpnurmi, line 46: is this needed inside the general purpose widget, or could we let the page decide how to layout? this might be used by dialogs too.

packages/ubuntu_desktop_installer/lib/widgets/validated_input.dart
```diff
@@ -0,0 +1,72 @@
...
+
+  /// The label above the [TextField]
+  final String label;
+
+  /// This boolean is forwarded to the [TextField] to decide
+  /// if the digits should be obscured.
+  final bool obscureText;
+
+  /// The specific widget shown right to the [TextField] if the
+  /// input value is validated.
```
jpnurmi, line 23: is "valid"?

packages/ubuntu_desktop_installer/lib/pages/who_are_you_page.dart
```diff
@@ -0,0 +1,328 @@
...
+                      final encrypter = encrypt.Encrypter(encrypt.AES(key));
+
+                      final encrypted =
+                          encrypter.encrypt(_passwordController.text, iv: iv);
+
+                      await client.setIdentity(IdentityData(
+                          hostname: _computerNameController.text,
+                          realname: _realNameController.text,
+                          username: _usernameController.text,
+                          cryptedPassword: encrypted.toString()));
```
jpnurmi, line 319: this kind of business logic would be nice to do in a view model, and test it with a mock client

packages/ubuntu_desktop_installer/lib/pages/who_are_you_page.dart
```diff
@@ -0,0 +1,328 @@
...
+  @override
+  _WhoAreYouPageState createState() => _WhoAreYouPageState();
+}
+
+/// An enum for storing the login strategy.
+enum LoginStrategy {
+  /// If selected can be used for the representation
+  /// of the user preference to always entering the password
+  /// at login.
+  REQUIRE_PASSWORD,
```
jpnurmi, line 29: in Dart 2, the naming convention for enum values is `lowerCamelCase`

packages/ubuntu_desktop_installer/lib/app.dart
```diff
@@ -2,6 +2,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:intl/intl.dart';
 import 'package:provider/provider.dart';
+import 'package:ubuntu_desktop_installer/pages/who_are_you_page.dart';
 import 'package:yaru/yaru.dart' as yaru;
```
jpnurmi, line 6: is there any chance we could keep `yaru.dart` dependencies in `app_theme.dart`?

we can provide custom `ThemeData` extensions like this: https://github.com/canonical/ubuntu-desktop-installer/commit/67f6241d88ef6f1842586ab9a5edc9ce6371e19f#diff-ae7dce7083af6843f072a97280a2bed2bed57742c87086bfe1000ac979121c30R34-R48

packages/ubuntu_desktop_installer/lib/pages/who_are_you_page.dart
```diff
@@ -0,0 +1,328 @@
...
+    _computerNameController = TextEditingController();
+    _usernameController = TextEditingController();
+    _passwordController = TextEditingController();
+    _confirmPasswordController = TextEditingController();
+
+    _loginStrategy = LoginStrategy.REQUIRE_PASSWORD;
+    super.initState();
+  }
+
+  Widget findPassWordStrengthLabel(String value, String weakPasswordLabel,
```
jpnurmi, line 62: can we describe the strength as an enum, determine the strength in view model, and merely create a widget here?

packages/ubuntu_desktop_installer/lib/pages/who_are_you_page.dart
```diff
@@ -0,0 +1,328 @@
...
+            color: yaru.Colors.red,
+          ));
+    }
+
+    return Text('');
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    const _screenFactor = 1.6;
```
jpnurmi, line 94: > DON’T use a leading underscore for identifiers that aren’t private.
> ...
> There is no concept of “private” for local variables, ...

https://dart.dev/guides/language/effective-dart/style#dont-use-a-leading-underscore-for-identifiers-that-arent-private

jpnurmi (commented):
I know it's still WIP but I took quick look and added a few initial suggestions here and there :)

packages/ubuntu_desktop_installer/lib/app.dart
Feichtmeier: Okay will do. How about defining an interface and let pages implement it? At least this is what I would do in Java :)

packages/ubuntu_desktop_installer/lib/widgets/validated_input.dart
Feichtmeier: Okay will move the size metrics widgets to the page

packages/ubuntu_desktop_installer/lib/app.dart
jpnurmi: The "interface" is https://api.flutter.dev/flutter/widgets/WidgetBuilder.html ;)

packages/ubuntu_desktop_installer/lib/pages/who_are_you_page.dart
Feichtmeier: Hm, isn't the user data used also in other pages? This would be an argument for a model dir, imo

packages/ubuntu_desktop_installer/lib/app.dart
Feichtmeier: I could use theme.of(context).... if this helps to get the yaru dep outside of this file?
only problem is the green, since in material elevated buttons are just colored with accent color, so we've overridden those in yaru.dart
but I could use

```dart
  Widget findPassWordStrengthLabel(
      BuildContext context,
      String value,
      String weakPasswordLabel,
      String averagePasswordLabel,
      String strongPasswordLabel) {
    final _strongPattenPasswordValidator =
        PatternValidator(r'(?=.*?[#?!@$%^&*-])', errorText: '');
    final _goodPatternPasswordValidator =
        PatternValidator(r'(^.*(?=.{6,})(?=.*\d).*$)', errorText: '');
    if (_strongPattenPasswordValidator.isValid(value)) {
      return Text(strongPasswordLabel,
          style: TextStyle(
            color: Theme.of(context)
                .elevatedButtonTheme
                .style!
                .backgroundColor!
                .resolve({MaterialState.dragged}),
          ));
    } else if (_goodPatternPasswordValidator.isValid(value)) {
      return Text(averagePasswordLabel,
          style: TextStyle(
            color: Theme.of(context).colorScheme.primary,
          ));
    }
    if (MultiValidator([
      RequiredValidator(errorText: ''),
      MinLengthValidator(2, errorText: ''),
    ]).isValid(value)) {
      return Text(weakPasswordLabel,
          style: TextStyle(
            color: Theme.of(context).colorScheme.error,
          ));
    }

    return Text('');
  }
 ```

File: packages/ubuntu_desktop_installer/lib/app.dart (hunk 1 above)
Comment (Feichtmeier) [reply packages/ubuntu_desktop_installer/lib/app.dart:58]: > could you follow the `static Widget WhoAreYouPage.create()` pattern shown in the _After_ section of #147, please?

Looked to your example and I think this would be too much for this PR for now. Or, if you can do it "on-the-fly" of your proposal implementation feel free to commit into this branch :)

File: packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart [hunk 10]
```diff
@@ -0,0 +1,368 @@
...
+                      ),
+                    ),
+                    Padding(
+                      padding: const EdgeInsets.only(top: 10, bottom: 10),
+                      child: Row(
+                        children: [
+                          SizedBox(
+                            width: MediaQuery.of(context).size.width /
+                                screenFactor,
+                            child: Form(
```
jpnurmi, line 209: there should be only one `Form` instance for all the fields

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+                      final encrypter = encrypt.Encrypter(encrypt.AES(key));
+
+                      final encrypted =
+                          encrypter.encrypt(_whoAreYouModel.password, iv: iv);
+
+                      await client.setIdentity(IdentityData(
+                          hostname: _computerNameController.text,
+                          realname: _realNameController.text,
+                          username: _usernameController.text,
+                          cryptedPassword: encrypted.toString()));
```
jpnurmi, line 359: can we have this logic in the model instead? `WhoAreYouModel.saveIdentity()` - a counterpart for `loadIdentity` that is called at init

packages/ubuntu_desktop_installer/lib/routes.dart
```diff
@@ -9,4 +9,5 @@ abstract class Routes {
   static const updatesOtherSoftware = '/updateothersoftware';
   static const writeChangesToDisk = '/writechangestodisk';
   static const chooseYourLook = '/chooseyourlook';
+  static const whoAreYouPage = '/whoareyoupage';
```
jpnurmi, line 12: other routes don't have the Page suffix

packages/ubuntu_desktop_installer/lib/widgets/validated_input.dart
```diff
@@ -0,0 +1,77 @@
+import 'package:flutter/material.dart';
+import 'package:form_field_validator/form_field_validator.dart';
+
+/// A [TextFormField] rewarding a validated input with a specific widget.
+class ValidatedInput extends StatelessWidget {
+  /// The key used to request validation before confirming the field values.
+  final GlobalKey<FormState> formKey;
```
jpnurmi, line 7: can be removed. only one Form + form key needed

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+        builder: (context, lang) => WizardPage(
+              title: Text(lang.whoAreYouPageTitle),
+              content: SingleChildScrollView(
+                child: Column(
+                  crossAxisAlignment: CrossAxisAlignment.start,
+                  children: [
+                    Padding(
+                      padding: const EdgeInsets.only(top: 10, bottom: 10),
+                      child: ValidatedInput(
+                        width: MediaQuery.of(context).size.width / screenFactor,
```
jpnurmi, line 131: `FractionallySizedBox` wouldn't need accessing `MediaQuery` repeatedly

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+                                ]),
+                                controller: _passwordController,
+                                obscureText: true,
+                                decoration: InputDecoration(
+                                    border: OutlineInputBorder(),
+                                    labelText: lang.whoAreYouPagePasswordLabel),
+                              ),
+                            ),
+                          ),
+                          ValueListenableBuilder<TextEditingValue>(
```
jpnurmi, line 230: can we make `ValidatedInput` more generic in the sense that it can build any indicator widget. not just the tickmark indicating success, but also the password strength label? then we can use `ValidatedInput` for all fields. it could be `ValueWidgetCallback` instead of a `Widget` and pass the valid state as a value

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+                              valueListenable: _passwordController,
+                              builder: (context, value, child) {
+                                return Padding(
+                                    padding: const EdgeInsets.only(
+                                        left: checkMarksLeftPadding),
+                                    child: findPassWordStrengthLabel(
```
jpnurmi, line 236: can we make `ValidatedInput` more generic in the sense that it can build any indicator widget. not just the tickmark indicating success, but also the password strength label? then we can use `ValidatedInput` for all fields. it could be `ValueWidgetCallback` instead of a `Widget` and pass the valid state as a value. then you can build the appropriate strength label here

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+                    Padding(
+                      padding: const EdgeInsets.only(top: 10, bottom: 10),
+                      child: Row(
+                        children: [
+                          SizedBox(
+                            width: MediaQuery.of(context).size.width /
+                                screenFactor,
+                            child: Form(
+                              key: _confirmPasswordFormKey,
+                              child: TextFormField(
```
jpnurmi, line 255: why not `ValidatedInput`?

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+                      ),
+                    ),
+                    Row(
+                      children: [
+                        Radio<LoginStrategy>(
+                            value: LoginStrategy.autoLogin,
+                            groupValue: _whoAreYouModel.loginStrategy,
+                            onChanged: (_) => setState(() {
+                                  _whoAreYouModel.loginStrategy =
+                                      _whoAreYouModel.loginStrategy;
```
jpnurmi, line 304: `model.loginStrategy = LoginStrategy.autoLogin`

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+      child: WhoAreYouPage(),
+    );
+  }
+
+  @override
+  _WhoAreYouPageState createState() => _WhoAreYouPageState();
+}
+
+class _WhoAreYouPageState extends State<WhoAreYouPage> {
+  late WhoAreYouModel _whoAreYouModel;
```
jpnurmi, line 34: don't store the model as a member, but use:
- `Provider.of<WhoAreYouModel>(context, listen: false)` in `initState`, and
- `Provider.of<WhoAreYouModel>(context)` in `build`

(it is important to not listen in the initialization phase, and must be listened in build to trigger rebuilds when the model notifies state changes)

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+  final _computerNameController = TextEditingController();
+  final _usernameController = TextEditingController();
+  final _passwordController = TextEditingController();
+  final _confirmPasswordController = TextEditingController();
+
+  final _usernameFormKey = GlobalKey<FormState>();
+  final _realNameFormKey = GlobalKey<FormState>();
+  final _computerNameFormKey = GlobalKey<FormState>();
+  final _passwordFormKey = GlobalKey<FormState>();
+  final _confirmPasswordFormKey = GlobalKey<FormState>();
```
jpnurmi, line 45: one key (with one form) should do

packages/ubuntu_desktop_installer/lib/widgets/validated_input.dart
```diff
@@ -0,0 +1,77 @@
...
+            ),
+          ),
+        ),
+        ValueListenableBuilder<TextEditingValue>(
+            valueListenable: controller,
+            builder: (context, value, child) {
+              return Padding(
+                  padding: EdgeInsets.only(left: spacing ?? 0.0),
+                  child: !validator.isValid(value.text)
+                      ? Text('')
```
jpnurmi, line 71: `SizedBox.shrink()` is a good way to present an empty widget

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+  void dispose() {
+    _realNameController.dispose();
+    _computerNameController.dispose();
+    _usernameController.dispose();
+    _passwordController.dispose();
+    _confirmPasswordController.dispose();
+    super.dispose();
+  }
+
+  Widget findPassWordStrengthLabel(
```
jpnurmi, line 82: - add `WhoAreYouModel.passwordStrength`
- determine the strength in the model instead as it's "business logic"
- return an enum value from the model, indicating the strength
- here in the view, build the representation (widget) for the strength returned from the model

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+                              })
+                        ],
+                      ),
+                    ),
+                    Row(
+                      children: [
+                        Radio<LoginStrategy>(
+                            value: LoginStrategy.autoLogin,
+                            groupValue: _whoAreYouModel.loginStrategy,
+                            onChanged: (_) => setState(() {
```
jpnurmi, line 302: please remove all `setState()` calls from the view. accessing the model in "listen" mode in the `build` method triggers rebuilds whenever the model notifies state changes

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
```diff
@@ -0,0 +1,368 @@
...
+/// The installer page for setting up the user data.
+///
+/// It uses [WizardPage] and [WizardAction] to create an installer page.
+class WhoAreYouPage extends StatefulWidget {
+  /// Creates a the installer page for setting up the user data.
+  const WhoAreYouPage({Key? key}) : super(key: key);
+
+  ///
+  static Widget create(BuildContext context) {
+    return ChangeNotifierProvider(
```
jpnurmi, line 23: `final client = Provider.of<SubiquityClient>(context, listen: false);` and pass it to `WhoAreYouModel` constructor so it will be able to call the client in a testable way

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_page.dart
Feichtmeier: i don't think this is a good idea, since the validation is only connected to the label "tip"
the current ubuntu installer validation only requires 2 digits, then you are allowed to proceed.

jpnurmi: i'm not sure i follow. :smile: the basic principle is to have functionality in the model and keep the view as a declarative presentation. you can keep the label creation here, just move the logic into the model so we can test it in a sensible way (not as a slow widget test, but as a unit test that spams the model with a combination of different password inputs and validates that the result is what we expect it to be). also, changing the rules later should not affect the UI as it's just a presentation of weak/good/strong password but it doesn't care how it's determined

jpnurmi: even if the validation requires only 2 characters, isn't the password strength going to be visualized as weak/good/strong? this is orthogonal to the validation logic, right?

P.S. I would very much like to have the actual validation logic out of the view too, for the very same purposes as described above, but let's take it step by step.

Feichtmeier: there are basically too validations going on here:

1. a bare minimum validation needed to proceed with the page, like "required" or "2 digits", which is in sync with the current ubuntu installer, so if this should be stronger we need someone from... design? so say something about this
2. a second validation _only_ needed for this "I am your password assistant telling you how good your password is but won't stop you to proceed with a bad password"-hint

Why should the second hinting be used inside the model? It is not relevant for the model, unless the design? wants us to force the user to make a better password to proceed

Feichtmeier: soooo the first validation could be indeed used to the model, but the second not, in my humble opinion ;)

jpnurmi: what i'm trying to say is that the responsibility of this function should be _visualizing_ "weak" "good" or "strong", not visualizing _and_ determining the strength.

> there are basically 2 validations going on here:

the validation rules are irrelevant for determining the strength of a password, right?  in other words, the strength does not affect the validation result. this is a standalone function, not strictly connected to the validation which is done elsewhere outside this function.

> Why should the second validation for hinting be used inside the model? It is not relevant for the model, unless the design?

if you mean software design, yes the view model is there specifically to help keep the view as simple, clean, and declarative as possible.

> wants us to force the user to make a better password to proceed

if the strength would affect the continue-button, that should be determined by the view model too. :wink:

Feichtmeier: right, I guess this is rather a question of taste... I would personally detach the model from the view completely. I will do as you suggested

jpnurmi: i have a feeling that the term "model" is the source of confusion here.

- in MVVM, which we are using, view model is like a relaxed version of MVC's controller
- some use the model term for data classes too (view model vs. data model)
- others prefer to call them just data classes (like the data classes in subiquity_client)
- some UI frameworks use the term "model" for item view's data source (lists, grids, tables...)
- probably many others...

is it possible that we are mixing up view models (business logic) with data models (pure data blob)?

Feichtmeier: also there can not be a global "model.isValid ?" like in the WSL page you created because each field could potentially have different validations. 
ontop there are multiple messages for those each field, so I would need to add a function for each of the wrong values like
you can not simple say "whoAreYouModel.isPasswordValid" because this axes again other options and thus other error texts

```dart
validator: (_) => whoAreYouModel.isPasswordEmpty
                                        ? lang.whoAreYouPagePasswordRequiredValidatorErrorText
                                        : whoAreYouModel.hasPasswordTheMinLengh ? lang.whoAreYouPagePasswordRequiredValidatorErrorText : null
```


```dart
validator: MultiValidator([
    RequiredValidator(
        errorText: lang
            .whoAreYouPagePasswordRequiredValidatorErrorText),
    MinLengthValidator(2,
        errorText: lang
            .whoAreYouPagePasswordMinLengthValidatorErrorText),
  ])
```

Do you really think this is an improvement? :thinking: @jpnurmi

Feichtmeier: and the error text string values must come from the lang object, which the model does not know

jpnurmi: yes, the model should not care about UI strings or widgets. the model knows the password, so it's a matter of querying the strength from the model (describe it as enum)

jpnurmi: Pseudo example:
```
enum PasswordStrength { weak, good, strong }

class WhoAreYouModel extends ChangeNotifier {
  // ...
  
  PasswordStrength? get passwordStrength {
    const strongPassword = RegExp(...);
    const goodPassword = RegExp(...);

    if (strongPassword.hasMatch(password)) {
      return PasswordStrength.strong;
    }
    if (goodPassword.hasMatch(password)) {
      return PasswordStrength.good;
    }
    if (password.length < something) {
      return PasswordStrength.weak;
    }
    return null;
  }
}
```

jpnurmi: the respective view would be something like:
```dart
switch (model.passwordStrength) {
  case PasswordStrength.weak: return Text(lang.foo, style: bar);
  case ...
  default: return SizedBox.shrink();
}
```

jpnurmi: notice that the only difference is that the logic and the presentation are separate. two functions, each with a clear responsibility, instead of one function doing both. :smile: 

not everyone is fluent in regexp patterns, so it's a good practice to test them to catch potential regressions when the rules are tweaked later. this is straightforward when all you need is to create a model instance, feed it with a ton of different passwords, and test that the result is as expected. this works great as a plain unit test. if it was done as a widget test, the test would have to create an instance of the whole page UI, find the correct input field instance, ensure that it has input focus, emulate entering text, find the result label, and then introspect its text.

jpnurmi:
These look promising:
- password strength: https://pub.dev/packages/password_strength
- encrypting passwords: https://pub.dev/packages/crypt

jpnurmi:
I've been pretty heavily involved in this so it would be nice if someone else takes a look.

jpnurmi:
Rebased to resolve conflicts

packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
```diff
@@ -61,6 +61,23 @@
   "diskHeadersSystem": "System",
   "diskHeadersFormat": "Format",
 
+  "whoAreYouPageTitle": "Who are you?",
+  "whoAreYouPageAutoLogin": "Log in automatically",
+  "whoAreYouPageRequirePassword": "Require my password to log in",
+  "whoAreYouPageRealNameLabel": "Your name",
+  "whoAreYouPageRealNameRequired": "A name is required",
+  "whoAreYouPageComputerNameLabel": "Your computers name",
```
robert-ancell, line 69: Missing apostrophe - "Your computer's name"

robert-ancell (approved):
LGTM

PR #145, opened 2021-07-10
Separate WelcomePage's business logic to WelcomeModel

jpnurmi:
This simplifies `WelcomePage`, and makes the business logic testable.

Ref: #147


Feichtmeier:
Just an idea since I also thought about a model for the where are you page: how about creating a 'model' dir in the root of lib?

jpnurmi:
We briefly discussed this at https://github.com/canonical/ubuntu-desktop-installer/pull/68#issuecomment-824761228. :smile: To reiterate, there are two common ways to categorize: a) by features or b) by types.

I'm in favor of the former, which is to group things into standalone pages/modules/features. When developing specific pages, the changes are mostly isolated into that directory where the page is placed into. It's also more straightforward to split modules into their own standalone packages should they be needed in multiple apps (thinking ahead to the similar network functionality in the installer and settings apps).

Of course, any shared services used across the pages, such as the upcoming network service, would be grouped into their own top-level services/ category.

jpnurmi:
Tiny sample size but https://twitter.com/FullmerS/status/1413401082458476544. There's a great comment by Oscar Martin from VGV:
> In my experience I found projects by function more complicate to find what you need, specially in big projects, while by feature you just need to go to one folder and you have everything you need related to that feature :D

Feichtmeier:
> It's also more straightforward to split modules into their own standalone packages should they be needed in multiple apps (thinking ahead to the similar network functionality in the installer and settings apps).

This makes sense :+1:

oSoMoN:
Much cleaner than the original implementation, thanks!

Issue #147, opened 2021-07-12
Proposal: separate business logic


jpnurmi:
I'd like to propose using MVVM for separating widgets (pages) and their business logic. Some benefits:
- present state with widgets, maintain state outside the widget tree
- notify state changes to control rebuilds (replaces setState() calls)
- testable business logic (mock services)
- easier to test widget tree (mock models)
- simpler and easier to read widget tree

### Before
```dart
class MyPage extends StatefulWidget {
  @override
  _MyPageState createState() => _MyPageState();
}

class _MyPageState extends State<MyPage> {
  int _value = 0;

  void _setValue(int value) {
    if (_value == value) return;
    setState(() => _value = value);
  }

  void _applyValue() {subjectively 
    final service = Provider.of<MyService>(context, listen: false);
    service.applyValue(_value);
  }

  @override
  Widget build(BuildContext context) {
    return Column(
      children: [
        Radio<int>(
          value: 1,
          groupValue: _value,
          onChanged: (v) => _setValue(v!),
        ),
        Radio<int>(
          value: 2,
          groupValue: _value,
          onChanged: (v) => _setValue(v!),
        ),
        Radio<int>(
          value: 3,
          groupValue: _value,
          onChanged: (v) => _setValue(v!),
        ),
        ElevatedButton(onPressed: _applyValue, child: Text('Apply')),
      ],
    );
  }
}
```

### After
```dart
class MyPage extends StatelessWidget {
  static Widget create(BuildContext context) {
    final service = Provider.of<MyService>(context, listen: false);
    return ChangeNotifierProvider(
      create: (_) => MyModel(service),
      child: MyPage(),
    );
  }

  @override
  Widget build(BuildContext context) {
    final model = Provider.of<MyModel>(context);
    return Column(
      children: [
        Radio<int>(
          value: 1,
          groupValue: model.value,
          onChanged: (v) => model.value = v!,
        ),
        Radio<int>(
          value: 2,
          groupValue: model.value,
          onChanged: (v) => model.value = v!,
        ),
        Radio<int>(
          value: 3,
          groupValue: model.value,
          onChanged: (v) => model.value = v!,
        ),
        ElevatedButton(onPressed: model.applyValue, child: Text('Apply')),
      ],
    );
  }
}

class MyModel extends ChangeNotifier {
  MyModel(MyService service) : _service = service;

  final MyService _service;

  int _value = 0;
  int get value => _value;
  set value(int value) {
[...]

PR #152, opened 2021-07-12
Split KeyboardLayoutPage & KeyboardLayoutModel

jpnurmi:
#147


packages/ubuntu_desktop_installer/lib/pages/keyboard_layout/keyboard_layout_model.dart
```diff
@@ -0,0 +1,104 @@
...
+import 'package:subiquity_client/subiquity_client.dart';
+
+import '../../keyboard_model.dart';
+
+/// Implements the business logic of the Keyboard Layout page.
+class KeyboardLayoutModel extends ChangeNotifier {
+  /// Creates a model with the specified [client] and [keyboardModel].
+  KeyboardLayoutModel({
+    required SubiquityClient client,
+    required KeyboardModel keyboardModel,
```
jpnurmi, line 12: would be nice to rename KeyboardModel to avoid confusion between similarly named models.

Some ideas:
- KeyboardDatabase
- KeyboardRepository
- KeyboardService

oSoMoN: Agreed. `KeyboardService` sounds fine to me.

oSoMoN:
This is going to conflict with #154, so we'll need to be careful when rebasing.

jpnurmi:
it's ok to merge the fix first. i can rebase later and add test cases for the new corner cases

oSoMoN:
This now needs rebasing (thanks for reviewing my other branch that created this conflict!).

jpnurmi:
It was easier to redo the changes than try to resolve the conflicts, but it's ok because the model changes are basically just moving small pieces of "logic" code from the page to the model. This time I did it in tiny steps, based on the latest main that contains the fix, to reduce the risk of regressions. Then I took the old PR tests and extended them with many cases with null/empty/unknown values.

packages/ubuntu_desktop_installer/lib/pages/keyboard_layout/keyboard_layout_model.dart
jpnurmi: :+1: I'll push a separate PR later

oSoMoN:
That looks very good, thanks for this thorough refactoring work!

PR #154, opened 2021-07-12
Call setxkbmap when changing the keyboard layout selection.

### LINKED ISSUE #148: Selecting a keyboard layout has no effect
Selecting a keyboard layout in the list doesn't change the active one, that's because the setKeyboard API from subiquity calls setupcon which works for VTs but not graphical sessions

The standard way to change the active layout for Xorg is to call setxkbmap -layout ... -variant ...

Under wayland there is no equivalent though. One way could be to call to localectl set-keymap and to teach g-s-d to listen to localed changes and reflect them?

oSoMoN:
Fixes #148.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/keyboard_layout_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/keyboard_layout_page.dart
@@ -23,13 +23,15 @@ class KeyboardLayoutPage extends StatefulWidget {
 }
 
 class _KeyboardLayoutPageState extends State<KeyboardLayoutPage> {
-  int _selectedLayoutIndex = 0;
-  KeyboardLayout? get _selectedLayout =>
-      Provider.of<KeyboardModel>(context, listen: false)
-          .layouts[_selectedLayoutIndex];
-  int _selectedVariantIndex = 0;
-  KeyboardVariant? get _selectedVariant =>
-      _selectedLayout!.variants![_selectedVariantIndex];
+  int _selectedLayoutIndex = -1;
+  KeyboardLayout? get _selectedLayout => (_selectedLayoutIndex > -1)
+      ? Provider.of<KeyboardModel>(context, listen: false)
+          .layouts[_selectedLayoutIndex]
+      : null;
+  int _selectedVariantIndex = -1;
+  KeyboardVariant? get _selectedVariant => (_selectedVariantIndex > -1)
+      ? _selectedLayout?.variants?.elementAt(_selectedVariantIndex)
+      : null;
 
   final _layoutListScrollController = AutoScrollController();
   final _keyboardVariantListScrollController = AutoScrollController();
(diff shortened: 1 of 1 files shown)


seb128:
Thanks Olivier, I tested and from an user perspective it is doing the work on an xorg session. One nitpick, the code doesn't handle cleanly setxkbmap not being installed and throw exception on stderr, would be nicer to rather log a warning or just do nothing in this case (the utility is standard but it's unclear it will always be available in a wayland world since it doesn't work there)

mwhudson:
I was sort of thinking we'd add code to the subiquity side for this, but I guess it's not required...

oSoMoN:
> I was sort of thinking we'd add code to the subiquity side for this, but I guess it's not required...

If calling `setxkbmap` has any benefit in a server environment, then sure, let's implement it in subiquity. Otherwise it's probably fine to have that logic remain in the desktop installer.

seb128:
> I was sort of thinking we'd add code to the subiquity side for this, but I guess it's not required...

It would perhaps still be useful if subiquity was using 'localectl set-keymap ...' rather than setupcon? we will eventually need to solve the wayland case and using localed might be the right solution

packages/ubuntu_desktop_installer/lib/pages/keyboard_layout_page.dart
```diff
@@ -39,24 +47,22 @@ class _KeyboardLayoutPageState extends State<KeyboardLayoutPage> {
           return layout.code == info.layout;
         });
         if (_selectedLayoutIndex > -1) {
-          _selectedLayoutName =
-              keyboardModel.layouts[_selectedLayoutIndex].code!;
-          _selectedVariantIndex = keyboardModel
-              .layouts[_selectedLayoutIndex].variants!
+          _selectedVariantIndex = _selectedLayout!.variants!
```
jpnurmi, line 50: is the list of variants guaranteed to be non-null or should it be `.variants?.indexWhere(...) ?? -1`?

packages/ubuntu_desktop_installer/lib/pages/keyboard_layout_page.dart
```diff
@@ -167,9 +181,8 @@ class _KeyboardLayoutPageState extends State<KeyboardLayoutPage> {
                 final client =
                     Provider.of<SubiquityClient>(context, listen: false);
                 final keyboard = KeyboardSetting(
-                    layout: _selectedLayoutName,
-                    variant: keyboardModel.layouts[_selectedLayoutIndex]
-                        .variants![_selectedVariantIndex].code!);
+                    layout: _selectedLayout!.code!,
+                    variant: _selectedVariant!.code!);
```
jpnurmi, line 185: - should the wizard action be disabled if there's no selected layout?
- KeyboardLayout.variants is nullable => can the selected variant be null at this point?

packages/ubuntu_desktop_installer/lib/pages/keyboard_layout_page.dart
oSoMoN: You're right, of course. Fixed.

packages/ubuntu_desktop_installer/lib/pages/keyboard_layout_page.dart
oSoMoN: Very good point. I disabled the button if there's no layout/variant selected. This ensures that when the action is triggered, the selected variant isn't null.

jpnurmi (approved):
LGTM :+1:

mwhudson:
Thinking about this a little more, I think doing this in the client for now makes total sense. If keyboard setting in general gets unified a bit we can look at making the server part use that.

packages/ubuntu_desktop_installer/lib/pages/keyboard_layout_page.dart
```diff
@@ -47,19 +49,20 @@ class _KeyboardLayoutPageState extends State<KeyboardLayoutPage> {
         if (_selectedLayoutIndex > -1) {
           _selectedVariantIndex = _selectedLayout!.variants!
               .indexWhere((variant) => variant.code == info.variant);
-          if (_selectedVariantIndex == -1) {
-            _selectedVariantIndex = 0;
```
jpnurmi, line 48: is it intentional that the default variant is no longer selected? the above `indexWhere` is comparing empty to null.

Screenshot from 2021-07-14 07-44-00

jpnurmi: Fix proposal: https://github.com/canonical/ubuntu-desktop-installer/pull/168

jpnurmi:
I noticed that `setxkbmap` is currently called in dry-run mode (https://github.com/canonical/ubuntu-desktop-installer/issues/250). Could we do it in subiquity so the UI doesn't need to know if it's running in dry-run mode?

PR #161, opened 2021-07-13
Tweak subiquity paths

seb128:
The subiquity util.py find_path logic fails to find curtin without the change


MarcusTomlinson (approved):
Thanks Seb

PR #167, opened 2021-07-13
Fix multiple u-d-i/subiquity/curtin integration issues

dbungert:
* set PY3OR2_PYTHON so that curtin finds the python bin we want it to
* point to a newer subiquity which uses SUBIQUITY_ROOT to find resources, which means we don't have to do a dance with SNAP anymore

### DIFF
--- a/snap/local/subiquity-server
+++ b/snap/local/subiquity-server
@@ -1,8 +1,7 @@
 #!/bin/bash
 
-# store current script and snap directories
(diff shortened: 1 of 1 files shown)


mwhudson:
This looks good to me now we merged the related subiquity PR. At least it should help.

seb128 (approved):
Great work Dan!

PR #168, opened 2021-07-14
Fix keyboard layout variant selection

jpnurmi:
Subiquity defines the default variant as an empty string: https://github.com/canonical/subiquity/blob/main/kbds/C.jsonl

vs.

The variant reported by keyboard_info is null when there's no variant defined in org.gnome.desktop.input-sources.sources:

> For “xkb” sources the second string is “xkb_layout+xkb_variant” or just “xkb_layout” if a XKB variant isn’t needed.


oSoMoN:
Good catch, thanks!

Issue #133, opened 2021-07-07
Proposal: decouple routing logic from pages


jpnurmi:
The proposal is to move routing logic out of wizard pages so that individual pages wouldn't have to care about what comes next.

Requirements:
- Introducing new pages shouldn't cause changes in existing pages.
- Re-ordering pages should be a matter of changing the order in one central place.
- It must be possible to have conditional routes (e.g. if online, skip the connect to internet page).

Example: [flutter_wizard_example](https://github.com/jpnurmi/flutter_wizard_example)

- Three dummy pages + an imaginary service that affects routing.
- Pages call `Wizard.of(context).next()` to request the next page.
- The list of pages and the routing order:
  https://github.com/jpnurmi/flutter_wizard_example/blob/main/lib/routes.dart
- The actual routing logic is implemented using [flow_builder](https://pub.dev/packages/flow_builder) which is based on the Navigator 2.0 Pages API:
  https://github.com/jpnurmi/flutter_wizard_example/blob/main/lib/wizard.dart

Any thoughts?

Feichtmeier:
Sounds great - I also missed a single place for the routes!

Feichtmeier:
Ah one idea: would it be possible to mark one of the routes somehow as "finalized route" or "lastRoute" so the wizardaction on the last page would be an ElevatedButton (yaru green buttons) instead of an OutlinedButton ?

jpnurmi:
@Feichtmeier The example is a bit simplified version. In the real installer, highlighted actions are defined as `WizardAction(highlighted: true)`. There can be any amount of actions and they can do basically anything.

Note to self: naturally, the syntax should be `Wizard.of(context).next()`. The context needs to be inside FlowBuilder and a little shortcut was taken in the example... :)

oSoMoN:
That looks like a sound proposal, thanks @jpnurmi!

What about pages that take arguments (see e.g. https://github.com/canonical/ubuntu-desktop-installer/blob/main/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space_page.dart#L135)?

jpnurmi:
Provider most often eliminates the need of constructor parameters. State/info can be shared via models and services (i.e. Subiquity client, network service, ...).

In this specific case, would it be a bad solution to make the "Write changes to disk" page use `SubiquityClient.getGuidedStorage` instead of passing it as a constructor argument? It's an extra HTTP request but would cut the hard dependency.

[...]

PR #169, opened 2021-07-14
Shared PartitionModel for AllocateDiskSpace and WriteChangesToDiskPage

jpnurmi:
Ref: #147


jpnurmi:
I tried to keep the changes minimal. Especially WriteChangesToDiskPage didn't change much at all. This would be good enough for me to be able to continue on the wizard routing changes #133.

oSoMoN:
This looks good to me. I'm not entirely convinced by the name `PartitionModel`. How about something like `DiskStorageModel`?

PR #173, opened 2021-07-15
Clean up a few analyzer warnings

jpnurmi:


packages/ubuntu_desktop_installer/lib/pages/updates_other_software/updates_other_software_model.dart
```diff
@@ -25,6 +25,7 @@ class UpdateOtherSoftwareModel extends ChangeNotifier {
     notifyListeners();
   }
 
+  // ignore: avoid_positional_boolean_parameters
```
jpnurmi, line 28: proposing to ignore this one because the signature is a perfect pair to CheckboxListTile.onChanged

PR #174, opened 2021-07-15
AppTheme: Add successColor as ThemeData extension

Feichtmeier:
AppTheme: Add successColor as ThemeData extension

The color to indicate success. For example used for
It can be loaded via
```dart
Theme.of(context).successColor
```

Useful to get yaru dependencies out of pages (for example WhoAreYouPage in #144 )


jpnurmi (approved):
Thanks! :+1:

PR #176, opened 2021-07-16
Update subiquity for timezone improvements

dbungert:
Current timezone code has two issues - dryrun can actually try to set
the timezone, and GET /timezone had the unexpected behavior of
triggering a set of the timezone to the system.  Update to a newer
subiquity that has these resolved.


jpnurmi (approved):
Thank you! I was going crazy with the timezone changes.

PR #181, opened 2021-07-19
[tests] replace NavigatorObserver with UI verifications

jpnurmi:
Observing explicit push calls is not compatible with the upcoming wizard routing, which is based on the declarative Navigator 2.0 pages API (#133).


oSoMoN:
LGTM, thanks!

PR #182, opened 2021-07-19
[tests] speedup welcome page tests

jpnurmi:
I'd propose dropping `find.descendant` because for some reason it slows down the test a lot, to a degree that it gets inconvenient to run them as a whole. Plain `find.byType(ListTile)` seems to do the right thing because there's only one `ListTile` for each language. This small change makes a huge difference in the execution time of tests.

Before:
```
$ flutter test
[...]
00:39 +78: All tests passed!
```

After:
```
$ flutter test
[...]
00:05 +78: All tests passed!
```

### DIFF
--- a/packages/ubuntu_desktop_installer/test/welcome_page_test.dart
+++ b/packages/ubuntu_desktop_installer/test/welcome_page_test.dart
@@ -73,14 +73,12 @@ void main() {
           findsOneWidget);
     }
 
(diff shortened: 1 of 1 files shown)


oSoMoN:
Nice one!
I really wonder why `find.descendant` is so slow, that sounds like something isn't right in its implementation. But this change looks perfectly valid and indeed makes a huge difference.

Issue #156, opened 2021-07-12
WSL Screen 2 - Profile setup


jibel:
This is the design for the stage "**Profile setup**" of WSL OOBE. 
Master issue: #160 

Default
![WSL Screen 2 - Identity](https://user-images.githubusercontent.com/1928546/125323410-09b16e80-e33f-11eb-917d-44bf6cdca7f1.png)

Filled
![WSL Screen 2 - Identity filled](https://user-images.githubusercontent.com/1928546/125323423-0f0eb900-e33f-11eb-9e5f-b43b094b55f4.png)

Design specification: [Default](https://people.canonical.com/~platform/desktop/installer/WSL/#s1
), [Filled](https://people.canonical.com/~platform/desktop/installer/WSL/#s5)

jpnurmi:
Should it have input validation similarly to the [Who are you](#40) page?

- error labels for invalid inputs (invalid username, password mismatch)
- green check marks for valid fields
- password strength label

jibel:
On 13/07/2021 09:35, J-P Nurmi wrote:
>
> Should it have input validation similarly to the Who are you <#40> page?
>
>   * error labels for invalid inputs (invalid username, password mismatch)
>   * green check marks for valid fields
>   * password strength label
>
Yes to the 3 points. The pages are identical in behaviour. The main 
difference is that there is no "hostname" for WSL.

jibel:
The "real name" is actually missing from this page. It is a required field.

jpnurmi:
Ok, let's add that right away

PR #185, opened 2021-07-20
Add helper package for determining password quality

jpnurmi:
This allows us to implement a proper password quality checking similarly to gnome-control-center: https://gitlab.gnome.org/GNOME/gnome-control-center/-/blob/master/panels/user-accounts/pw-utils.c#L137

Needed for #40 (Who are you) and #156 (WSL Profile setup).


jpnurmi:
@robert-ancell What do you think, would it be worth it to port https://github.com/libpwquality/libpwquality to Dart?

jpnurmi:
Another option: https://pub.dev/packages/password_strength

robert-ancell (approved):
This looks like a reasonable strategy to me - I don't think we want to go down the path of implementing the libpwquality logic in Dart as this would be a security risk and we could diverge from the gnome-control-center behaviour.

In the future we could consider the password_strength package or writing our own, but it doesn't seem worth it at this point.

jpnurmi:
An alternative based on [password_strength](https://pub.dev/packages/password_strength): https://github.com/canonical/ubuntu-desktop-installer/pull/186

Issue #42, opened 2021-03-08
Screen 16 - Installation Slides


seb128:
The design for screen n°16, Installation Slides

![slides](https://user-images.githubusercontent.com/10974330/110314968-dbc2c300-8008-11eb-9e1c-935e3267b102.png)

Design details https://people.canonical.com/~platform/desktop/installer/16.%20Installation%20Slides/

Feichtmeier:
On this screen there is a big purple progress bar.
With flutter 2.3 it is possible to style the progress indicators independent of the accent color.
I had this in yaru.dart for some days but then decided to revert this until this feature is in the beta channel.
Just FYI

charafau:
Is it possible to get other slide images for reference ?

oSoMoN:
The source for all the slides is [on Launchpad](https://bazaar.launchpad.net/~ubiquity-slideshow/ubiquity-slideshow-ubuntu/html/files/head:/slideshows/ubuntu/slides).

jpnurmi:
@charafau Did you already start working on this?

charafau:
Yes, I will try to upload my code today

jpnurmi:
Awesome, thanks!

jpnurmi:
@charafau Just a heads-up that we are adding a simple temporary installation page to show that something is happening: https://github.com/canonical/ubuntu-desktop-installer/pull/186. We can replace it with the slides when ready. :)

charafau:
thanks for info, hopefully soon! (famous last words)

jpnurmi:
The last remaining piece is to allow expanding and collapsing the terminal output. And maybe resize the window too to make more room for the terminal, like in Ubiquity?

Feichtmeier:
My 2 cents on the terminal:
I always felt that the terminal window inside the installation window looked too techy for the average dude installing Ubuntu. Is the terminal window a must?

jpnurmi:
Even if it's collapsed by default?

Feichtmeier:
> Even if it's collapsed by default?

Yes - when people click there they might think they broke something. Even if looking on the terminal might be the most natural thing in the world for "us" devs/techy people, for normal users this could mean that something went wrong.
I mean is there anything important from the design perspective or in the spec that says that the terminal is needed for the UX? (really curious because I was not part of the shaping process :) )

jpnurmi:
[...]

PR #186, opened 2021-07-20
A temporary installation page

jpnurmi:
The real one will come with slides (#42).

/cc @charafau


jpnurmi:
installation_page

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
```diff
@@ -0,0 +1,56 @@
...
+  bool get isUnknown => _state == ApplicationState.UNKNOWN;
+
+  /// Whether an error has occurred.
+  bool get hasError => _state == ApplicationState.ERROR;
+
+  /// Whether the installation process is active.
+  bool get isInstalling => !isDone && !hasError;
+
+  /// The current installation step.
+  int get currentStep => _state.index;
```
oSoMoN, line 27: This works well when using this getter to display a progress label in the form `i/n` only because by chance the two values that are being excluded are `UNKNOWN` (index 0) and `ERROR` (index n-1). It might be worth explaining it in a comment?

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
```diff
@@ -0,0 +1,56 @@
...
+  bool get isDone => _state == ApplicationState.DONE;
+
+  /// Whether the installation state is unknown.
+  bool get isUnknown => _state == ApplicationState.UNKNOWN;
+
+  /// Whether an error has occurred.
+  bool get hasError => _state == ApplicationState.ERROR;
+
+  /// Whether the installation process is active.
+  bool get isInstalling => !isDone && !hasError;
```
oSoMoN, line 24: There should probably be an additional AND clause to this condition: `_state.index >= ApplicationState.RUNNING` (before that state, it's not really installing).

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
```diff
@@ -0,0 +1,56 @@
...
+
+  /// The total number of installation steps (excludes unknown & error states).
+  int get totalSteps => ApplicationState.values.length - 2;
+
+  ApplicationState get _state => _status.state ?? ApplicationState.UNKNOWN;
+
+  void _updateStatus(ApplicationStatus status) {
+    if (_status == status) return;
+    _status = status;
+    notifyListeners();
```
oSoMoN, line 37: For debugging purposes, what do you think of printing to the console, something like `"subiquity status changed to $status"`?

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_page.dart
```diff
@@ -0,0 +1,66 @@
...
+              textAlign: TextAlign.center,
+            ),
+          ],
+        ),
+      ),
+      actions: <WizardAction>[
+        WizardAction(
+          label: 'Restart',
+          enabled: model.isDone || model.hasError,
+          onActivated: model.reboot,
```
oSoMoN, line 61: As discussed elsewhere, let's close the app after requesting subiquity to reboot (especially useful in dry run mode, where rebooting doesn't actually order the machine to restart but shuts down the subiquity server).

oSoMoN (changes requested):

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
jpnurmi: what would be a good description for that state? "Preparing..."?

oSoMoN: "preparing" sounds reasonable to me.

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_page.dart
jpnurmi: :+1: will fix when we get a response from subiquity

jpnurmi: i made it exit without waiting for a response that it never gets. we can add response handling when the client provides it

oSoMoN: SGTM, thanks.

oSoMoN:
LGTM, thanks!

PR #187, opened 2021-07-21
Password encryption & strength estimation

jpnurmi:
There is a need to estimate password strength for #40 (Who are you) and \#156 (WSL Profile setup).

The strength estimation is based on a 3rd-party [password_strength](https://pub.dev/packages/password_strength) package.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/utils.dart
+++ b/packages/ubuntu_desktop_installer/lib/utils.dart
@@ -1,1 +1,2 @@
 export 'utils/equal_validator.dart';
+export 'utils/password.dart';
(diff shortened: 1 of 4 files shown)


jpnurmi:
Out of curiosity, I gave it a shot with the `password_strength` package. There are some clear benefits - the implementation is _a lot_ simpler than #185 (`libpwquality` FFI bindings) and there are no CI issues. I'm leaning towards this version due to simplicity, but WDYT?

jpnurmi:
Rebased and squashed to make this easier to use as a dependency in other PRs.

jpnurmi:
There are two pages waiting for this so I'm going ahead and merge this. We can revisit this later if necessary.

PR #189, opened 2021-07-22
Remove UbuntuDesktopInstallerApp dependencies from individual pages

jpnurmi:
WSL OOBE wizard is going to reuse the `WelcomePage`, but the app will be different with another set of routes. Providing the app locale helps to eliminate the dependency to `UbuntuDesktopInstallerApp`.

Locale and theme are similar in the way that changing either rebuilds the whole app, so the proposal is to bundle them together as app-wide settings to avoid yet another tiny provider (if theme & locale were provided separately).


jpnurmi:
Rebased to resolve conflicts

packages/ubuntu_desktop_installer/lib/settings.dart
```diff
@@ -0,0 +1,58 @@
...
+  /// changes (the entire app will be rebuilt anyway).
+  static Settings of(BuildContext context, {bool listen = true}) {
+    return Provider.of<Settings>(context, listen: listen);
+  }
+
+  /// The current application theme.
+  ThemeMode get theme => _theme;
+  ThemeMode _theme = ThemeMode.system;
+
+  /// Applies a theme matching the given [brigtness].
```
robert-ancell, line 26: typo in "brightness".

packages/ubuntu_desktop_installer/lib/main.dart
```diff
@@ -9,13 +9,13 @@ import 'package:subiquity_client/subiquity_client.dart';
...
 import 'app.dart';
-import 'app_theme.dart';
 import 'disk_storage_model.dart';
 import 'keyboard_service.dart';
 import 'l10n/app_localizations.dart';
+import 'settings.dart';
 
 Future<void> main() async {
-  final themeSettings = GSettings(schemaId: 'org.gnome.desktop.interface');
+  final gsettings = GSettings(schemaId: 'org.gnome.desktop.interface');
```
robert-ancell, line 18: This is probably better named "interfaceSettings".

robert-ancell (approved):
Aside from the small nits LGTM.

Issue #158, opened 2021-07-12
WSL Screen 4 - Configuration UI


jibel:
This is the design for the stage "Configuration UI" of WSL OOBE

Default
![WSL Screen 5 - Configuration UI](https://user-images.githubusercontent.com/1928546/125324400-2ac68f00-e340-11eb-94a0-6245efe8091c.png)

Filled
![WSL Screen 5 - Configuration UI filled](https://user-images.githubusercontent.com/1928546/125324414-2d28e900-e340-11eb-8ca4-a91f920fd707.png)

Design specification: [Default](https://people.canonical.com/~platform/desktop/installer/WSL/#s4), [Filled](https://people.canonical.com/~platform/desktop/installer/WSL/#s7)

jpnurmi:
The description for "Mount `/etc/fstab`" is incomplete:

![125324400-2ac68f00-e340-11eb-94a0-6245efe8091c](https://user-images.githubusercontent.com/140617/130071025-35eff26b-2d16-48b1-9596-cc9ba0beadad.png)

jibel:
On 19/08/2021 14:50, J-P Nurmi wrote:
>
> The description for "Mount |/etc/fstab|" is incomplete:
>
> 125324400-2ac68f00-e340-11eb-94a0-6245efe8091c 
> <https://user-images.githubusercontent.com/140617/130071025-35eff26b-2d16-48b1-9596-cc9ba0beadad.png>
>
It's an excerpt from the man page:

"The file fstab contains descriptive information about the filesystems 
the system can mount."

PR #193, opened 2021-07-27
WizardPage: expose header & footer padding

jpnurmi:
Needed for WSL Configuration UI (#158) which has no common page header but separate section headers.


robert-ancell (approved):
LGTM!

PR #194, opened 2021-07-27
Return future from SubiquityClient.close() to allow await in tests

jpnurmi:
Allow await until the socket is cleanly closed. For testing WSL, the client needs to be closed and the server restarted in the middle of the tests


MarcusTomlinson:
Thanks 👌

PR #195, opened 2021-07-28
Add Subiquity endpoints and data classes for WSL configs

jpnurmi:
/cc @patrick330602

### DIFF
--- a/packages/subiquity_client/lib/src/types.dart
+++ b/packages/subiquity_client/lib/src/types.dart
@@ -220,3 +220,39 @@ class StorageResponse with _$StorageResponse {
   factory StorageResponse.fromJson(Map<String, dynamic> json) =>
       _$StorageResponseFromJson(json);
 }
+
+@freezed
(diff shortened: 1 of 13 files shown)


jpnurmi:
For the time being, this requires the [WSL feature branch](https://github.com/ubuntu/subiquity/tree/wsl/main) of Subiquity.

jpnurmi:
Notes:
- The WSL tests are in the same `subiquity_client_test.dart` file because `flutter test` runs multiple parallel test jobs by default. If the tests were in different files, they would fail unless the `-j1` command-line argument was passed.
- The test file diff looks exhausting but most of it is indentation changes for grouping the existing tests. The two groups have a different setup for the server.

jpnurmi:
I have removed all server changes etc. and left the endpoints and data classes only. This is enough for being able to run mocked tests for WSL pages and models.

jpnurmi:
Ha, I got caught right away 😁

oSoMoN:
I cannot comment on the various fields of the configuration data structures, but the change otherwise looks sound.

PR #197, opened 2021-07-30
Add `runWizardApp()` to allow multiple entry-points

jpnurmi:
For example:
- `main.dart` runs `UbuntuDesktopInstallerApp`
- `main_wsl.dart` runs `UbuntuOobeWizardApp`

By default, Flutter chooses the `lib/main.dart`. Another entry-point can be specified via command-line (or chosen in the IDE, of course).

For example:
- `flutter run --target lib/main_wsl.dart`
- `flutter build linux --target lib/main_wsl.dart`


jpnurmi:
The diff looks big but it's kind of just swapping code around. :slightly_smiling_face:

Also, in the long term, we probably want to split the apps. It's a bit bigger task since then we need a shared package too because the two apps have a lot in common. This should be a good enough starting point as it makes it possible to build and run either wizard.

oSoMoN:
LGTM, thanks!

PR #199, opened 2021-07-30
Allocate disk space improvements

jpnurmi:
- made data table scrollable
- indented partitions
- added a simple bar to visualize partitions


jpnurmi:
| Before | After |
|---|---|
| allocate-disk-space-before | allocate-disk-space-after |

jpnurmi:
By the way, is there's going to be a pre-defined set of partition colors? I made it go around the color wheel for now.

oSoMoN:
> By the way, is there's going to be a pre-defined set of partition colors? I made it go around the color wheel for now.

That needs to be discussed with design (and the outcome documented). @long-chung ^

oSoMoN:
This looks much better than the initial dummy implementation, thanks!

PR #200, opened 2021-08-03
Remove default identity in favor of the Who are you page

jpnurmi:
The comment suggests that this could be removed now that the _Who are you_ page exists:

> Define a default identity until a UI page is implemented for it.

The commit message of a8c5f66 makes me feel a bit unsure, though:

> Instruct subiquity to use default values for various endpoints earlier in the app's lifecycle.

Are these contradicting or did I misunderstand? Can the default identity be cleaned up?

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/main.dart
+++ b/packages/ubuntu_desktop_installer/lib/main.dart
@@ -38,15 +38,6 @@ Future<void> main() async {
     'timezone',
   ]);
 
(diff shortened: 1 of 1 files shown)


oSoMoN:
Yes that's fine. The confusing commit message was just moving the call to `setIdentity()` with a default identity earlier in the app's lifecycle, but now that we have a proper UI page for it it needs to be removed.

Issue #36, opened 2021-03-08
Screen 11 - Allocate Disk Space


seb128:
The design for screen n°11, Allocate Disk Space

![space](https://user-images.githubusercontent.com/10974330/110314286-f183b880-8007-11eb-8082-529bcaac31b4.png)

Design details https://people.canonical.com/~platform/desktop/installer/11.%20Allocate%20Disk%20Space/

felagund:
Would be great if there was an option to use systemd-boot as the Boot loader. It saves me 4 seconds every boot. GRUB2 is slow.

galgalesh:
I think the vertical layout might be confusing because Windows and other distros show these horizontally. Having the installer also show these horizontally has the advantage that users will be better able to link what they see in Windows to what they see in this installer.

I think it might also be useful to show more explicitly that these partitions are part of a single disk. For example, by drawing a rectangle around all partitions.

madsrh:
All mockups are a bit unbalanced / asymmetric IMHO, but this one is the most confusing one to look at 😄 
Why isn't the length of the drive selection dropdown centered in the window? It's seems strange that it's not half the length and not full 🤷‍♂️ 

Perhaps [Elementary](https://www.fosslinux.com/3432/elementary-os-to-get-cool-new-installer-possible-debut-in-juno.htm) could be an inspiration?

![image](https://user-images.githubusercontent.com/3986894/110596013-13537b80-817f-11eb-8561-b7ec1cdf4daf.png)

long-chung:
There is a before and after state on this screen.
@galgalesh the 2 states are in different columns hence why there is no border around the whole of the sum.
@madsrh same point as above

long-chung:
@galgalesh You've raised some excellent points. Thank you for sharing

galgalesh:
A final note; I just noticed that in the "before" and "after" graphics, the size of the Windows 10 partition does not change even though the size of the Windows rectangle changes. I was convinced this setup was going to shrink the Windows partition until I read the numbers.

Mannshoch:
As an Idea, 
why not preset wished mount points and **after that** allow to design the Disks. It may allow the installation manager to help not to accidentally remove or formatting a partition you wish to use.

**See Mockup:**
![grafik](https://user-images.githubusercontent.com/7795891/111607362-72953980-87d8-11eb-94b0-fcb0da3f4131.png)

seb128:
[...]

PR #202, opened 2021-08-04
Allocate disk space: add partition legend

jpnurmi:
allocate-disk-space-disk-legend

Design: #36


oSoMoN:
That looks great, thanks!

PR #203, opened 2021-08-04
Fix an exception and an assertion on the "Write changes to disk" page.

oSoMoN:
Those happen when the storage config has unexpected values, and can be reproduced in dry run mode by using the examples/existing-partitions.json machine config for the subiquity server.
Unhandled storage config types include:
 - lvm_volgroup
 - lvm_partition
 - raid


jpnurmi (approved):
LGTM :+1:

PR #205, opened 2021-08-05
Allocate disk space: add remaining UI elements

jpnurmi:
Note: the new buttons don't do anything yet.

Screenshot from 2021-08-05 13-44-49

Design: #36


oSoMoN:
Thanks for reworking the models, this is a much sounder approach.

PR #206, opened 2021-08-05
subiquity_client: define PYTHONPATH

jpnurmi:
To ensure correct versions of curtin and probert, as discussed in canonical/subiquity#1019.


dbungert (commented):
Yes, this is exactly the concept I had in mind.
I'll let someone who has written dart before ponder the actual logic :)

PR #207, opened 2021-08-06
Installation slides

charafau:
Took me some time and not finish yet, because I ran into some design problems (there's no design for individual slides so I tried to copy values from existing installer)

Stuff needs to be done
- [x] cancel slide animation when user clicks next slide
- [ ] bind this view with notifier to display status, update progress
- [ ] how should background of slide behave (what fit it should have, currently it moves a litle bit when termianl opens)
- [ ] change current back/next buttons to the onces from original installer
- [x] animate slide title and update it

I upload few screenshots of current state

Screenshot from 2021-08-06 03-36-50

Screenshot from 2021-08-06 03-36-55

Screenshot from 2021-08-06 03-36-59

Screenshot from 2021-08-06 03-36-47


jpnurmi:
@long-chung, are you able to help with the design questions above?

long-chung:
@charafau @jpnurmi this page was always intended to act and display the same way that it does in the current installer.
The intention was to minimise development time and allow for any future revision should content/marketing/brand decide to update content.
Hope this helps.

charafau:
@long-chung should I change back and forward buttons to look like they do on current installer?

long-chung:
@charafau Yes lets revert back the prev/back buttons to how they are in the current installer.
They seem to get lost and obscure content below.
Also you might need to move the left content area in a bit more so the text doesn't get obstructed by the prev button.

Many thanks

jpnurmi:
Hi @charafau, I gave the latest version a quick try. These are probably already known, but here's a list of notes without having looked at the code:

- Some of the slides have different sizes. Perhaps @long-chung can help with this?
- Some text labels are missing margins and cramped at the slide edges.
- The buttons still need re-styling. If I understood correctly, we decided to go with the white half circles like in the old installer.
- There's something wrong with the slide timer, the slide show goes wonky after pressing the buttons back and forth. The behavior feels like if it would start new timers without canceling the old ones. Could that be the case?

https://user-images.githubusercontent.com/140617/136455022-d8fd79bc-21e9-4674-905c-19ce932703ce.mp4

charafau:
Yes, still work in progress. Haven't fixed slide designs yet, will fix them next.

Thanks for timer issue, that should be fixed. I'm updating to do list in first comment.

Things go slowly for to my lack of time but hopefully soon™️

jpnurmi:
I'm closing this out of the way because we've come a long way since this PR stalled and it serves to purpose to keep this open. Thank you for all the time and effort you put into this!

PR #208, opened 2021-08-06
Pass subiquity a full locale in the form language_COUNTRY.UTF-8.

### LINKED ISSUE #175: Locale not being set on target system
It looks like cloud-init wants locale in the following format
language_COUNTRY.UTF-8

Please change from the current, which only sends language as a 2 letter code, with no _COUNTRY after it.

oSoMoN:
Fixes #175.


oSoMoN:
The diff looks artificially large because git isn't clever when detecting files being moved, but the actual changes are rather trivial.

jpnurmi (approved):
LGTM :+1:

PR #210, opened 2021-08-09
Subiquity client: enable keepalive packets

jpnurmi:
Ref: canonical/subiquity#1022

### DIFF
--- a/packages/subiquity_client/lib/src/http_unix_client.dart
+++ b/packages/subiquity_client/lib/src/http_unix_client.dart
@@ -6,6 +6,12 @@ import 'dart:typed_data';
 import 'package:http/http.dart';
 import 'package:synchronized/synchronized.dart';
 
+const SOL_SOCKET = 1; // socket.h
+const SO_KEEPALIVE = 9; // socket.h
+
+const IPPROTO_TCP = 6; // netinet/in.h
+const TCP_KEEPINTVL = 5; // netinet/tcp.h
+
 enum _HttpParserState {
   status,
   header,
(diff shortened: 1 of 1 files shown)


jpnurmi:
Hold on, this does not work as expected.

packages/subiquity_client/lib/src/http_unix_client.dart
```diff
@@ -60,7 +66,15 @@ class HttpUnixClient extends BaseClient {
     if (_socket == null) {
       var address = InternetAddress(path, type: InternetAddressType.unix);
       _socket = await Socket.connect(address, 0);
-      _socket?.listen(_processData);
+
+      final keepalive =
+          RawSocketOption.fromBool(SOL_SOCKET, SO_KEEPALIVE, true);
+      _socket!.setRawOption(keepalive);
+
+      final keepintvl = RawSocketOption.fromInt(IPPROTO_TCP, TCP_KEEPINTVL, 60);
```
jpnurmi, line 74: Hard to see what happens under the hood in Dart, but I get "Operation not supported" with a minimal C test app:
```
int keepintvl = 10;
int rc = setsockopt(fd, IPPROTO_TCP, TCP_KEEPINTVL, &keepintvl, sizeof(int));
printf("TCP_KEEPINTVL: %d (%s)\n", rc, strerror(errno)); // TCP_KEEPINTVL: -1 (Operation not supported)
```

jpnurmi:
For what it's worth, I tried `Connection: keep-alive` and `Keep-Alive: timeout=foo` HTTP headers too without any success. Am I on the wrong path here? Is manually re-opening the socket the correct way to go?

MarcusTomlinson:
> For what it's worth, I tried `Connection: keep-alive` and `Keep-Alive: timeout=foo` HTTP headers too without any success. Am I on the wrong path here? Is manually re-opening the socket the correct way to go?

I don’t really think this is all that worth pursuing now that we’ve upped the timeout.

mwhudson:
Uh. We're using unix domain sockets to communicate so it's not surprising that trying to set TCP keep alives fails. Even if we were using TCP, they are at a different level: the TCP keep alive will keep the socket open at the TCP level but it will still appear to be idle to the http server which will still close the connection after its timeout is reached.

I think it's a general HTTP thing that if you try to reuse a socket and it turns out the server has closed the connection you should transparently retry.

jpnurmi:
Right, and at first, I even thought it was working because the socket was no longer closed after a timeout. I just didn't realize the socket option call was hanging. :smile: /me goes back to tweaking pixels

mwhudson:
I guess if we continue to have problems with the http client we can change subiquity to listen on a TCP socket instead / as well and then the desktop installer can use the real dart http client implementation, which probably gets all of the million details right (and we can all subscribe to https://github.com/dart-lang/sdk/issues/42716).

PR #211, opened 2021-08-09
Subiquity client: synchronize socket creation

jpnurmi:
This change fixes the issue that multiple sockets were created on startup. Concurrent requests (mark configured, keyboard, locale...) ended up awaiting a new socket connection because the `_socket` member variable is set after the socket has connected.


jpnurmi:
I found this by accident while debugging the socket timeout issue. I noticed that `Socket.onDone` was getting called three times, whereas it should be only two - once for the client and once for the server.

MarcusTomlinson (approved):
Ah good catch!

PR #213, opened 2021-08-10
Add support for command line arguments

jpnurmi:
So far, `--dry-run` and `--machine-config <config>` are supported. The WSL OOBE wizard will later replace `--machine--config` with `--reconfigure`.


packages/ubuntu_desktop_installer/lib/app.dart
```diff
@@ -71,3 +70,57 @@ Future<void> runWizardApp(
...
+ArgResults? parseCommandLine(
+  List<String> args, {
+  bool? showMachineConfig,
+  @visibleForTesting io.IOSink? out,
+  @visibleForTesting void Function(int exitCode) exit = io.exit,
+}) {
+  final parser = ArgParser();
+  parser.addFlag('help', abbr: 'h', negatable: false);
+  parser.addFlag('dry-run',
+      defaultsTo: io.Platform.environment['LIVE_RUN'] != '1',
```
MarcusTomlinson, line 84: Seems redundant even to have this flag. Whether you pass `--dry-run` or not, it'll run dry right?

MarcusTomlinson: Oh I see, it's necessary to set `options['dry-run']` here for later use in actually assigning the server mode.

MarcusTomlinson (approved):
Happy with this. Thanks!

PR #214, opened 2021-08-10
Update Yaru & adapt to changes

jpnurmi:
Explicit input decoration borders are no longer required, because Yaru
provides them via input decoration according to the design.


oSoMoN:
LGTM, thanks!

Issue #31, opened 2021-03-08
Screen 6 - Updates and other software


seb128:
The design for screen n°6, Updates and other software

![update](https://user-images.githubusercontent.com/10974330/110313678-10ce1600-8007-11eb-9b09-18f0e19bb360.png)

Design details https://people.canonical.com/~platform/desktop/installer/6.%20Updates%20and%20Other%20Software/

pieqq:
A suggestion here:

the "Install third-party software ..." option is coming from the current installer, but I think it's always been a bit confusing, especially since Secure Boot started being a thing: why would I need to input a security password for Secure Boot to install "additional media formats"?! (that's honestly the feeling I've had ever since this option has appeared in the installer)

I think this option was originally to install non-free packages, but it has evolved into handling Secure Boot key enrollment in order to install proprietary DKMS (nvidia drivers) that do not use the default security keys stored on the device.

It would be good to separate both things:

- Install 3rd party drivers for graphics and WiFi hardware (_These drivers are subject to license terms included with their documentation. They are proprietary._)
- Install support for additional media formats (_This software is subject to license terms included with its documentation. Some is proprietary._)

Bonus point if some help (or a link to the Ubuntu Wiki) can be added to explain exactly what these "graphics and WiFi drivers" and "additional media formats" are...

fcore117:
Really hope that Minimal this time is **really** minimal without extraweight packages and so on and installer do not act counter-productive by installing all first and then uninstalling like now, but instead it should install minimally from beginning. Whole world needs a real true minimal barebone desktop that has none of extraweight packages from braille to unneeded fonts and local help files. Third option is to add ultra minimal installation option that has truly pure and barebone desktop and extra fonts and languages are only installed based on language selected by user.

Feichtmeier:
I think this could be closes as #88 has been merged :)

jpnurmi:
Is there not going to be a _"Download updates while installing Ubuntu"_ checkbox like in Ubiquity? /cc @elioqoshi

![image](https://user-images.githubusercontent.com/140617/144043378-c0924149-6103-4e57-9ad1-546e8867d2dc.png)

PR #215, opened 2021-08-10
Sync Updates and Other Software with the design

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2021-08-10 14-08-54 | Screenshot from 2021-08-10 14-08-30 |

Design: #31


oSoMoN:
LGTM, and thanks for adding missing tests!

PR #217, opened 2021-08-10
Correct dropdown looks

jpnurmi:
DropdownButtonFormField uses themed input decorations (with borders from Yaru).

| Before | After |
|---|---|
| Screenshot from 2021-08-10 20-30-52 | Screenshot from 2021-08-10 20-34-40 |


robert-ancell (approved):
Code LGTM. I'm assuming the change is the desired design we are going for.

jpnurmi:
Thanks! For reference, here's a design example with a dropdown at the bottom:

image

PR #218, opened 2021-08-11
Create partition dialog

jpnurmi:
Screenshot from 2021-08-13 16-40-39

Issue: #36
Design: https://people.canonical.com/~platform/desktop/installer/11.%20Allocate%20Disk%20Space/#artboard2


jpnurmi:
Note: depends on PR #205

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_page.dart
```diff
@@ -1,6 +1,8 @@
 import 'package:filesize/filesize.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
```
oSoMoN, line 4: That import can now be converted to `import '../../l10n.dart';`.

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_page.dart
```diff
@@ -1,6 +1,8 @@
...
+import 'package:flutter_spinbox/material.dart';
```
oSoMoN, line 5: Shouldn't that import be in `allocate_disk_space_dialogs.dart`, where a `SpinBox` is instantiated?

Feichtmeier:
Nice dialog!
I feel like there is a slight misalignment here
grafik

jpnurmi:
> I feel like there is a slight misalignment here

Thanks, good eye! :) Indeed, according to the spec, the bounds of the inner Radio should be aligned with the input fields. Currently, it's the entire RadioListTile, which has some internal content padding.

I'm struggling a bit with the hover highlight for the list tiles because it looks weird no matter if it's aligned with the input fields or if it stretches all the way to the edge of the dialog. I wonder if we could do something on a global theming level in yaru.dart, or if we should try to override the highlight for specific list tiles. What do you think?

| | |
|---|---|
| Screenshot from 2021-08-13 11-52-15 | Screenshot from 2021-08-13 11-52-15 |

jpnurmi:
This is how it's aligned in the spec. Notice that it's not aligned by the inner visual of the radio indicator, but by its bounds:

Screenshot from 2021-08-13 11-55-10

Feichtmeier:
Looks indeed better :+1:  I think alignment by bounds is fine
I would not put this into the theme though :thinking:

jpnurmi:
To be clear, for theming I only meant specifying or rather removing the hover highlight:

Screenshot from 2021-08-13 11-52-15

Such hover is not present in any of the designs I've seen so far.

@long-chung, should radio buttons have any background hover highlight? The round highlight for the indicator is probably fine because it's even used as a reference for alignment, but the highlighted background looks strange anywhere besides list views.

long-chung:
@jpnurmi hover is only present over the actual radio as per the spec and material design guidelines.
Screenshot 2021-08-13 at 12 04 45

jpnurmi:
@long-chung, thanks! The whole area including the label is interactive, though?

@Feichtmeier Perhaps it's best for us to implement a custom radio button without a background hover effect because the global theme cannot know the context where radio tiles are used.

long-chung:
@jpnurmi Yes, the whole area around the label is interactive. Many thanks.

packages/ubuntu_desktop_installer/lib/l10n/app_en_US.arb
```diff
@@ -70,6 +72,21 @@
   "newPartitionTable": "New partition table",
   "bootLoaderDevice": "Device for boot loader installation",
 
+  "partitionCreateTitle": "Create partition",
+  "partitionSizeLabel": "Size (MB):",
```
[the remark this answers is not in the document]
jpnurmi: I took a little shortcut because I'm having big problems getting SpinBox to behave when switching units. :(

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_page.dart
jpnurmi: `_dialogs.dart` and `_widgets.dart` are `part of` `_page.dart`.

it's a trick to be able to keep the widgets private and avoid documenting them. i'll make them standalone files instead... :P

packages/ubuntu_desktop_installer/lib/l10n/app_en_US.arb
```diff
@@ -70,6 +72,21 @@
...
+  "partitionTypeLabel": "Type for the new Partition:",
+  "partitionTypePrimary": "Primary",
+  "partitionTypeLogical": "Logical",
+  "partitionLocationLabel": "Location for the new Partition:",
+  "partitionLocationBeginning": "Beginning of this space",
+  "partitionLocationEnd": "End of this space",
+  "partitionFormatLabel": "Used as:",
+  "partitionFormatExt4": "Ext4 (recommended)",
+  "partitionFormatExt3": "Ext3",
+  "partitionFormatExt2": "Ext2",
```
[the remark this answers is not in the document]
jpnurmi: I added the same options for now until someone confirms the list.

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
```diff
@@ -0,0 +1,326 @@
...
+                            );
+                          },
+                        ),
+                      ),
+                      const SizedBox(height: kContentSpacing),
+                      ConstrainedBox(
+                        constraints: BoxConstraints(maxHeight: tileHeight),
+                        child: Autocomplete<String>(
+                          initialValue: TextEditingValue(
+                            text: lang.partitionMountPointAutomatic,
```
oSoMoN, line 139: This makes use of https://github.com/flutter/flutter/pull/80257, which isn't available on flutter stable yet.

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
```diff
@@ -0,0 +1,326 @@
...
+          ],
+        );
+      },
+    );
+  }
+}
+
+extension _PartitionUnitLang on DataUnit {
+  String localize(BuildContext context) {
+    final lang = AppLocalizations.of(context)!;
```
oSoMoN, line 279: No need for null checking here, as the AppLocalizations getter is non nullable since https://github.com/canonical/ubuntu-desktop-installer/commit/658a7aeca1693170baf42cc8d5b4dc03286c578f.

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
```diff
@@ -0,0 +1,326 @@
...
+        return lang.partitionUnitMB;
+      case DataUnit.gigabytes:
+        return lang.partitionUnitGB;
+    }
+  }
+}
+
+extension _PartitionFormatLang on PartitionFormat {
+  String localize(BuildContext context) {
+    final lang = AppLocalizations.of(context)!;
```
oSoMoN, line 295: Same here, no need for null checking.

oSoMoN (changes requested):
Lost mostly good to me, but will require a change to run with flutter from the stable channel.

oSoMoN (approved):

oSoMoN:
All good now, thanks!

PR #223, opened 2021-08-12
Add convenience import for l10n too

jpnurmi:


jpnurmi:
As you can see, I love these forwarding imports... :grin: Easier to remember and looks prettier than `flutter_gen/foo` and/or `l10/foo`, and no need to stop to think about what comes and from where. :wink:

PR #226, opened 2021-08-12
Ubuntu WSL OOBE entry-point

jpnurmi:
`flutter run -t lib/main_wsl.dart` or select `main_wsl.dart` as run target in an IDE.


jpnurmi:
Perhaps this intermediate step could be skipped and we could go straight to a separate app package.

jpnurmi:
> Perhaps this intermediate step could be skipped and we could go straight to a separate app package.

Started with https://github.com/canonical/ubuntu-desktop-installer/pull/231

PR #227, opened 2021-08-12
Factor out the LangTester extension on the WidgetTester class.

oSoMoN:


jpnurmi (approved):

jpnurmi:
How would you prefer to share these test extensions between the packages? Moving them to the shared `ubuntu_wizard` package is not good because that would make `flutter_test` a regular dependency for the app packages. Would it be overkill to create a separate package for test utils alone?

Issue #160, opened 2021-07-12
WSL OOBE Design specification


jibel:
This is the master issue to track the work on the design of the GUI of WSL OOBE.

The OOBE starts on the first boot of a WSL instance. It is similar, in concept, to the OEM setup of a current desktop installation. The OOBE is a first boot wizard which guides the user through the initial configuration of the WSL instances and helps him to:
- Select a language,
- Configure the default user,
- Configure various advanced options of WSL.

The OOBE can be executed afterwards to reconfigure WSL.

It is possible to preseed the OOBE and call it in automatic mode for automated setup of WSL.

The OOBE relies on the same components as the flutter based Ubuntu Desktop Installer such as Subiquity .

In first boot configuration mode the following stages are displayed:
- #155 
- #156 
- #157 
- #159 

In system reconfiguration mode (when the default user already exists) the following stages are displayed:
- #157 
- #158

jpnurmi:
It's not a pixel-perfect replica of the design, but here's the current status (UI parts only, no actual functionality under the hood).

![wsl](https://user-images.githubusercontent.com/140617/125471383-5b169a15-0cef-4581-9db8-ec4b569d5b6c.gif)

I'm going to help with #144 so we can use the validated input field @Feichtmeier made.

jpnurmi:
Feel free to reopen if there's anything missing. :)

PR #231, opened 2021-08-16
Move common parts to a separate ubuntu_wizard package

jpnurmi:
Will be shared with the WSL setup (#160).


jpnurmi:
https://github.com/canonical/ubuntu-desktop-installer/pull/232 helps with the formatting issue

oSoMoN:
How are localizations in this new package handled?
I modified `packages/ubuntu_wizard/lib/src/l10n/ubuntu_fr_FR.arb` to add translations for the password strength related strings, but `packages/ubuntu_wizard/lib/src/l10n/ubuntu_localizations_fr.dart` wasn't re-generated, and as a consequence my translations weren't picked up in the UI.

jpnurmi:
> How are localizations in this new package handled?

Sorry, I forgot to explain this. In short: 1) edit .arb, 2) run flutter gen-l10n, and 3) commit lib/src/l10n.

Unlike ubuntu_desktop_installer's localizations, ubuntu_wizard's localizations are not auto-generated into a hidden build directory. The localizations must be available/exported from the package to the app, so they are generated into the source tree.

I wonder if we could actually do this for the app too because the auto-generation is either somehow unreliable or Dart-Code has problems resolving the generated files (need to run pub get multiple times). Have you had that problem too?

packages/ubuntu_wizard/linux/ubuntu_wizard_plugin.cc
```diff
@@ -0,0 +1,94 @@
...
+static void on_method_call(FlMethodChannel* channel, FlMethodCall* method_call,
+                           gpointer user_data) {
+  FlPluginRegistrar* registrar = FL_PLUGIN_REGISTRAR(user_data);
+  ubuntu_wizard_plugin_handle_method_call(registrar, method_call);
+}
+
+static gboolean on_delete_event(GtkWidget* /*window*/, GdkEvent* /*event*/,
+                                gpointer user_data) {
+  g_autoptr(GError) error = nullptr;
+  FlValue* event = fl_value_new_string("deleteEvent");
```
robert-ancell, line 64: This should be `g_autoptr(FlValue) event = fl_value_new_string("deleteEvent")`.

packages/ubuntu_wizard/lib/src/l10n/ubuntu_nl_NL.arb
```diff
@@ -0,0 +1,3 @@
+{
+  "@@locale": "nl_NL"
+}
```
robert-ancell, line 3: Nit: A number of these translation files don't have trailing newlines.

packages/ubuntu_wizard/lib/src/l10n/ubuntu_localizations.dart
```diff
@@ -0,0 +1,223 @@
...
+/// dependencies:
+///   # Internationalization support.
+///   flutter_localizations:
+///     sdk: flutter
+///   intl: any # Use the pinned version from flutter_localizations
+///
+///   # rest of dependencies
+/// ```
+///
+/// ## iOS Applications
```
robert-ancell, line 50: Is this copy-pasted from somewhere?

packages/ubuntu_wizard/linux/ubuntu_wizard_plugin.cc
robert-ancell: (This is a bug in the existing code).

robert-ancell (approved):
Other than the comments given, LGTM.

packages/ubuntu_wizard/lib/src/l10n/ubuntu_localizations.dart
jpnurmi: auto-generated by `flutter gen-l10n`: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/localizations/gen_l10n_templates.dart#L54

oSoMoN:
> > How are localizations in this new package handled?
> 
> Sorry, I forgot to explain this. In short: 1) edit .arb, 2) run flutter gen-l10n, and 3) commit lib/src/l10n.

Ack. This should be documented in the "Translations" section of CONTRIBUTING.md.


> Unlike ubuntu_desktop_installer's localizations, ubuntu_wizard's localizations are not auto-generated into a hidden build directory. The localizations must be available/exported from the package to the app, so they are generated into the source tree.
> 
> I wonder if we could actually do this for the app too because the auto-generation is either somehow unreliable or Dart-Code has problems resolving the generated files (need to run pub get multiple times). Have you had that problem too?

Yes, I see that from time to time. I have no strong opinion on this, we can try versioning the generated localizations in the source tree and see how (in)convenient it is.

jpnurmi:
> Yes, I see that from time to time. I have no strong opinion on this, we can try versioning the generated localizations in the source tree and see how (in)convenient it is.

Yeah, it's easy to forget to update. What do you think about having a CI check that re-generates all generated files and checks if there are any changes?

oSoMoN:
> > Yes, I see that from time to time. I have no strong opinion on this, we can try versioning the generated localizations in the source tree and see how (in)convenient it is.
> 
> Yeah, it's easy to forget to update. What do you think about having a CI check that re-generates all generated files and checks if there are any changes?

Good idea!

Issue #224, opened 2021-08-12
System restart doesn't work


jpnurmi:
Steps:
- Finish installation
- Press restart

Result:
- Expected: the system restarts
- Actual: UI exits without restarting the system

It should help if Subiquity is able to ack the reboot request in the future, so the UI can wait until the request has been received.

mwhudson:
The POST you do to subiquity to request a restart doesn't return until /sbin/reboot has been called. I'm not sure what you mean by "acking the reboot request"?

jpnurmi:
I didn't dig deeper yet, but I'd guess subiquity never receives the request because the UI exits too soon. Flushing the socket didn't seem to help, though. If subiquity would send a response, the UI would know that the request has been received. Does that make sense?

mwhudson:
Not really, no :)

The flow in the text mode client is: the user clicks reboot, this starts a POST request to /reboot and sets up the UI to indicate that the reboot has been requested and waits until a response is received, then the client exits. Is there some reason the graphical client can't do this?

jpnurmi:
> The flow in the text mode client is: the user clicks reboot, this starts a POST request to /reboot and sets up the UI to indicate that the reboot has been requested and waits until a response is received, then the client exits.

Hmm, this is exactly the flow we wanted to have in the GUI too.

> Is there some reason the graphical client can't do this?

The GUI never received a response for /reboot (even before we made it exit). I recall this was a known issue. @dbungert was it you working on it?

dbungert:
u-d-i is not currently sending the /reboot request.
I verified this by MITMing the socket - https://superuser.com/questions/484671/can-i-monitor-a-local-unix-domain-socket-like-tcpdump

@jpnurmi - I have delayed the work on changes to the behavior of reboot until the originally committed roadmap items, all of which are u-d-i related, have been addressed.

jpnurmi:
Perhaps we could make the UI not exit in live-run mode.

oSoMoN:
Yes, assuming that it will exit cleanly when being killed by the /reboot request.

mwhudson:
It's certainly hard to be sure what's happening during shutdown, the subiquity server won't return from /reboot until /sbin/reboot has been invoked, so it's all a bit of a race as to whether the response makes it to the socket before everything gets torn down anyway. It also doesn't really make any difference to anything though.

The client not sending the /reboot request does sound like it is exiting too early though.

jpnurmi:
[...]

PR #233, opened 2021-08-16
Close the window after a reboot request

jpnurmi:
Ref: #224

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages.dart
@@ -1,6 +1,6 @@
 export 'pages/allocate_disk_space/allocate_disk_space_page.dart';
 export 'pages/choose_your_look_page.dart';
-export 'pages/installation_complete_page.dart';
+export 'pages/installation_complete/installation_complete_page.dart';
 export 'pages/installation_slides/installation_slides_page.dart';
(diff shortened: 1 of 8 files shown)


mwhudson:
It's safe to call the reboot endpoint in dry run mode, it just causes the server to exit.

mwhudson:
In fact in general the client should not have to care about dry-run at all -- the tui client barely does (it affects how crash reports are generated and enables some dry-run only keybindings).

jpnurmi:
With these changes the request is sent, but `RebootController.POST` gets stuck at [`await self.rebooting_event.wait()`](https://github.com/canonical/subiquity/blob/main/subiquity/server/controllers/reboot.py#L44) and `RebootController.reboot()` never gets called. What could be the problem?

jpnurmi:
Reboot works after installation when the app status is done, but not from the `Turn off RST` or `Turn off BitLocker` screens.

mwhudson:
Oh yeah that makes some kind of sense, the reboot controller waits for the installation to complete before actually restarting. We should add a parameter for this -- "immediate" or something I guess?

robert-ancell (approved):
LGTM

jpnurmi:
> We should add a parameter for this -- "immediate" or something I guess?

Sounds good :+1:

PR #236, opened 2021-08-17
Subiquity client: don't await response for /reboot

jpnurmi:
Split `write()` and `send()` to be able to await `write()` but ignore any response. For `/reboot`, there's no response at all.

Ref: #224


MarcusTomlinson (approved):
Looks good. Thanks!

PR #237, opened 2021-08-17
Edit partition dialog

jpnurmi:
Screenshot from 2021-08-17 14-16-59

Issue: #36

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/l10n/app_en_US.arb
+++ b/packages/ubuntu_desktop_installer/lib/l10n/app_en_US.arb
@@ -72,6 +72,7 @@
   "bootLoaderDevice": "Device for boot loader installation",
 
   "partitionCreateTitle": "Create partition",
+  "partitionEditTitle": "Edit partition",
   "partitionSizeLabel": "Size:",
   "partitionUnitB": "B",
   "partitionUnitKB": "KB",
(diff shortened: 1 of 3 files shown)


jpnurmi:
Depends on #218

jpnurmi:
There's no actual functionality under the hood so one cannot create partitions yet. The dialog is not available with the default partitionless "simple" machine config, but with the "existing partitions" config one can select partitions for editing.

P.S. Sorry for not including any tests in these partitioning-related PRs. I feel like we're not quite ready for extensive testing until we start working on the actual functionality.

oSoMoN:
One tiny comment on the functional side: when clicking outside the `AutoComplete` for the mount point, I would expect it to be dismissed (with the previous value restored if the currently typed one isn't valid). But I can't seem to dismiss it unless I select a value.

That's a detail and the rest looks fine to me, so let's go ahead and merge, it can be addressed later on.

jpnurmi:
Gladly, `Autocomplete` works much better with Flutter 2.5. The popup gets dismissed on clicks outside too.

oSoMoN:
> Gladly, `Autocomplete` works much better with Flutter 2.5. The popup gets dismissed on clicks outside too.

Excellent, thanks for checking!

PR #240, opened 2021-08-18
CI: check outdated generated files

jpnurmi:


oSoMoN:
The CI failure is raising an interesting question: mocks will differ sligthly if they were generated with a different minor version of mockito. How do we deal with that?

jpnurmi:
I believe `pubspec.lock` is supposed to address the problem by pinning dependencies to certain versions, but it never worked too well when jumping between various Flutter versions. Perhaps we should restrict all generator package versions (i.e. remove the hat) in `pubspec.yaml`?

oSoMoN:
I guess that would work, yes.

oSoMoN:
LGTM, thanks!

PR #243, opened 2021-08-19
Fix loading translations on the Welcome page.

### LINKED ISSUE #241: Regression: translations not loaded on Welcome screen
This seems to be a fairly recent regression (although I only noticed it yesterday): when running the installer from a git checkout on a system where the default locale is fr_FR.UTF-8, the language selected by default on the Welcome screen is French (as expected), but UI elements aren't translated in French until I click again on the "French" entry in the list.

screenshot

oSoMoN:
Fixes #241.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/welcome/welcome_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/welcome/welcome_model.dart
@@ -66,11 +66,31 @@ class WelcomeModel extends ChangeNotifier {
 
   /// Selects the given [locale].
   void selectLocale(String locale) {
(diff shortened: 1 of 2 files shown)


oSoMoN:
Before this is ready for review, I'd like to get to the bottom of the unexpected format for `Platform.localeName` I'm seeing in my local tests (see discussion on #241).

Ideally, the improved implementation of `selectLocale()` also needs more complete tests (which would require mocking the list of available translations).

jpnurmi:
Would it be enough to fix the default value of `Settings.locale`? All values applied later are valid locales resolved from the .arb files.

jpnurmi:
Superseded by https://github.com/canonical/ubuntu-desktop-installer/pull/284

PR #247, opened 2021-08-20
Make language names available to WSL setup

jpnurmi:
Move localized language names to the shared `ubuntu_wizard` package and provide a helper function for easy loading of available languages.


oSoMoN:
LGTM, thanks!

PR #248, opened 2021-08-23
WSL variant of SubiquityServer

jpnurmi:
- the Python module is "system_setup.cmd.server"
- should be always started (no systemd)
- no snap or python path environment variables

Ref: canonical/ubuntu-desktop-installer#160, canonical/subiquity#1012


robert-ancell (approved):
LGTM

PR #242, opened 2021-08-19
Adapt to subiquity's new /shutdown endpoint 

jpnurmi:
Solves: #224

PR #249, opened 2021-08-23
Update subiquity for /shutdown changes

jpnurmi:


jpnurmi:
Dry-run installation didn't expose any problems apart from the /shutdown change: https://github.com/canonical/ubuntu-desktop-installer/pull/242

dbungert (approved):
I expect this should be no issue, but if people see subiquity not making installation progress, we'll want a copy of the subiquity-server-debug.log file.

jpnurmi:
Is there something unfinished? Would you prefer to update subiquity later?

dbungert:
No, I don't think so.
There was a rewrite of handling of what controllers are mapped to the client installer variants, so there is a nonzero chance that the install gets blocked waiting on a controller that is not going to be marked as configured.
You should know pretty quick by doing a test install.

jpnurmi:
Ok. A dry-run installation went through fine. Let's merge this so we can give it a spin with the daily image tomorrow.

PR #260, opened 2021-08-25
Shared ubuntu_test package providing common finders and mocks

jpnurmi:
Provide SubiquityClient & GSettings mocks from a shared test package.


jpnurmi:
This is not as intuitive because it's not obvious just by reading the test code, where these shared mocks come from. On the upside, this removes large chunks of duplicate generated code.

I chose specifically these two mocks because:

- `MockSubiquityClient` is used all around the place. Sharing this mock from one place avoids the need of regenerating it in multiple places whenever the `SubiquityClient` API changes.
- `MockGSettings` is needed in both apps' tests. It's a special mock because it's hand-modified due to FFI incompatibility in Mockito.

PR #266, opened 2021-08-26
Share and unify constants

jpnurmi:
- Share the page content width fraction commonly used on pages with
  input form fields.

- Set the common icon spacing as the default for ValidatedFormField
  instead of repeating it all around.


robert-ancell (approved):
LGTM

PR #267, opened 2021-08-26
Share and update analysis options

jpnurmi:
Dart analyzer can find shared `analysis_options.yaml` up in the dir hierarchy. Also, exclude generated `l10n/*.dart` files.


robert-ancell (approved):
LGTM

Issue #155, opened 2021-07-12
WSL Screen 1 - Select your language


jibel:
This is the design for the stage "**Select your language**" of WSL OOBE.
Master issue: #160 

![WSL Screen 1 - Language](https://user-images.githubusercontent.com/1928546/125322820-7415df00-e33e-11eb-9f31-a3ac05406018.png)

Design specification: https://people.canonical.com/~platform/desktop/installer/WSL/#s0

jpnurmi:
Just wanted to check because I was going to re-use the same page as is, but then I noticed that the text is different.

PR #269, opened 2021-08-27
Add WSL Select language page

jpnurmi:
Screenshot from 2021-08-27 13-30-09

Close: #155


oSoMoN:
On my system (default locale is fr_FR.UTF-8), when running the WSL setup wizard, French is listed in the list of languages, but not selected by default, unlike in the desktop installer. Also, manually selecting it doesn't switch the UI to French (I would expect at least the "Go Back" and "Continue" buttons to be translated).

jpnurmi:
> On my system (default locale is fr_FR.UTF-8), when running the WSL setup wizard, French is listed in the list of languages, but not selected by default, unlike in the desktop installer.

Ok, I'll check.

> Also, manually selecting it doesn't switch the UI to French (I would expect at least the "Go Back" and "Continue" buttons to be translated).

There are no translations yet. Only language names are shared between the apps.

oSoMoN:
> There are no translations yet. Only language names are shared between the apps.

We should probably consider sharing common strings, such as "Go Back" and "Continue".

jpnurmi:
> On my system (default locale is fr_FR.UTF-8), when running the WSL setup wizard, French is listed in the list of languages, but not selected by default, unlike in the desktop installer.

The difference was that the Desktop Installer does a language code comparison and the WSL Setup does a full locale comparison. Neither is correct... :point_right: https://github.com/canonical/ubuntu-desktop-installer/pull/289

PR #273, opened 2021-08-31
Add melos gen-l10n for generating localizations in all packages

jpnurmi:
Runs `flutter gen-l10n` in all relevant packages.


robert-ancell (approved):
LGTM

PR #276, opened 2021-08-31
subiquity_client: allow local dependencies from path

jpnurmi:
Must be marked as non-publishable package:

> Publishable packages can't have 'path' dependencies. Try adding a
> 'publish_to: none' entry to mark the package as not for publishing
> or remove the path dependency. - invalid_dependency


robert-ancell (approved):
LGTM

PR #279, opened 2021-09-01
CI: split Flutter build/tests and analysis

jpnurmi:
Let the CI run separate jobs in parallel for faster feedback.

- Before: ~13 minutes
- After: ~7 minutes


oSoMoN:
Good idea!

PR #280, opened 2021-09-01
Split WSL setup and reconfiguration wizards

jpnurmi:
Easier to read and fixes the reconfiguration flow.

Ref: #160

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/wizard.dart
+++ b/packages/ubuntu_wsl_setup/lib/wizard.dart
@@ -5,22 +5,16 @@ import 'pages.dart';
 import 'routes.dart';
 
 class UbuntuWslSetupWizard extends StatelessWidget {
-  const UbuntuWslSetupWizard({
-    Key? key,
-    required this.reconfigure,
-  }) : super(key: key);
-
-  final bool reconfigure;
+  const UbuntuWslSetupWizard({Key? key}) : super(key: key);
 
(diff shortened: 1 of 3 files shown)


robert-ancell (approved):
LGTM

packages/ubuntu_wsl_setup/lib/wizard.dart
```diff
@@ -29,11 +23,26 @@ class UbuntuWslSetupWizard extends StatelessWidget {
...
+class UbuntuWslReconfigureWizard extends StatelessWidget {
+  const UbuntuWslReconfigureWizard({Key? key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Wizard(
+      initialRoute: Routes.advancedSetup,
+      routes: <String, WidgetBuilder>{
+        Routes.advancedSetup: AdvancedSetupPage.create,
+        Routes.configurationUI: ConfigurationUIPage.create,
```
oSoMoN, line 44: Isn't this missing `Routes.setupComplete` in the list of routes? Otherwise, when clicking "Save" on the configuration UI page, what will happen when `Wizard.next()` is called?

jpnurmi: Thanks, I removed the next() call because it was there only to make all pages clickable through in the beginning.

PR #284, opened 2021-09-01
Fix Settings.locale initialization

jpnurmi:
Use full system locale reported by Flutter.

Close: #241


oSoMoN:
LGTM, I verified that it fixes #241 indeed.

I'm seeing an unhandled exception at startup when running with `LANG=oc_FR.UTF-8` (this is a valid locale and we have translations for it):

    LocaleDataException: LocaleDataException: Locale data has not been initialized, call initializeDateFormatting(<locale>)
    
However this is not a regression introduced by this change, it's already behaving like this on main, so I'll file a separate issue and we can address it separately.

PR #286, opened 2021-09-02
WIP: Check for GL support and force software rendering if necessary

kenvandine:
Use glxinfo to attempt to determine if we need to fallback to software rendering.


kenvandine:
> Why is that needed? Shouldn't the graphical stack already fallback in the case where direct rendering isn't available?

It's not though, which I think we should investigate as well.  This is an attempt to work around that.  The installer performs pretty well with software rendering, and this is probably a suitable check.  I want to test this on some specific hardware to ensure it does what I want.

PR #283, opened 2021-09-01
Desktop installer integration test

jpnurmi:
Test cases:
- minimal installation (navigates through the entire wizard, guided partitioning)
- manual partitioning
- turn off bitlocker

PR #287, opened 2021-09-02
Downgrade file package dependency

jpnurmi:
This is necessary for compatibility with `integration_test` from the Flutter stable SDK:

    ERR : Because every version of integration_test from sdk depends on file 6.1.0 and ubuntu_desktop_installer depends on file ^6.1.2, integration_test from sdk is forbidden.
        | So, because ubuntu_desktop_installer depends on integration_test any from sdk, version solving failed.


jpnurmi:
Extracted from https://github.com/canonical/ubuntu-desktop-installer/pull/283

seb128:
thanks

PR #289, opened 2021-09-02
Fix initial selection for languages with multiple countries

jpnurmi:
We can't compare the language code only, because then e.g. `fr_CA` would be always chosen over `fr_FR`. We can't expect a full match either, so find the best possible match (full => lang+country => lang).

Close: #241

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/welcome/welcome_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/welcome/welcome_model.dart
@@ -55,13 +55,11 @@ class WelcomeModel extends ChangeNotifier {
   /// Returns the name of the language at the given [index].
   String language(int index) => _languageList[index].name;
 
-  /// Selects the given [locale].
-  void selectLocale(String locale) {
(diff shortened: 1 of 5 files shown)


oSoMoN:
Nicely architected, I like the extension on `List<LocalizedLanguage>`.
The initial language selection works well, however translations are not loaded in the case of a partial match, i.e. if I run the application with `LANG=pt_PT.UTF-8`, the "Brazilian Portuguese" entry is selected by default, but the UI is in English.

jpnurmi:
> The initial language selection works well, however translations are not loaded in the case of a partial match, i.e. if I run the application with LANG=pt_PT.UTF-8, the "Brazilian Portuguese" entry is selected by default, but the UI is in English.

Oh right, if `pt_PT` is not available, we cannot substitute it with a random other `pt_XX`. :facepalm: The only allowed substitute is the parent locale `pt`. The wrong language selection in the UI was misleading. There were no translations because `pt.arb` is empty.

oSoMoN:
LGTM, thanks!

PR #291, opened 2021-09-02
Use the logging framework instead of ad-hoc print() statements

oSoMoN:
and use it to replace all calls to print().

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/installer.dart
+++ b/packages/ubuntu_desktop_installer/lib/installer.dart
@@ -1,8 +1,4 @@
-import 'dart:io' as io;
-
 import 'package:flutter/material.dart';
-import 'package:logger/logger.dart';
-import 'package:path/path.dart' as p;
 import 'package:provider/provider.dart';
 import 'package:subiquity_client/subiquity_client.dart';
 import 'package:ubuntu_wizard/settings.dart';
@@ -37,11 +33,6 @@ class UbuntuDesktopInstallerApp extends StatelessWidget {
       home: _UbuntuDesktopInstallerWizard.create(context),
     );
   }
-
-  static final _appName = p.basename(io.Platform.resolvedExecutable);
-
-  /// Static reference to an application-level common [Logger].
-  static final log = Logger(_appName);
 }
 
 class _UbuntuDesktopInstallerWizard extends StatefulWidget {
--- a/packages/ubuntu_desktop_installer/lib/installer.dart
+++ b/packages/ubuntu_desktop_installer/lib/installer.dart
@@ -1,4 +1,8 @@
+import 'dart:io' as io;
+
 import 'package:flutter/material.dart';
+import 'package:logger/logger.dart';
+import 'package:path/path.dart' as p;
 import 'package:provider/provider.dart';
 import 'package:subiquity_client/subiquity_client.dart';
 import 'package:ubuntu_wizard/settings.dart';
@@ -33,6 +37,11 @@ class UbuntuDesktopInstallerApp extends StatelessWidget {
       home: _UbuntuDesktopInstallerWizard.create(context),
     );
   }
+
+  static final _appName = p.basename(io.Platform.resolvedExecutable);
+
+  /// Static reference to an application-level common [Logger].
+  static final log = Logger(_appName);
 }
 
 class _UbuntuDesktopInstallerWizard extends StatefulWidget {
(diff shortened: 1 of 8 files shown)


packages/ubuntu_desktop_installer/lib/installer.dart
```diff
@@ -37,6 +41,11 @@ class UbuntuDesktopInstallerApp extends StatelessWidget {
       home: _UbuntuDesktopInstallerWizard.create(context),
     );
   }
+
+  static final _appName = p.basename(io.Platform.resolvedExecutable);
+
+  /// Static reference to an application-level common [Logger].
+  static final log = Logger(_appName);
```
jpnurmi, line 48: what if we had this as a global in `logger.dart`? then we could use `log.foo()` everywhere and it would default to the application-wide logger, but subiquity_client and others could still use a local logger instance.

oSoMoN: Good suggestion. I named the common application logger `appLogger`, because I feel `log` is a bit too generic, but I don't have a strong opinion on this, so let me know if you'd rather rename it.

jpnurmi: I'd vote for `log` because then you wouldn't need to stop to think and look around, but you'd automatically get the nearest logger in the scope. :slightly_smiling_face: 

I'm not against a general app-wide logger as a fallback, but I had somehow envisioned that we'd use smaller scoped loggers. The `<app-name>` context is redundant because the log-file already tells that. Smaller scopes would make it clear where the output comes without needing to specify that in the output itself.

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
```diff
@@ -47,7 +48,7 @@ class InstallationSlidesModel extends ChangeNotifier with SystemShutdown {
 
   void _updateStatus(ApplicationStatus? status) {
     if (state == status?.state) return;
-    print(
+    appLogger.debug(
         'Subiquity state: ${_formatState(state)} => ${_formatState(status?.state)}');
```
jpnurmi, line 52: now that i think of this, we should probably output subiquity states `subiquity_client.dart` instead :)

packages/ubuntu_desktop_installer/lib/installer.dart
jpnurmi: > I had somehow envisioned that we'd use smaller scoped loggers.

For example, everything in `pages/keyboard_layout/*.dart` would log to `Logger('keyboard_layout')`.

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
jpnurmi: This is what we have in the current main:
```
flutter: Subiquity state: UNKNOWN => RUNNING
flutter: DEBUG subiquity_client: GET http://localhost/meta/status?cur=%22RUNNING%22
flutter: Subiquity state: RUNNING => POST_RUNNING
flutter: DEBUG subiquity_client: GET http://localhost/meta/status?cur=%22POST_RUNNING%22
flutter: Subiquity state: POST_RUNNING => UU_RUNNING
flutter: DEBUG subiquity_client: GET http://localhost/meta/status?cur=%22UU_RUNNING%22
flutter: Subiquity state: UU_RUNNING => DONE
```

What if we printed all subiquity responses? (Shortened to max XX chars?) Then we could throw print this out.

packages/ubuntu_desktop_installer/lib/installer.dart
oSoMoN: Ah, that makes sense, I'll update the PR to implement it.

jpnurmi: Regarding the automatic scoping discussion on Friday. Perhaps we don't actually need anything fancy like that, because if we manage to keep views dumb and declarative, all(?) logging would be in models and services.

oSoMoN: Agreed. I went back on my initial proposal, and used per-file `Logger` instances to log relevant information, only in models and services.

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
oSoMoN: I moved the logging of state changes to `subiquity_client`. We can improve/systematize logging of subiquity responses in a follow-up PR, if that works for you.

jpnurmi (approved):
Thanks, LGTM :+1:

PR #296, opened 2021-09-03
Improve input validation

jpnurmi:
- Allow using `EqualValidator` for equal empty input.
- Expose the validation mode for `ValidatedFormField`.
- Validate confirmation fields immediately when the respective password
  field is being filled up to avoid leaving the confirmation field in
  an unclear seemingly valid empty/unvalidated state.

This improves password input validation on the existing "Who are you?"
and "WSL profile setup" pages, but is even more important on the
upcoming "Choose a security key" where there are only two input fields.


robert-ancell (approved):
LGTM

PR #297, opened 2021-09-06
Remove LocalizedView

jpnurmi:
`LocalizedView` cannot be shared in `ubuntu_wizard:widgets`, because `AppLocalizations` is an app-specific type, and Dart does not allow accessing static members of generic types.

Access `AppLocalizations` directly using the same syntax view models, `Settings`, `Wizard`, and many built-in Flutter types such as `Theme` and `MediaQuery` are accessed.

`AppLocalizations` is available via `l10n.dart` and the getter is non-nullable since #246, which makes ot convenient enough to use.

As a bonus, this reduces the widget tree depth by one level in a majority of pages and widgets.


jpnurmi:
A brutal diff for such a simple change :(

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete_page.dart
```diff
@@ -23,63 +23,61 @@ class InstallationCompletePage extends StatelessWidget {
...
+              crossAxisAlignment: CrossAxisAlignment.stretch,
+              children: [
+                Padding(
+                  padding: const EdgeInsets.symmetric(vertical: 24),
+                  child: ElevatedButton(
                     onPressed: () {
                       Provider.of<InstallationCompleteModel>(context,
                               listen: false)
-                          .shutdown(immediate: false);
+                          .reboot(immediate: false);
```
oSoMoN, line 62: Nice catch!
It would have been better in a separate commit, but I won't be the one to pick nits over this.

oSoMoN (approved):
LGTM, thanks!
It's a shame GH's side by side diff viewer isn't able to do a better job at hiding the noise in this diff to help visualize the really meaningful changes…

oSoMoN:
Feel free to go ahead and merge once you've resolved the two conflicts.

packages/ubuntu_desktop_installer/lib/pages/installation_complete/installation_complete_page.dart
jpnurmi: The diff is really bad and shows these actions shifted. I don't admit hiding any other changes here. :)

Issue #34, opened 2021-03-08
Screen 9 - Choose a Security Key


seb128:
The design for screen n°9, Choose a Security Key

![key](https://user-images.githubusercontent.com/10974330/110314059-a7023c00-8007-11eb-8615-a8bd332af302.png)

Design details https://people.canonical.com/~platform/desktop/installer/9.%20Choose%20a%20Security%20Key/

lyrebird-sound-cloning-system:
"A recovery key is generated and will be temporarily saved on the live system. You can select an alternative location. Save this file and keep it in a safe place elesewhere before rebooting." 

I think that the installer should explain to the end-user what's the purpose of the recovery key.

jpnurmi:
The recovery key part is still missing but there's a separate dedicated ticket for it:
- https://github.com/canonical/ubuntu-desktop-installer/issues/1350

PR #298, opened 2021-09-07
Add Choose security key page

jpnurmi:
Screenshot from 2021-09-07 09-08-13

Ref: #34


jpnurmi:
#34 is in fact two pages: 1) choose security key and 2) generate recovery key. The second one would follow in a separate PR later.

jpnurmi:
Depends on #297.

jpnurmi:
Outdated. Needs to be rebased.

Issue #294, opened 2021-09-03
Cannot write logs - permission denied


jpnurmi:
daily -canary 20210902:
```
$ ubuntu-desktop-installer
[ERROR:flutter/lib/ui/ui_dart_state.cc(199)] Unhandled Exception: FileSystemException: Cannot create link, path = '/var/log/installer/ubuntu_desktop_installer.log' (OS Error: Permission denied, errno = 13)
#0      _Link.throwIfError (dart:io/link.dart:263)
#1      _Link.createSync (dart:io/link.dart:184)
#2      _createSymlink (package:logger/logger.dart:145)
#3      Logger.setup (package:logger/logger.dart:65)
#4      runWizardApp (package:ubuntu_wizard/app.dart:37)
#5      main (package:ubuntu_desktop_installer/main.dart:20)
#6      _runMainZoned.<anonymous closure>.<anonymous closure> (dart:ui/hooks.dart:140)
#7      _rootRun (dart:async/zone.dart:1354)
#8      _CustomZone.run (dart:async/zone.dart:1258)
#9      _runZoned (dart:async/zone.dart:1789)
#10     runZonedGuarded (dart:async/zone.dart:1777)
#11     _runMainZoned.<anonymous closure> (dart:ui/hooks.dart:138)
#12     _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:283)
#13     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:184)
```

jpnurmi:
Fix verified on daily-canary 20210927.1.

```bash
ubuntu@ubuntu:~$ ls -la /var/log/installer/ubuntu_desktop_installer.log
lrwxrwxrwx 1 ubuntu ubuntu 52 Sep 28 08:15 /var/log/installer/ubuntu_desktop_installer.log -> /var/log/installer/ubuntu_desktop_installer.log.2936
ubuntu@ubuntu:~$ cat /var/log/installer/ubuntu_desktop_installer.log
2021-09-28 08:15:58.715880 INFO subiquity_client: Opening socket /run/subiquity/socket
```

jpnurmi:
daily-canary 20220306:
```
[...]

PR #299, opened 2021-09-09
Handle file logging errors

jpnurmi:
Makes sure the installer runs even if it has no permissions to write logs to `/var/log/installer` (#294).


jpnurmi:
This would be nice to have in so the installer will run again, regardless of /var/log/installer permissions.

oSoMoN (approved):
LGTM, thanks!

PR #300, opened 2021-09-10
subiquity: update, disable autoinstall

dbungert:
Running ubuntu-desktop-installer with a subiquity autoinstall file does
work, but the UI has no knowledge of this interaction.  Disable it for
today until UI integration can be completed.

Update the version of subiquity.  Changes include
* support on the autoinstall argument needed to disable autoinstall
* WSL functionality
* pubsub refactor
* updated curtin


jpnurmi:
Thanks! :+1:

PR #301, opened 2021-09-13
Share common wizard actions

jpnurmi:
Suggested in https://github.com/canonical/ubuntu-desktop-installer/pull/269#issuecomment-911596591

### DIFF
--- a/packages/ubuntu_desktop_installer/test/write_changes_to_disk_page_test.dart
+++ b/packages/ubuntu_desktop_installer/test/write_changes_to_disk_page_test.dart
@@ -83,10 +83,7 @@ void main() {
       WidgetTester tester, List<Map<String, dynamic>> storageConfig) async {
     app = MaterialApp(
       supportedLocales: AppLocalizations.supportedLocales,
-      localizationsDelegates: [
-        ...AppLocalizations.localizationsDelegates,
-        ...UbuntuLocalizations.localizationsDelegates,
-      ],
+      localizationsDelegates: localizationsDelegates,
       locale: Locale('en'),
       home: Wizard(
         routes: {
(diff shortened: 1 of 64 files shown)


jpnurmi:
This proposal abstracts `UbuntuLocalizations` away from UDI and WSL, so they don't have to access both, app-specific `AppLocalizations` and shared `UbuntuLocalizations`.

packages/ubuntu_wizard/lib/src/widgets/wizard_action.dart
```diff
@@ -0,0 +1,86 @@
...
+          onActivated();
+        } else {
+          Wizard.of(context).back();
+        }
+      },
+    );
+  }
+
+  /// A _Continue_ action that advances to the next page.
+  factory WizardAction.next(
```
jpnurmi, line 43: Named `next` because `continue` is a reserved keyword

packages/ubuntu_desktop_installer/test/write_changes_to_disk_page_test.dart
```diff
@@ -83,7 +83,10 @@ void main() {
       WidgetTester tester, List<Map<String, dynamic>> storageConfig) async {
     app = MaterialApp(
       supportedLocales: AppLocalizations.supportedLocales,
-      localizationsDelegates: AppLocalizations.localizationsDelegates,
+      localizationsDelegates: [
+        ...AppLocalizations.localizationsDelegates,
+        ...UbuntuLocalizations.localizationsDelegates,
```
oSoMoN, line 88: Just thinking out loud here without having checked the feasibility of my suggestion: is there a way to include `UbuntuLocalizations.localizationsDelegates` in `AppLocalizations.localizationsDelegates` (both here and in the `ubuntu_wsl_setup` package), so that there's no need to copy/paste this snippet all over the place?
Or maybe declare a global static `localizationsDelegates` that can be referenced everywhere?

oSoMoN (commented):
Thanks for implementing my suggestion! See one question/suggestion inline.

packages/ubuntu_desktop_installer/test/write_changes_to_disk_page_test.dart
jpnurmi: Great idea! :+1:

I can't think of a way to inject extra items into the auto-generated list and extension methods cannot be static, so I went with the second option.

oSoMoN: LGTM, thanks!

PR #302, opened 2021-09-13
inject-snap: add test script

dbungert:
This inject-snap script can be used to modify an existing iso to inject
a snap into it.  This process is used often in subiquity development and
can be a convenient method of testing changes.


jpnurmi (approved):
Very nice, thanks!

PR #304, opened 2021-09-13
subiquity: add a basic integration test

dbungert:
The version of curtin and probert expected by a given version of
subiquity should match the actual version in use for
ubuntu-desktop-installer.


dbungert:
Previous check run correctly showing this test failing: https://github.com/canonical/ubuntu-desktop-installer/runs/3590339107?check_suite_focus=true

jpnurmi:
Could we place the script in `.github/scripts`?

jpnurmi (approved):
Thanks, LGTM :+1:

seb128:
How is that working? Does it mean that when we update subiquity we need to also think about updating our snapcraft to use a matching curtin version? If so maybe it would make to add a script 'update-subiquity' which does that for us? (the check is still useful to have but the less manual work the better right?)

jpnurmi:
Would it be ok to have a bot that creates a PR to auto-roll the submodule nightly or weekly?

jpnurmi:
Something similar to what Flutter does for the engine and plugins:
https://github.com/flutter/flutter/pulls?q=is%3Apr+is%3Aclosed+roll

seb128:
I guess it depends of our strategy, do we want to follow subiquity closely and update often, in which case the auto PR makes sense,  or do we want to avoid changes when not needed and only pull an update when it has been a while or if there is a reason to do so?

jpnurmi:
Yeah. It could help to catch issues early, at least. Oh, and the PRs would not be auto-merged, of course.

On a related note, I have also prepared an integration test that goes through the entire installation wizard for an extra safety net: https://github.com/canonical/ubuntu-desktop-installer/pull/283

dbungert:
I wanted to explicitly say that the version of curtin that subiquity expects was the version that we were using by default for u-d-i.

Yes, a script to automate updating the snapcraft.yaml for u-d-i sounds like a good addition, I can put that together.
The test is still valuable because someone might manually adjust the snapcraft.yaml.

dbungert:
#312

PR #305, opened 2021-09-15
CI: add check for outdated generated localizations files

jpnurmi:


jpnurmi:
I was hesitant to add this before because Flutter stable and dev/master produced slightly differently formatted files. This is no longer the case after the latest stable 2.5 release.

seb128:
Thanks for the work, one question, why did we have a git clean and why isn't that needed anymore?

jpnurmi:
Sorry for trying to slip in changes that are on the borderline of being unrelated. :) Cleaning the repo was a workaround for the problem that got fixed by https://github.com/canonical/ubuntu-desktop-installer/pull/274

seb128 (approved):
Alright, ideally those would be different commits in the same PR at least so it's easier to understand the change when reading history but it's minor enough that we should be able to deal with having the explanation here if someone really wants to figure out the why :-)

jpnurmi:
True! I've split the commits, to leave a clear track.

seb128:
Thanks! Feel free to merge, I'm unsure why the codecov/project check is read but it shows changes in dart files so it doesn't much sense there?

jpnurmi:
Looks a lot like something that was merged earlier today. Rebasing would help but I'm not going to bother in this case. :)

PR #306, opened 2021-09-15
Fix alignment of checkbuttons on the Who are you? page

jpnurmi:
Spacing slipped between wrong UI elements in the column.


jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2021-09-15 09-28-21 | Screenshot from 2021-09-15 09-29-19 |

jpnurmi:
Spec for reference:

oSoMoN:
Good catch, thanks!

PR #307, opened 2021-09-15
Use check- and radio buttons in Allocate disk space dialogs

jpnurmi:
To get rid of the highlight background which is meant for list views
and looks a bit awkward on a frameless background. Our custom desktop
oriented check and radio buttons were added for exactly this reason.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
@@ -64,17 +64,32 @@ Future<void> showCreatePartitionDialog(BuildContext context, DiskModel disk) {
               child: Column(
                 crossAxisAlignment: CrossAxisAlignment.end,
                 children: <Widget>[
-                  _PartitionDialogLabel(lang.partitionSizeLabel),
+                  _PartitionDialogLabel(
+                    lang.partitionSizeLabel,
+                    height: tileHeight,
+                  ),
                   const SizedBox(height: kContentSpacing),
-                  _PartitionDialogLabel(lang.partitionTypeLabel),
-                  ListTile(),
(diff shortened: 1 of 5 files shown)


jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2021-09-15 09-37-44 | Screenshot from 2021-09-15 09-37-32 |

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
```diff
@@ -191,7 +191,7 @@ class _RadioValueTile<T> extends StatelessWidget {
     return ValueListenableBuilder<T>(
       valueListenable: groupValue,
       builder: (context, groupValue, child) {
-        return RadioListTile<T>(
+        return RadioButton<T>(
```
oSoMoN, line 194: Vertical alignment of the two columns is broken, as evidenced by your "after" screenshot (and tested locally).

oSoMoN (changes requested):

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
jpnurmi: Good catch, thanks!

jpnurmi:
I rebased and added a separate commit to "Share and expose kRadioSize constant" in ubuntu_wizard. I hope it's ok to include it here as a separate commit, if you could rebase or merge to keep it so, to avoid creating a dependency PR.

jpnurmi:
Kuvakaappaus - 2021-09-21 10-28-56
Kuvakaappaus - 2021-09-21 10-28-44

oSoMoN:
LGTM now, thanks!

PR #309, opened 2021-09-15
Allow disabled state for CheckButton and RadioButton

jpnurmi:
To allow graying out irrelevant selections.


jpnurmi:
Screenshot from 2021-09-15 14-20-16

oSoMoN:
LGTM, thanks!

Issue #33, opened 2021-03-08
Screen 8 - Installation Type


seb128:
The design for screen n°8, Installation Type

![installation-type](https://user-images.githubusercontent.com/10974330/110313929-71f5e980-8007-11eb-97b0-15b4e4e8aafb.png)

Design details https://people.canonical.com/~platform/desktop/installer/8.%20Installation%20Type/

BillDietrich:
Could we have a preview of "Advance Features", please ?

I would hope that under there would be simple radio-buttons to choose among:

- a single / partition, or separate / and /home partitions, or systemd-homed

- no swap, swap partition, swap file, zram, zswap

- filesystem type: zfs, Btrfs, ext4, etc

And maybe simple check-boxes to turn on/off:

- LUKS / LVM for whole Linux system

- eCryptfs for home dir

- hardware key versus password for LUKS encryption

- hardware key versus password for login

In particular, I think decisions about filesystem type and swap and number of partitions should NOT be left until the "Allocate Disk Space" page.  There should be simple policy-type choices made in this "Installation Type" page.

Thanks.

long-chung:
@BillDietrich Click the "Design details link" below the thumbnail. There are other states as well as specifications in that link

ArishSultan:
@oSoMoN I would like to work on this screen.

oSoMoN:
All yours ;)
Looking forward to reviewing your code.

ArishSultan:
I have completed the UI. but the `yaru.dart` theme is not fully implemented. For Example, the OutlinedButton theme is not the same as in the provided design.

Last week I created a PR at `yaru.dart` to fix the TextStyles. what should I do, should I wait for the next update of `yaru.dart` or just continue building.

If we can wait for `yaru.dart` to complete then I might work on other screens that are unassigned yet.

BillDietrich:
Under "Advanced Features", I see none of the options I requested, and no Issue for that screen.

oSoMoN:
@ArishSultan as this is early stages for this project, we're taking an iterative approach, and if UI elements don't look exactly like the specification, it won't be considered a blocker. So I'd suggest going ahead with a PR if the bulk of the screen is implemented.

ArishSultan:
Ok

jpnurmi:
[...]

PR #310, opened 2021-09-15
Add Installation type page

jpnurmi:
NOTE: There is no communication with subiquity yet. Any selection takes currently to the Allocate disk space page.

Ref: #33


jpnurmi:
Screenshot from 2021-09-15 16-16-21
Screenshot from 2021-09-15 16-16-27
Screenshot from 2021-09-15 16-17-03

packages/ubuntu_desktop_installer/test/installation_type_model_test.dart
```diff
@@ -0,0 +1,43 @@
...
+
+    wasNotified = false;
+    expect(model.encryption, isFalse);
+    model.encryption = true;
+    expect(wasNotified, isTrue);
+  });
+
+  test('product info', () {
+    final model = InstallationTypeModel(MockSubiquityClient());
+    expect(model.productInfo, isNot(equals('')));
```
oSoMoN, line 41: Isn't this equivalent to `isNotEmpty` ?

packages/ubuntu_desktop_installer/test/installation_type_page_test.dart
```diff
@@ -0,0 +1,141 @@
...
+      home: Wizard(
+        routes: {'/': (_) => buildPage(model)},
+        onNext: (settings) => '/',
+      ),
+    );
+  }
+
+  testWidgets('reinstall', (tester) async {
+    final model = buildModel(
+      installationType: InstallationType.reinstall,
```
oSoMoN, line 56: I'm not sure I understand why the model's `installationType` is initialized to `InstallationType.reinstall`, if we're checking the behaviour of selecting the option from the UI (same question for the next 3 tests, `alongside`, `erase` and `manual`)?

oSoMoN (commented):
LGTM, just a couple of minor questions inline.

packages/ubuntu_desktop_installer/test/installation_type_model_test.dart
jpnurmi: /me hides

packages/ubuntu_desktop_installer/test/installation_type_page_test.dart
jpnurmi: i removed these lines because the initial type doesn't actually matter. this is testing that the correct installation type is passed to the model when the respective radio button is clicked

jpnurmi:
> Thanks for the work there, one thing is that we should probably disable the options which are not going to work this cycle, either by hidding or disabling the controls.

At the moment, only the "Erase disk and install Ubuntu" option for guided partitioning and the "Something else" option for manual partitioning can be visible to users.

There's no actual query for existing OS installations in place. We discussed adding API for this to subiquity later. For the screenshots, I modified the code to pretend that there was an existing OS installation to show that the UI has been prepared.

madsrh:
@jpnurmi the warning label (red text on dark background) is very hard to read.

Should I create a separate issue and tag someone from design or do you have a fix in mind that works with both the light and dark theme?

jpnurmi:
@madsrh Yes please, that would be great! It's the standard `ThemeData.errorColor`. I recall @Feichtmeier wanted to make it lighter.

Feichtmeier:
Sorry! Forgotten about this one

How is this (10% lighter)

grafik
FLAMESHOT_2021-09-21_09-00

Feichtmeier:
@jpnurmi please include yaru.dart 0.1.4 for a fix :)

jpnurmi:
@Feichtmeier Great, thanks!

jpnurmi:
https://github.com/canonical/ubuntu-desktop-installer/pull/325

PR #311, opened 2021-09-15
Update of the Occitan locale

Mejans:
Based on the EN file on 15/09/2021
I wonder what to do with the placeholder things, do we have to translate what's next to them?


oSoMoN:
> Based on the EN file on 15/09/2021
> I wonder what to do with the placeholder things, do we have to translate what's next to them?

Thanks for the update @Mejans !
Placeholders don't need to be translated, they are mainly for translators like you to get some additional context or information on the string being translated. Translating them doesn't hurt of course, but it's not necessary.

PR #312, opened 2021-09-15
update-subiquity script

dbungert:
* Add an update-subiquity script which updates the submodule and the snapcraft
* Yaml file manipulation managed by ruamel, which can mostly roundtrip the file
  but won't preserve inconsistent indentation (arguably a good thing) and has
  some opinions on the strings.
* probert build-packages list change is intentional, to better match the rest
  of the file

### DIFF
--- a/scripts/update-subiquity
+++ b/scripts/update-subiquity
@@ -1,42 +1,27 @@
 #!/usr/bin/env python3
 
 import subprocess
-import yaml
-
-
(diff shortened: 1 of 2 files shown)


jpnurmi:
@seb128 Is this ok?

seb128 (approved):
thanks, the code makes sense and a local try shows it works as expected

PR #313, opened 2021-09-16
Allow resizing the window in debug mode for testing purposes

jpnurmi:
A fixed window size makes it tempting to hard-code geometries for UI
elements. Putting in the little extra effor to keep things scalable pays
off if and when the design changes or if we ever decide to change the
default window size in the future.

This change keeps the window size fixed in production/release mode but
makes it resizable in debug mode to motivate scalable layouts.


oSoMoN:
LGTM, thanks!

PR #314, opened 2021-09-16
UDI: switch to manual gen-l10n

jpnurmi:
Automatic `app_localizations*.dart` generation is frustrating to work
with, because either the generation is unreliable or IDEs have problems
resolving the files generated into `.dart_tool/flutter_gen/gen_l10n/`.
As a result, half of the IDE is very often red.

After this change, `flutter gen-l10n` (or `melos run gen-l10n`) needs
to be run manually when changing `.arb` files. Given that translations
do not change nearly as often as the source code, this makes the overall
development experience much more pleasant regardless of an extra manual
step.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,17 +24,30 @@ Bugs are also tracked as [GitHub issues](https://github.com/canonical/ubuntu-des
 
 ## Translations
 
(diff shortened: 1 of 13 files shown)


jpnurmi:
With auto-generated l10n-files it often looks like this. Notice that the generated files are there. Usually, it takes a few rounds of `flutter pub get` to recover.

Screenshot from 2021-09-16 11-16-28

jpnurmi:
CONTRIBUTING.md needs an update

oSoMoN:
LGTM, thanks!

PR #315, opened 2021-09-16
CI: regenerate files automatically

jpnurmi:


jpnurmi:
- To check any potential issues at PR phase (e.g. missing commas in .arb files :smile:), `melos run generate` and `melos run gen-l10n` are kept in the analysis step that is run for PRs, but we no longer care if there are changes.

- The bot cannot commit to a PR, so the automatic code generation step runs only after new changes are pushed to the mainline. The bot creates a new commit if there are any changes after `melos run generate` or `melos run gen-l10n`.

jpnurmi:
The commit would look like this:
```
Author: Ubuntu Desktop Installer <ubuntu-desktop-installer@users.noreply.github.com>
Date:   ...

    Regenerate files
```

Better name or commit message, anyone?

oSoMoN:
>     but we no longer care if there are changes

Would it be possible for the bot to add a comment to the PR to inform that generated files are outdated?

jpnurmi:
It gets a bit convoluted and I wonder if it might get annoying too (like the Codecov comments we had earlier)... Perhaps a status check would be better?

jpnurmi:
I learned that ideally `generate: true` and `synthetic-package: false` should work together, but it's a bug in the flutter tool that they don't. If that gets fixed at some point, we can re-enable `generate: true`, and then it will be easy to notice outdated l10n files locally too.

jpnurmi:
@oSoMoN Let me know if you're happy with a failing status check as shown above, and its output on the details page. If so, I'll clean up the commits.

oSoMoN:
> @oSoMoN Let me know if you're happy with a failing status check as shown above, and its output on the details page. If so, I'll clean up the commits.

Yes, the failing status check looks good, let's go ahead with that.

jpnurmi:
I removed the custom author for auto-generation commits because GitHub has a special visualization for standard bot commits:

Screenshot from 2021-09-20 11-21-52

oSoMoN:
LGTM, thanks!

PR #322, opened 2021-09-20
ubuntu_wizard: add Finnish translation

jpnurmi:


oSoMoN:
I can't speak Finnish/Suomi, but the changes look fine. Now we need translations for the desktop installer / WSL setup apps, so that the language shows up on the welcome page!

PR #325, opened 2021-09-21
Isolate and update yaru.dart

jpnurmi:
I took the opportunity to clean up the dependencies a bit to isolate yaru.dart usage to one place so we don't have to update it in several places...


jpnurmi:
Technically, we shouldn't have to bump up 0.x.y patch versions, because they are considered compatible and fetched automatically, but it doesn't hurt to ensure that everyone has an up-to-date dependency and gets the correct looks.

oSoMoN:
LGTM, thanks!

PR #327, opened 2021-09-21
SubiquityClient.getKeyboardStep(): use logged _send()

jpnurmi:
Logging and keyboard layout detection were added at the same time so
logging changes didn't end up to the latter.


oSoMoN:
Good catch, thanks!

Issue #319, opened 2021-09-17
The installer often fails to connect on first run right after startup


jpnurmi:
Steps (tested with daily-canary 20210916):
- run the latest daily canary
- start the installer as soon as the desktop shows up
- navigate to the keyboard layout page => the list is empty
- restart => the list is populated

```
flutter: ERROR ubuntu_desktop_installer: Unhandled exception
	SocketException: SocketException: Connection failed (OS Error: No such file or directory, errno = 2), address = /run/subiquity/socket, port = 0
```

jpnurmi:
In dry-run mode, the UI waits up to 10s for subiquity to start responding. Is something similar needed in live-run mode too?

jpnurmi:
Reopening because the issue probably still exists as reported by @guiverc [on Discourse](https://discourse.ubuntu.com/t/new-desktop-installer-preview-build/24765/16?u=jpnurmi). With daily canary 2021-11-01: 
> I cannot select a keyboard layout. I have none to choose from? I can use the “Detect Keyboard Layout”; but other than that my go back.

dbungert:
I ran into the same thing with automated API testing, and when I run several tests in parallel I can hit the longer end of the poll time.  Here's what I did to address it for API test:
```
    async def poll_startup(self):                                                                                              
        for _ in range(20):                                                                                                    
            try:                                                                                                               
                await self.get('/meta/status')                                                                                 
                return                                                                                                         
            except aiohttp.client_exceptions.ClientConnectorError:                                                             
                await asyncio.sleep(.5)                                                                                        
        raise Exception('timeout on server startup')                                                                           
```

jpnurmi:
When the issue was last reported, the UI already had a similar loop but 10 x 1s instead.

It would be great to show a warning or something when it takes suspiciously long to connect, but we could let the client keep trying under the hood. The UI is unusable anyway regardless of whether it keeps trying or gives up after an arbitrary amount of attempts.

jpnurmi:
The GUI waits up to 30s these days.

PR #328, opened 2021-09-21
Wait for subiquity in live-run mode

jpnurmi:
The wait is non-blocking (`sleep` changed to `Future.delayed`) to make it possible to visualize the wait in the UI.

Ref: #319

### DIFF
--- a/packages/subiquity_client/lib/subiquity_server.dart
+++ b/packages/subiquity_client/lib/subiquity_server.dart
@@ -60,7 +60,7 @@ abstract class SubiquityServer {
         if (serverMode == ServerMode.DRY_RUN) '--dry-run',
         ...?args,
       ];
-      _startSubiquity(subiquityCmd);
+      await _startSubiquity(subiquityCmd);
     }
 
     return _waitSubiquity(socketPath).then((_) {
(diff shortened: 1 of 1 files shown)


jpnurmi:
> to make it possible to visualize the wait in the UI.

For example, to let the window content show up in a disabled state instead of an empty window that feels broken.

packages/subiquity_client/lib/subiquity_server.dart
```diff
@@ -53,17 +53,23 @@ abstract class SubiquityServer {
...
-    if (!_shouldStart(serverMode)) {
-      return socketPath;
+    if (_shouldStart(serverMode)) {
+      var subiquityCmd = <String>[
+        '-m',
+        _pythonModule,
+        if (serverMode == ServerMode.DRY_RUN) '--dry-run',
+        ...?args,
+      ];
+      _startSubiquity(subiquityCmd);
```
oSoMoN, line 63: Shouldn't we be `await`ing the result of `_startUbiquity(…)` here? Otherwise `_waitSubiquity(…)` risks starting to poll the server too early, no?

oSoMoN (commented):
LGTM, just one question inline.

packages/subiquity_client/lib/subiquity_server.dart
jpnurmi: Yes, it's definitely the right thing to do. Thanks!

oSoMoN:
LGTM, thanks!

PR #329, opened 2021-09-21
Add SubiquityClient.isOpen

jpnurmi:
To let the UI know when subiquity is ready in a way that the UI is able
to visualize the waiting state without blocking the whole UI from
starting up.

Ref: #319

### DIFF
--- a/packages/subiquity_client/lib/subiquity_client.dart
+++ b/packages/subiquity_client/lib/subiquity_client.dart
@@ -1,5 +1,6 @@
 library subiquity_client;
 
+import 'dart:async';
 import 'dart:convert';
 import 'package:http/http.dart';
 import 'package:logger/logger.dart';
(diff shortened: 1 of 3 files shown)


packages/subiquity_client/lib/subiquity_client.dart
```diff
@@ -23,9 +24,13 @@ String _formatResponseLog(String method, String response) {
 
 class SubiquityClient {
   late HttpUnixClient _client;
+  final _completer = Completer<bool>();
+
+  Future<bool> get isOpen => _completer.future;
 
   void open(String socketPath) {
     log.info('Opening socket $socketPath');
+    _completer.complete(true);
```
robert-ancell, line 33: Does it make any difference if this is done as the first thing or the last thing in this function? Wondering if there's a case where something could react to this before `_client` is set (and any potential future code that is later added).

robert-ancell (approved):
One question about when the completer runs, but otherwise LGTM.

packages/subiquity_client/lib/subiquity_client.dart
jpnurmi: true, completing it after opening the client is the correct order, even if it didn't seem to make a difference to the UI, probably because it's asynchronous

PR #330, opened 2021-09-21
Write changes to disk: parse storage config only once

jpnurmi:
Previously, the list of disks and partitions was parsed and collected to an internal list every time `build()` was called. This caused the respective UI elements to accumulate when navigating back and forth, for example. Moving the parsing code to `initState()` ensures that the list is initialized once.


jpnurmi:
Aha, I was navigating back and forth between different pages. This fixes the issue that navigating _forward_ and coming back to _Write changes to disk_ accumulates the items.

I think the other case needs a reset or to restore the original config in subiquity because it's sending previously selected disks as "preserve: false".

oSoMoN:
LGTM, thanks!

PR #331, opened 2021-09-21
Expose the subiquity API to set the variant ('server'/'desktop'), and use it.

oSoMoN:


packages/subiquity_client/lib/src/types.dart
```diff
@@ -6,6 +6,8 @@ part 'types.g.dart';
 
 // ignore_for_file: invalid_annotation_target
 
+enum Variant { SERVER, DESKTOP }
```
jpnurmi, line 9: was there one for WSL?

oSoMoN: No, only "server" and "desktop": https://github.com/canonical/subiquity/blob/main/subiquity/server/server.py#L240

oSoMoN: We might need/want a WSL variant added to subiquity though, I guess this needs to be discussed with @dbungert and @jibel.

jpnurmi (approved):
Ok, LGTM :+1:

packages/subiquity_client/lib/src/types.dart
dbungert: yes, there are 2 wsl variant declarations, on one of the WSL branches.  Also a get() function for getting the variant.

oSoMoN: Ack, thanks for the info. I subscribed to https://github.com/canonical/subiquity/pull/1042, and once it lands and we update our subiquity checkout I'll add the "wsl_setup" and "wsl_configuration" variants as well as a getter.

PR #318, opened 2021-09-17
Add page for selecting guided storage

jpnurmi:
Similarly to Uniquity, this page opens if there are multiple disks when the guided partitioning is selected. After selecting a disk, the wizard continues to the "Write changes to disk" page.

Steps:
- run with `existing-partitions.json`
- navigate to _Installation type_
- keep _Erase disk and install Ubuntu_ selected

PR #332, opened 2021-09-22
Add SubiquityClient.resetStorage()

jpnurmi:


robert-ancell (approved):
Doesn't seem to be used yet, but looks fine!

jpnurmi:
It was noticed in PR https://github.com/canonical/ubuntu-desktop-installer/pull/318#issuecomment-924683628 that disks selected for guided partitioning would pile up when navigating back and forth and selecting different disks. SubiquityClient.resetStorage() will be used to avoid that multiple disks end up being configured for guided partitioning.

Issue #32, opened 2021-03-08
Screen 7 - Turn Off Secure Boot


seb128:
The design for screen n°7, Turn Off Secure Boot

![secure](https://user-images.githubusercontent.com/10974330/110313806-407d1e00-8007-11eb-8473-956d6c02b3a1.png)

Design details https://people.canonical.com/~platform/desktop/installer/7.%20Turn%20Off%20Secure%20Boot/

pieqq:
Hello! A few comments.

# Misleading wording

I don't think the wording "turn off secure boot" is correct.

From what I understand, this screen is meant to add a Machine Owner Key (MOK) into the device's UEFI (the whole process is described on the [Ubuntu Wiki](https://wiki.ubuntu.com/UEFI/SecureBoot)). Selecting the first option would not "turn off" secure boot, but instead generate a key that would have to be enrolled when the device is rebooted.

So I think "Turn off secure boot" should instead be "Configure secure boot".

# More detailed explanations for the Secure Boot MOK enrolling process

I understand the need to keep installation screens nice and tidy, but the concept of adding a MOK into the system is not trivial, and what happens when the device reboots can be really confusing, even for power users. In case you've never done it, you are shown a bunch of blue screens with very cryptic information:

!["Perform MOK management" UEFI screen](https://i.stack.imgur.com/5LMMH.png)

!["Enroll MOK" UEFI screen](https://i.stack.imgur.com/MCsP4.png)

!["Enroll the key?" UEFI screen](https://i.stack.imgur.com/emUvx.png)

!["Enroll the key? Password" UEFI screen](https://i.stack.imgur.com/IvYtO.png)

For security reason, this is not a process that can be automated, therefore the user will have to navigate through these screens and select the correct options, otherwise the 3rd party drivers won't be authenticated/installed properly.

It would be good to have a simple page on the Ubuntu Wiki and point to this from this installer screen, or, if possible, integrate a small help screen so that people without an Internet connection can still figure out how to achieve this.

mwhudson:
Can someone please edit at least the title of this issue to be something less triggeringly misleading?

Feichtmeier:
Fixed by #90

jpnurmi:
Thanks @Feichtmeier :slightly_smiling_face: 

> So I think "Turn off secure boot" should instead be "Configure secure boot".

[...]

PR #333, opened 2021-09-22
Update Configure Secure Boot strings as requested

jpnurmi:
Ref: #32


jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2021-09-22 10-46-56 | Screenshot from 2021-09-22 10-46-40 |

seb128 (approved):
Thanks

PR #335, opened 2021-09-22
Write changes to disk: reset the storage when going back

jpnurmi:
This avoids that multiple disks get configured for guided partitioning when navigating back from the "Write changes to disk" page, and selecting another disk for guided partitioning.


jpnurmi:
Even though this change applies to the current main that is always using guided partitioning, this functionality fits better in https://github.com/canonical/ubuntu-desktop-installer/pull/335 because the storage config should only be reset in guided partitioning mode.

PR #337, opened 2021-09-22
Create a launcher script to set up the right environment

### LINKED ISSUE #323: Desktop Installer doesn't start if yaru-theme-icon is not installed on the host
**Description**
The installer crashes with the following error when yaru-theme-icon is not installed on the host system.

```
root@ubuntu:~# ubuntu-desktop-installer

(ubuntu_desktop_installer:5341): Gtk-WARNING **: 15:41:46.254: Could not load a pixbuf from icon theme.
This may indicate that pixbuf loaders or the mime database could not be found.
**
Gtk:ERROR:../../../../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Format d’image non reconnu (gdk-pixbuf-error-quark, 3)
Bail out! Gtk:ERROR:../../../../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Format d’image non reconnu (gdk-pixbuf-error-quark, 3)
Abandon (core dumped)

```

**Test Case**
1. Boot Ubuntu Desktop ISO of Impish
2. Install ubuntu-desktop-installer snap package
3. Run it from the command line with: ubuntu-desktop-installer
=> Verify that it starts successfully
4. Remove the package yaru-theme-icon 
5. Run it from the command line with: ubuntu-desktop-installer
=> Verify that it fails to start with the trace above

**Issues**
1. It shows that the snap is not self contained and depends on packages installed on the host to run.
2. There is a problem with the fallback used by gdk-pixbuf.

seb128:
Since the snap is classic we are not using the standard gnome launcher.
Without the right variables set we use the gdk-pixbuf loaders cache from
the system. In newer Ubuntu series the png loader is built-in but that's
the case for the version bundled with the installer which leads to have no
png loader available and gtk bailing out while trying to load icons.

Fixes #323

### DIFF
--- a/snap/local/launcher
+++ b/snap/local/launcher
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+source "$SNAP_USER_DATA/.last_revision" 2>/dev/null || true
+if [ "$SNAP_DESKTOP_LAST_REVISION" = "$SNAP_REVISION" ]; then
+  needs_update=false
+else
+  needs_update=true
+fi
+
+if [ "$SNAP_ARCH" = "amd64" ]; then
+  ARCH="x86_64-linux-gnu"
+elif [ "$SNAP_ARCH" = "armhf" ]; then
+  ARCH="arm-linux-gnueabihf"
+elif [ "$SNAP_ARCH" = "arm64" ]; then
+  ARCH="aarch64-linux-gnu"
+elif [ "$SNAP_ARCH" = "ppc64el" ]; then
+  ARCH="powerpc64le-linux-gnu"
+else
+  ARCH="$SNAP_ARCH-linux-gnu"
+fi
+
+# Set cache folder to local path
+export XDG_CACHE_HOME="$SNAP_USER_COMMON/.cache"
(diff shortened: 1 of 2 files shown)


seb128:
The script is basically a copy from what we have in gnome-3-38. Having a script isn't really the nicest thing but it should resolve the issue. 

One alternative option would be to build the cache in the override-build section, include it in the snap and set the environment to point to it from the corresponding yaml section which might be more elegant?

jpnurmi:
Should we merge this?

seb128:
@kenvandine could your review this one? the issue was reported for the WSL wizard where the icon theme is missing and they needed to fallback to png but they workarounded it, that's still a bug just without direct consequence for us atm

kenvandine:
Looks good

PR #341, opened 2021-09-23
WizardPage: fix AppBar theme

### LINKED ISSUE #340: WizardPage: incorrect AppBar text theme
Peek 2021-09-23 11-23

The AppBar theme is not correctly taken from yaru.
Which does not happen with other apps using yaru
grafik

Feichtmeier:
The AppBar theme is not correctly taken from yaru.dart and is overwritten. This sets the theme correctly.

fixed

Closes #340

### DIFF
--- a/packages/ubuntu_wizard/lib/src/widgets/wizard_page.dart
+++ b/packages/ubuntu_wizard/lib/src/widgets/wizard_page.dart
@@ -57,7 +57,11 @@ class WizardPage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return Scaffold(
-      appBar: AppBar(title: title, automaticallyImplyLeading: false),
+      appBar: AppBar(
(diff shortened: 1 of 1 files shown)


jpnurmi:
If you put `Text` widget inside `AppBar`, it inherits the appbar's text theme - the same way it inherits a button's text theme when placed as a child of the button.

Let's try to understand what's going on. The `Text` widget is clearly inheriting some part of appbar's text theme because the font is right. Only the color is different? Could it be an issue in Flutter? Potentially fixed since we're not seeing it?

If we add a piece of code that is not supposed to have any effect, the chances are high that someone will remove the seemingly pointless code later. Even if we left there a comment explaining why it was added, if we are not able to reproduce the issue, it's impossible to know when it's ok to clean it up. :slightly_smiling_face:

jpnurmi:
The problem was sorted out. More details at #340.

PR #342, opened 2021-09-23
Wsl setup run in snap

didrocks:
This is the dart part of the fixes for running WSL setup from the snap. Those are concentrated around the subprocess call for subiquity.

Note that we will need a more recent subiquity to be able to run it without dry-un, but as the variants changes will be needed as well, let’s update the submodule at this time.

### DIFF
--- a/packages/subiquity_client/lib/subiquity_server.dart
+++ b/packages/subiquity_client/lib/subiquity_server.dart
@@ -74,7 +74,7 @@ abstract class SubiquityServer {
     }
 
     _serverProcess = await Process.start(
-      '/usr/bin/python3',
+      'python3',
       subiquityCmd,
       workingDirectory: subiquityPath,
       environment: _getEnvironment(subiquityPath),
--- a/packages/subiquity_client/lib/subiquity_server.dart
+++ b/packages/subiquity_client/lib/subiquity_server.dart
@@ -65,6 +65,11 @@ abstract class SubiquityServer {
     ];
 
     var subiquityPath = p.join(Directory.current.path, 'subiquity');
+    String? workingDirectory;
+    // try using local subiquity
+    if (Directory(subiquityPath).existsSync()) {
+      workingDirectory = subiquityPath;
+    }
 
     // kill the existing test server if it's already running, so they don't pile
     // up on hot restarts
@@ -76,7 +81,7 @@ abstract class SubiquityServer {
     _serverProcess = await Process.start(
       'python3',
       subiquityCmd,
-      workingDirectory: subiquityPath,
+      workingDirectory: workingDirectory,
       environment: _getEnvironment(subiquityPath),
     ).then((process) {
       stdout.addStream(process.stdout);
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -69,6 +69,7 @@ parts:
       - -lib/python*/site-packages/wheel*
       - -lib/python*/site-packages/probert
       - -bin/activate*
+      - -bin/python3*
       - -lib/python3.8/site-packages/__pycache__/six.cpython*
       - -lib/python3.8/site-packages/pip-*.dist-info/RECORD
       - -lib/python3.8/site-packages/wheel-*.dist-info/RECORD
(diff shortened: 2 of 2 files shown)


packages/subiquity_client/lib/subiquity_server.dart
```diff
@@ -64,7 +66,15 @@ abstract class SubiquityServer {
       ...?args,
     ];
 
-    var subiquityPath = p.join(Directory.current.path, 'subiquity');
+    // try using local subiquity
+    String? subiquityPath = p.join(Directory.current.path, 'subiquity');
+    if (subiquityPath != null) {
+      bool exists = await Directory(subiquityPath).exists();
```
jpnurmi, line 72: there's also `existSync()` which is rumoured to be 10x faster because it doesn't need to go through the event loop. a blocking call would be fine here because the UI is not running yet.

jpnurmi (approved):
Thanks! The code LGTM, but I wouldn't mind someone else double-checking the snap config changes.

packages/subiquity_client/lib/subiquity_server.dart
jpnurmi: oh, and since p.join() returns a non-nullable value, we could leave out the null check and use the bang operator: `Directory(subiquityPath!)`

didrocks: > oh, and since p.join() returns a non-nullable value, we could leave out the null check and use the bang operator: Directory(subiquityPath!)

@jpnurmi: this was exactly my first try, which lead to this linter error: `warning • The '!' will have no effect because the receiver can't be null • lib/subiquity_server.dart:71:48 • unnecessary_non_null_assertion`
See https://github.com/canonical/ubuntu-desktop-installer/runs/3689760243. Line 71 in this commit was:
`bool exists = await Directory(subiquityPath!).exists();`

jpnurmi: it looks like the analyzer isn't fully satisfied
```
   info • The operand can't be null, so the condition is always true • lib/subiquity_server.dart:71:23 • unnecessary_null_comparison
   info • Omit type annotations for local variables • lib/subiquity_server.dart:72:7 • omit_local_variable_types
```

packages/subiquity_client/lib/subiquity_server.dart
```diff
@@ -41,13 +41,15 @@ abstract class SubiquityServer {
...
   // Optional environment variables for the server process.
-  Map<String, String> _getEnvironment(String subiquityPath) {
+  Map<String, String> _getEnvironment(String? subiquityPath) {
     // prefer local curtin and probert python modules that are pinned to the
     // correct versions
     final pythonPath = (Platform.environment['PYTHONPATH'] ?? '').split(':');
-    pythonPath.add(subiquityPath);
-    pythonPath.add(p.join(subiquityPath, 'curtin'));
-    pythonPath.add(p.join(subiquityPath, 'probert'));
+    if (subiquityPath != null) {
```
jpnurmi, line 48: what about having the directory exists check here instead?

```diff
diff --git a/packages/subiquity_client/lib/subiquity_server.dart b/packages/subiquity_client/lib/subiquity_server.dart
index b862748..a56c887 100644
--- a/packages/subiquity_client/lib/subiquity_server.dart
+++ b/packages/subiquity_client/lib/subiquity_server.dart
@@ -45,9 +45,11 @@ abstract class SubiquityServer {
     // prefer local curtin and probert python modules that are pinned to the
     // correct versions
     final pythonPath = (Platform.environment['PYTHONPATH'] ?? '').split(':');
-    pythonPath.add(subiquityPath);
-    pythonPath.add(p.join(subiquityPath, 'curtin'));
-    pythonPath.add(p.join(subiquityPath, 'probert'));
+    if (Directory(subiquityPath).existsSync()) {
+      pythonPath.add(subiquityPath);
+      pythonPath.add(p.join(subiquityPath, 'curtin'));
+      pythonPath.add(p.join(subiquityPath, 'probert'));
+    }
     return {'PYTHONPATH': pythonPath.join(':')};
   }
 
```

didrocks: Yes, but we will still need to have the null checking here, won’t we? (same issue due to the analyzer)

jpnurmi: not if we leave out the subiquityPath changes below and keep it non-nullable?

jpnurmi (approved):
LGTM :+1:

PR #344, opened 2021-09-24
subiquity: revert to fix install failures

dbungert:
For yet unknown reasons, running the install from current master
(snap 0+git.d623e92 2021-09-23 (106)) will fail in curtin.
This revert allows install success again while we figure out what happened.


dbungert:
Appears to be a different problem entirely.
https://code.launchpad.net/~dbungert/livecd-rootfs/+git/livecd-rootfs/+merge/409163

PR #346, opened 2021-09-24
WSL Advanced setup: allow mount location end with slash

jpnurmi:
Even the default value of the mount location is `/mnt/`.


didrocks:
This looks good! We will test it and keep you posted. Thanks!

didrocks:
Tested and it works fine. Thanks! :+1:

didrocks (approved):

PR #348, opened 2021-09-24
Build wsl setup

didrocks:
As we won’t run as part of the snap, and we don’t have a full classic environment, build but don’t expose the binary in the snap.

Ship a script that mimic the desktop extension behaviour, we will have a script in the .deb package that mounts the needed snaps at the right place and ship it where it’s needed.

Ensure finally that the socket is passed to subiquity in dry-run, so that the latter creates the directory if not present.

This was tested and runs on WSL impish.


didrocks:
withdrawing this PR, as the branch had one leftover commits which was then taking back multiple commits from main.

PR #349, opened 2021-09-24
Build and expose WSL setup

didrocks:
As we won’t run as part of the snap, and we don’t have a full classic environment, build but don’t expose the binary in the snap.

Ship a script that mimic the desktop extension behaviour, we will have a script in the .deb package that mounts the needed snaps at the right place and ship it where it’s needed.

Ensure finally that the socket is passed to subiquity in dry-run, so that the latter creates the directory if not present.

This was tested and runs on WSL impish.

### DIFF
--- a/packages/ubuntu_wsl_setup/ubuntu-wsl-setup
+++ b/packages/ubuntu_wsl_setup/ubuntu-wsl-setup
@@ -1,0 +1,593 @@
+#!/bin/bash
+
+#################
+# Launcher init #
+#################
+
+START=$(date +%s.%N)
+
+# ensure_dir_exists calls `mkdir -p` if the given path is not a directory.
+# This speeds up execution time by avoiding unnecessary calls to mkdir.
+#
+# Usage: ensure_dir_exists <path> [<mkdir-options>]...
+#
+function ensure_dir_exists() {
+  [ -d "$1" ] ||  mkdir -p "$@"
+}
+
+declare -A PIDS
+function async_exec() {
+  "$@" &
+  PIDS[$!]=$*
+}
+function wait_for_async_execs() {
+  for pid in "${!PIDS[@]}"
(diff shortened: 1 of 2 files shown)


didrocks:
Converting it as a draft as CI don’t pass. I think I understand the assumption of tests and CI about the socket argument, which is wrong IMHO as it prevents using the real system with it. Maybe it’s better to fix it beforehand on a separate branch.

seb128:
The snapcraft ubuntu-wsl-setup parts is quite similar to the ubuntu-desktop-installer one, would it make sense perhaps to just have one so we don't do flutter channel/upgrade/config again? or are those calls basically a no-op if they have already been issued for another parts?

didrocks:
This is not a no-op between the parts, but changing the channel/upgrade is not what is taking most of the time.

I don’t have strong opinion here, but this was to make the separation clearer. If you prefer one part, we can merge them.

didrocks:
ok, so I withdraw latest commit for the socket change as the CI is heavily depending on it. The issue is that we can’t run (except if we create those symlinks…)  on a real system…

This needs to be fixed in a separate PR IMHO and adding also a socket argument to the desktop installer.

Ready for review then!

seb128:
> This is not a no-op between the parts, but changing the channel/upgrade is not what is taking most of the time.
> 
> I don’t have strong opinion here, but this was to make the separation clearer. If you prefer one part, we can merge them.

Let's not delay the delay the merge on that, we can reshuffle later if needed. The snap is slow enough to build so it would be nice to not make it even longer if we can avoid, I will benchmark the alternative options and see what impacts it has and we can decide if we want to do a follow up tweak according to the results

seb128 (approved):
Let's merge, the launcher is non trivial but similar to what we already use

PR #351, opened 2021-09-27
Upgrade `wizard_router` and isolate the dependency into `ubuntu_wizard`

jpnurmi:
`Wizard.next()` returns now a future that completes if/when the wizard
returns back to the awaiting page. This mechanism can be used to reset
the previously selected guided storage from within the guided storage
selection page.

Ref: #318


oSoMoN:
LGTM, thanks!

PR #365, opened 2021-09-28
subiquity_client: restore tests that used to hang

jpnurmi:
The issue referenced by the TODO comments was fixed by #179.

### DIFF
--- a/packages/subiquity_client/test/subiquity_client_test.dart
+++ b/packages/subiquity_client/test/subiquity_client_test.dart
@@ -120,16 +120,15 @@ void main() {
         password: '',
       );
 
-      // TODO: Re-enable once we figure out why _client.send() sometimes hangs in setGuidedStorage()
-      // try {
-      //   await _client.setGuidedStorage(gc); // should throw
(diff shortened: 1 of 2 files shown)


packages/subiquity_client/test/subiquity_client_test.dart
```diff
@@ -120,16 +120,15 @@ void main() {
...
-      //   await _client.setGuidedStorage(gc); // should throw
-      //   // ignore: avoid_catches_without_on_clauses
-      // } catch (e) {
-      //   expect(
-      //       e,
-      //       startsWith(
-      //           'setGuidedStorage({"disk_id":"invalid","use_lvm":false,"password":""}) returned error 500'));
-      // }
+      try {
+        await _client.setGuidedStorage(gc); // should throw
```
oSoMoN, line 124: Instead of a try…catch block, shouldn't we use `expect` and `throwsA` with a `predicate()`?

oSoMoN (commented):
LGTM but I think the test that was re-enabled can be re-written in a more concise and elegant way.

oSoMoN:
LGTM, thanks!

PR #374, opened 2021-09-29
udev: add support for looking up block devices

jpnurmi:
The device partitioning pages need to show pretty disk names.

For example:

```dart
final device = UdevDevice.fromSysname(udev, subsystem: 'block', sysname: 'sda');
print(device?.vendor); // "SanDisk"
print(device?.model); // "Ultra_Fit"
```


robert-ancell (approved):
LGTM

PR #381, opened 2021-10-03
subiquity: update

dbungert:
* langpack support - relevant for Desktop install
* improvements for WSL
* fixes for autoinstall data generation
* deserialization improvement - it is no longer required to send JSON
  fields if the relevant type in subiquity is marked with a default
* fix issue where a disk of size zero could be considered as an possible
  install target
* add some awaits that were missing in many controllers
* round partition sizes downwards to match 1MiB round size numbers


dbungert:
Something is bugged in the analyze step.

In the current PR it demands this change:
https://github.com/canonical/ubuntu-desktop-installer/pull/381/checks?check_run_id=3785202880#step:11:22

I tried adding that change to the commit, and it demands I reverse things back to the current state!

jpnurmi:
Maybe something in our dependencies has changed that causes a slight difference in the generation. It's not critical as the bot will automatically re-generate after anything is pushed to main.

jpnurmi:
Dry-run works fine :+1:

PR #382, opened 2021-10-04
Allow manually running workflows

jpnurmi:
https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow


jpnurmi:
Disclaimer: I have never tried this before.

jpnurmi:
I tested with my fork and it seems to work:
image

PR #334, opened 2021-09-22
WIP: Allocate disk space

jpnurmi:
TODO:
- [ ] Format check boxes
- [ ] Primary vs. logical
- [ ] Beginning vs. end
- [ ] Unallocated areas
- ...

https://user-images.githubusercontent.com/140617/134344831-c11c849c-97c1-4ecc-946f-042d24ba96cc.mp4

Close: #36

PR #384, opened 2021-10-04
subiquity_client: add storage/v2 API

jpnurmi:
Split off from the large _Allocate disk space_ PR #334.


jpnurmi:
Requires [dbungert:subiquity/fr-1710-storage](https://github.com/dbungert/subiquity/tree/fr-1710-storage) but a preliminary review is already welcome.

PR #385, opened 2021-10-04
Fix dropdown's size and interactive area

jpnurmi:
`DropdownButtonFormField` adds undesired non-interactive padding. Use an explicit `InputDecorator` instead to achieve the Yaru looks and a matching size with `TextField`.


jpnurmi:
On the left, the interactive area is marked with red. Clicking outside the interactive area but within the dropdown's visual boundaries does not open the dropdown.

| Before | After |
|---|---|
| Screenshot from 2021-10-04 23-08-56 | Screenshot from 2021-10-04 23-10-10 |

PR #386, opened 2021-10-05
Re-generate outdated mock

jpnurmi:


jpnurmi:
What the... what was outdated yesterday is no longer outdated today. :thinking:

PR #387, opened 2021-10-05
Try to get deterministic output from melos generate

jpnurmi:
- upgrade and pin mockito & build_runner
- verbose output for easier inspecting


jpnurmi:
This doesn't seem to help either.

PR #393, opened 2021-10-05
DropdownBuilder: allow null value

jpnurmi:
For the "do not use" partition format option, which is presented as a
null value.


jpnurmi:
Temporarily marked as a draft to avoid conflicts with #334

PR #394, opened 2021-10-06
Adapt to new subiquity WSL API

didrocks:
subiquity is now following the <section>_<key> from wsl.conf for its API
to avoid another internal mapping. Follow that change.
Make it thus camelCase for properties separating section and key like
sectionKey.

Keeping it as a draft until https://github.com/canonical/subiquity/pull/1066 is merged and referenced in submodule.


packages/subiquity_client/lib/src/types.dart
```diff
@@ -254,10 +254,10 @@ class StorageResponse with _$StorageResponse {
...
 class WSLConfigurationBase with _$WSLConfigurationBase {
   const factory WSLConfigurationBase({
-    @JsonKey(name: 'custom_path') String? customPath,
-    @JsonKey(name: 'custom_mount_opt') String? customMountOpt,
-    @JsonKey(name: 'gen_host') bool? genHost,
-    @JsonKey(name: 'gen_resolvconf') bool? genResolvconf,
+    @JsonKey(name: 'automount_root') String? automountRoot,
+    @JsonKey(name: 'automount_options') String? automountOptions,
+    @JsonKey(name: 'network_generatehosts') bool? networkGeneratehosts,
+    @JsonKey(name: 'networkGenerateresolvconf') bool? networkGenerateresolvconf,
```
jpnurmi, line 260: network_generateresolvconf

jpnurmi: Actually, screw that! I have it on my TODO-list to make all these use automatic snake_case renames. We could go straight for this to let the JSON serializer do it for us:
```dart
@freezed
class WSLConfigurationBase with _$WSLConfigurationBase {
  @JsonSerializable(fieldRename: FieldRename.snake)
  const factory WSLConfigurationBase({
    String? automountRoot,
    String? automountOptions,
    bool? networkGeneratehosts,
    bool? networkGenerateresolvconf,
  }) = _WSLConfigurationBase;

  factory WSLConfigurationBase.fromJson(Map<String, dynamic> json) =>
      _$WSLConfigurationBaseFromJson(json);
}
```

packages/subiquity_client/lib/src/types.dart
```diff
@@ -269,12 +269,13 @@ class WSLConfigurationAdvanced with _$WSLConfigurationAdvanced {
   const factory WSLConfigurationAdvanced({
```
jpnurmi, line 269: same here, we could use `@JsonSerializable(fieldRename: FieldRename.snake)` to avoid the error-prone manual field name mapping.

PR #403, opened 2021-10-07
Use automatic JSON field rename to/from snake_case

jpnurmi:
`json_serialiable` offers automatic renaming to avoid error-prone manual mapping of field names.


jpnurmi:
Temporarily marked as a draft to avoid conflicts with #334

oSoMoN:
LGTM, thanks!

Issue #373, opened 2021-09-29
Hide or disable unimplemented UI elements


jpnurmi:
- [ ] repair ubuntu
- [ ] download third-party checkbox
- [ ] advanced options for guided partitioning

jpnurmi:
The two first items are not yet in the mainline but they have been done in the storage branch.

Issue #362, opened 2021-09-28
Who are you page: login strategy is not stored anywhere


jpnurmi:
On the _Who are you?_ page, the user can select between the following options:
- _Log in automatically_
- _Require my password to log in_

This information is never passed to subiquity or stored anywhere.

seb128:
I've opened a subiquity request to add a backend for that on https://bugs.launchpad.net/subiquity/+bug/1946490

N0rbert:
Really needed.

jpnurmi:
Fixed by:
- https://github.com/canonical/ubuntu-desktop-installer/pull/1196

PR #407, opened 2021-10-07
Hide unsupported login "strategy"

jpnurmi:
Ref: #373
Ref: #362


jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2021-10-07 13-48-29 | Screenshot from 2021-10-07 13-48-15 |

PR #408, opened 2021-10-07
Hide unsupported 3rd-party software selection

jpnurmi:
Ref: #373


jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2021-10-07 13-55-00 | Screenshot from 2021-10-07 13-55-04 |

Issue #368, opened 2021-09-28
Multiple copies of the UI can be opened


dbungert:
If you keep double clicking the Install icon on the desktop, you'll keep getting new copies of the UI.

jpnurmi:
_Originally posted by @seb128 in https://github.com/canonical/ubuntu-desktop-installer/pull/422#pullrequestreview-775949426_

>Thanks and sorry for not thinking about that when I reviewed the initial change. Indeed since the snap is classic it should be able to own the dbus name without needing snapd permissions

How should we proceed?

seb128:
I think that's fixed, it seems to do the right thing here. The slot isn't needed since as a classic snap owning the dbus name is blocked by snap

PR #409, opened 2021-10-07
Activate an existing app instance if appropriate

jpnurmi:
https://docs.gtk.org/gtk3/migrating-libunique.html

Close: #368


jpnurmi:
Maybe it would be better to allow multiple app instances in debug mode to make development, debugging, and testing easier.

jpnurmi:
Thanks! :slightly_smiling_face:

PR #411, opened 2021-10-08
Fix subiquity error in WSL Profile setup

jpnurmi:
The WSL Profile setup UI has no hostname field, but subiquity expects that none of the `IdentityData` fields are null. Specify the same empty defaults that are used in `subiquity/common/types.py`.

### DIFF
--- a/packages/subiquity_client/test/subiquity_client_test.dart
+++ b/packages/subiquity_client/test/subiquity_client_test.dart
@@ -176,12 +176,8 @@ void main() {
       expect(id.cryptedPassword, '');
       expect(id.hostname, 'ubuntu-desktop');
 
-      newId = IdentityData(
-        realname: '',
-        username: '',
-        cryptedPassword: '',
-        hostname: '',
-      );
+      // empty defaults for null values
+      newId = IdentityData();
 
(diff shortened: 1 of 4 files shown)


packages/subiquity_client/test/subiquity_client_test.dart
```diff
@@ -176,12 +176,8 @@ void main() {
...
       expect(id.hostname, 'ubuntu-desktop');
 
-      newId = IdentityData(
-        realname: '',
-        username: '',
-        cryptedPassword: '',
-        hostname: '',
-      );
+      // empty defaults for null values
+      newId = IdentityData();
```
didrocks, line 180: Yes, this will prevent us getting this error again :)

didrocks (approved):
Indeed, we didn’t have any drastic change in that regard on the subiquity side. Weird. As we inherit it from the server model, it makes sense to default it as empty. Looks good, test pass. Let’s merge.

jpnurmi:
I opened https://github.com/canonical/ubuntu-desktop-installer/issues/412 because it's a good practice to use a nullable type for every single field. :)

PR #420, opened 2021-10-08
Integration test for the WSL OOBE setup wizard

jpnurmi:
A full end-to-end integration test that uses Flutter's integration test driver to run the app, navigate through the wizard, and enter test values. Finally, it verifies that the appropriate values have been written to `.subiquity/etc/wsl.conf` and `.subiquity/etc/ubuntu-wsl.conf`.

Test cases:
- Basic setup: _Select language_ and fill up _Profile setup_ 
  - TODO: is the profile information stored anywhere in dry-run mode?
- Advanced setup: select _Show advanced options_ and fill up _Advanced setup_
- Reconfiguration: pass `--reconfigure` and fill up _Configuration UI_

To run the test outside of IDE, run `flutter test integration_test/` in `packages/ubuntu_wsl_setup`.


jpnurmi:
https://user-images.githubusercontent.com/140617/136602013-b968d96c-a7ac-405e-b82f-490ea0536919.mp4

packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
```diff
@@ -0,0 +1,379 @@
...
+      await tester.pump();
+    }
+    await tester.pumpAndSettle();
+
+    await tester.tapContinue();
+  }
+
+  Future<void> testProfileSetup(
+    WidgetTester tester,
+    IdentityData? profile, {
```
didrocks, line 50: I didn’t see any code path where you call it with null, is it intended to be nullable?
Same for the 3 other fields

packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
```diff
@@ -0,0 +1,379 @@
...
+          equals(config.motdWSLnewsenabled.toString()));
+    });
+  }
+
+  testWidgets('basic setup', (tester) async {
+    app.main(<String>[]);
+    await waitForSubiquityServer();
+    await tester.pumpAndSettle();
+
+    await testSelectYourLanguage(tester, language: 'Français');
```
didrocks, line 233: Merci ! ;)

packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
```diff
@@ -0,0 +1,379 @@
...
+
+    await verifyBaseConfig(config);
+  });
+
+  testWidgets('reconfiguration', (tester) async {
+    app.main(<String>['--reconfigure']);
+    await waitForSubiquityServer();
+    await tester.pumpAndSettle();
+
+    await testAdvancedSetup(tester);
```
didrocks, line 298: Mind adding a comment that this is testing the first screen (WSLConfigurationBase)?

packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
```diff
@@ -0,0 +1,379 @@
...
+      '.subiquity',
+      'etc',
+      fileName,
+    );
+    final lines = File(path).readAsLinesSync();
+    return test(Config.fromStrings(lines));
+  }
+
+  Future<void> verifyBaseConfig(WSLConfigurationBase config) async {
+    await verifyIniFile('wsl.conf', (ini) {
```
didrocks, line 198: Note: we are more used to use golden files as best practice where you can pass scenarios (base inputs) and serialize the results in a file, then comparing that file (with a mecanism to refresh them with running the tests with -update). Not a blocker as this is also valid (a little bit less agile), but maybe something to keep in mind? I’m happy to expand my explanations if needed

didrocks (commented):
> TODO: is the profile information stored anywhere in dry-run mode?
Indeed, this is a task that is on Patrick’s roadmap. Please have a chat with him (I think writing to a file in .subiquity for both actions would be good, or adding a mock adduser command which does this writing)

Some questions & small remarks attached above

jpnurmi:
Hmm, sometimes the .conf files don't exist. I'm not sure if it's because the GUI is abruptly killing the subiquity process or what. Needs investigation...

jpnurmi:
Rebased, cleaned up, and squashed. Depends on https://github.com/canonical/ubuntu-desktop-installer/pull/504.

packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
jpnurmi: The idea was that one could surf through pages without having to pass anything in test cases where that page is irrelevant.

packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
jpnurmi: It turns out Flutter exposes the information whether `--update-goldens` was passed, so we can use the "standard" `flutter-test --update-goldens integration_test`, after all. There are instructions in the [README](https://github.com/canonical/ubuntu-desktop-installer/blob/7e4f79fe0ad27aae12439b9244f51c2f9e5f28ca/packages/ubuntu_wsl_setup/integration_test/README.md#update-golden-files).

didrocks:
LGTM, the golden file is testing the very important use-case (what should I say, even the MAIN and only one we all care, right? :p)

I wonder if we can have an integration test doing no change and checking that no change is done on the wsl.conf. We can have with a pre-existing wsl.conf, and one without one. What do you think?

jpnurmi:
> LGTM, the golden file is testing the very important use-case (what should I say, even the MAIN and only one we all care, right? :p)

Yeah, it's great especially with the update flag when needed.

> I wonder if we can have an integration test doing no change and checking that no change is done on the wsl.conf. We can have with a pre-existing wsl.conf, and one without one. What do you think?

Sounds good! :+1:

didrocks (approved):

PR #424, opened 2021-10-11
DEENG-1 - Added option to start setup in text mode

jibel:
By default WSL setup starts in graphics mode, there is now an option
-t|--text to start it in text mode.

Co-authored-by: Didier Roche <didrocks@ubuntu.com>


didrocks:
Tests pass. We did test it manually and it works as expected. Merging.

Issue #390, opened 2021-10-05
Generated hostname not based on the machine?


seb128:
Doing an installation using canary the hostname generated is -ubuntu which using ubiquity is a more unique identify PC-QR35-.... which seems based on the hardware somehow?

PR #425, opened 2021-10-11
Who are you: base computer name on the DMI product name

jpnurmi:
Even though the computer name ends up being the hostname on the target
system, the initial live session hostname is not a good source for
generating one because it's always "ubuntu".

Close: #390


oSoMoN:
LGTM, thanks!

Issue #401, opened 2021-10-06
Don't restrict the languages to the available translations


seb128:
We current have a limited set of translations but are able to install any langpack available so we shouldn't limit the choice on the first screen

seb128:
The subiquity list is quite limite as well though, no italian, turkish, portugues, japanese, arabic, ...

@subiquity maintainers, is your languagelist limited by design or should those be added?

dbungert:
Language selection is two things
* what language should the installer run in?
* what language should the target system be configured for?

For the former, a fixed set of languages makes sense - we have the translations we have.  For the latter, any valid value should be possible.  See also @mwhudson's desired locale overhaul.

long-chung:
@oSoMoN I think it is what it is. The message would have to be in English(?) anyway so defeats the purpose.
I think we should just display the available languages we do have as this would be updated later in subsequent releases?

seb128:
The recent update resolve the issue described there I think, let's close it now

PR #430, opened 2021-10-13
Use subiquity's languagelist

jpnurmi:
image

Ref: #401

### DIFF
--- a/.github/workflows/flutter-ci.yml
+++ b/.github/workflows/flutter-ci.yml
@@ -58,6 +58,8 @@ jobs:
     steps:
 
     - uses: actions/checkout@v2
+      with:
+        submodules: recursive
 
     - uses: subosito/flutter-action@v1
       with:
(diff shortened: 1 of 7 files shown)


gunnarhj:
Please No!

The proper languagelist in the desktop installer is a list which represents the languages in Ubuntu's language packs. Whether subiquity strings have (yet) been translated into respective language should not be a criterion.

jpnurmi:
This was originally proposed as a stop-gap solution to get more languages on the list, but we have now generated lots of .arbs to populate the list: https://github.com/canonical/ubuntu-desktop-installer/pull/456.

gunnarhj:
Ok, that makes more sense. But not understanding 3-letter language codes seems as a Flutter bug which should be fixed.

jpnurmi:
Agreed, I opened https://github.com/flutter/flutter/issues/92818 earlier.

Issue #402, opened 2021-10-07
CI: open PRs for changes in generated files


jpnurmi:
> This thing is unfortunate indeed :/
>
> Thanks, I won’t need to look for the annotation format that I need to do on another project ;)
> Maybe as the known issue is only about swapping 2 lines, we can show in the annotation number of lines that changed? +/-? If i’ts +1/-1, we know without looking at the run annotations, depending on the number of files and line changes that this is probably a false positive (lines swapped) or not?
>
> It makes sense, se we know that this job will probably always fail (for now), but with the annotation, that can help.
>
> on
> c) The "generate" workflow can be manually run to let the bot regenerate files (it's no longer automatic to avoid ping pong changes).
>
> I would suggest you can do soomething similar that I use in multiple of my projects for l10n, readme generations… (the generators in Go are stable and so we can enforce it on PR analyzes test)
> - on main branch, with the push event, run the generator,
> - commit the diff,
> - push a branch with a fixed name
> - propose a PR (it will be always the same PR in your case) with the changes
>
> Then, only when there is a real diff, you can review and press the merge button.
>
> This isn’t ideal ofc, due to the generator unstability but at least, you can keep a single branch, which will force-push and a single PR. I think this will be a little bit less silent and won’t be in the camp of "Oh, I didn’t run the action manually for 3 weeks and forgot there were generations diff to be done". What do you think?
> One example: https://github.com/ubuntu/adsys/blob/main/.github/workflows/auto-updates.yaml

_Originally posted by @didrocks in https://github.com/canonical/ubuntu-desktop-installer/issues/395#issuecomment-937531109_

PR #437, opened 2021-10-15
CI: open a PR when outdated generated files are detected

jpnurmi:
Close: #402

### DIFF
--- a/.github/workflows/flutter-ci.yml
+++ b/.github/workflows/flutter-ci.yml
@@ -117,9 +117,9 @@ jobs:
           exit 1
         fi
 
-    - name: Commit outdated generated files
-      if: github.event_name == 'workflow_dispatch'
-      uses: actions-js/push@master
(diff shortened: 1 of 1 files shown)


seb128 (approved):
Alright. Depending on actions from individual maintainers on github does sound like somewhat risky to me but it seems a common practice and we already have cases in our workflows and in other projects so let's go for it

jpnurmi:
Very good point.

- "checkout" and "codecov" actions are from verified publishers
- "cla check" action is ours
- "flutter" action is from an individual, but would be trivial to replace with the flutter snap
- the current "push" action, to be replaced by the "create pr" action, are both from individuals. they are not trivial to replace, but neither is critical. they are both there for convenience, basically

A related article: [Protect against malicious usage of GitHub Actions](https://github.com/UnlyEd/next-right-now/discussions/223)

PR #438, opened 2021-10-18
Upgrade wizard_router for scoped route conditions

jpnurmi:
The complexity of `Wizard.onNext` had increased to a point where it got
hard to follow the route conditions that were placed separately from the
list of routes.

The new version is a bit more verbose, but makes it easier to read the
conditions that are scoped together with the respective route, which is
a welcome improvement because several more conditions are coming later
(internet, secure boot, security key, active directory, ...)


jpnurmi:
Fixes https://github.com/canonical/ubuntu-desktop-installer/issues/469 too.

PR #441, opened 2021-10-19
Add missing confirmation dialog for creating a new partition table

jpnurmi:


packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
```diff
@@ -8,6 +8,35 @@ import 'package:ubuntu_wizard/widgets.dart';
...
+          ),
+          OutlinedButton(
+            child: Text(lang.continueAction),
+            onPressed: () => Navigator.of(context).pop(true),
+          ),
+        ],
+      );
+    },
+  );
+  return result == true;
```
oSoMoN, line 37: `result` being a boolean, this is redundant. You could just return `result`.

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_widgets.dart
```diff
@@ -370,6 +370,22 @@ class PartitionButtonRow extends StatelessWidget {
...
+    final disk = model.selectedDisk!;
+    if (disk.ptable != null) {
+      final confirmed = await showConfirmationDialog(
+        context,
+        title: lang.newPartitionTableConfirmationTitle,
+        message: lang.newPartitionTableConfirmationMessage,
+      );
+      if (!confirmed) return;
+    }
+    model.reformatDisk(model.selectedDisk!);
```
oSoMoN, line 387: `model.selectedDisk!` can be replaced by `disk`

packages/ubuntu_desktop_installer/test/allocate_disk_space_page_test.dart
```diff
@@ -263,6 +263,41 @@ void main() {
...
+    expect(find.byType(AlertDialog), findsOneWidget);
+    verifyNever(model.reformatDisk(disk));
+
+    final continueButton = find.widgetWithText(
+      OutlinedButton,
+      tester.ulang.continueAction,
+    );
+    expect(continueButton, findsNWidgets(2));
+
+    await tester.tap(continueButton.last);
```
oSoMoN, line 294: Are we guaranteed that this will always be the right continue button?

oSoMoN (commented):
LGTM, and works as expected. Just a couple of minor comments and a question before it's ready for merging.

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
jpnurmi: It can be null when dismissing the dialog by clicking outside of it.

packages/ubuntu_desktop_installer/test/allocate_disk_space_page_test.dart
jpnurmi: Good point. I made it to explicitly look for an ancestor of AlertDialog.

oSoMoN (approved):

PR #442, opened 2021-10-19
Added '--prefill' command line option to WSL.

cnihelton:
Hi @jpnurmi !

I am working on prefilling the WSL OOBE with user and locale information acquired from Windows through the launcher, as you may know. Since in WSL Subiquity is started by the OOBE I had to touch this code base. Could you take some time to code review it?

The proposed changes do the following:
- Check if the file passed by CLI exists before passing to the server (not sure if this kind of check belongs here, though).
- When server starts, log.info also it's CLI options.

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/main.dart
+++ b/packages/ubuntu_wsl_setup/lib/main.dart
@@ -7,10 +7,23 @@ import 'app.dart';
 void main(List<String> args) {
   final options = parseCommandLine(args, onPopulateOptions: (parser) {
     parser.addFlag('reconfigure');
+    parser.addOption(
+      'prefill',
+      valueHelp: 'prefill',
+      help: '''
+Path of the YAML file containing prefill information, which
+feeds the installer with partial information to prefill the
+screens, yet allowing user to overwrite any of those during setup.
+  ''',
+    );
   })!;
   var variant = Variant.WSL_SETUP;
+  List<String>? serverArgs;
   if (options['reconfigure'] == true) {
     variant = Variant.WSL_CONFIGURATION;
+  }
+  if (options['prefill'] != null) {
+    serverArgs = ['--prefill', options['prefill']];
   }
   runWizardApp(
     UbuntuWslSetupApp(
@@ -21,5 +34,6 @@ void main(List<String> args) {
     subiquityClient: SubiquityClient(),
     subiquityServer: SubiquityServer.wsl(),
     onInitSubiquity: (client) => client.setVariant(variant),
+    serverArgs: serverArgs,
   );
 }


packages/subiquity_client/lib/subiquity_server.dart
```diff
@@ -94,7 +94,9 @@ abstract class SubiquityServer {
       stderr.addStream(process.stderr);
       return process;
     });
-    log.info('Starting server (PID: ${_serverProcess!.pid})');
+    log.info(
+      'Starting server (PID: ${_serverProcess!.pid}) with args: $subiquityCmd',
```
jpnurmi, line 98: this is nice. :+1: it would be nice to have it in a separate commit, though

packages/ubuntu_wsl_setup/lib/main.dart
```diff
@@ -7,11 +9,26 @@ import 'app.dart';
...
   })!;
   var variant = Variant.WSL_SETUP;
+  List<String>? serverArgs;
   if (options['reconfigure'] == true) {
     variant = Variant.WSL_CONFIGURATION;
   }
+  if (options['prefill'] != null) {
+    final prefillFile = options['prefill'];
+    final fileExists = File(prefillFile).existsSync();
+    if (fileExists) serverArgs = ['--prefill', options['prefill']];
```
jpnurmi, line 30: should we print an error if the file doesn't exist, to avoid silently ignoring the option?

how does the server handle it if the file doesn't exist? I'm wondering if simply passing the option through and letting the server throw an error would result in a clear enough error message?

packages/subiquity_client/lib/subiquity_server.dart
CarlosNihelton: I'll do some git tricks to split that commit in two.

packages/ubuntu_wsl_setup/lib/main.dart
```diff
@@ -7,11 +7,24 @@ import 'app.dart';
 void main(List<String> args) {
   final options = parseCommandLine(args, onPopulateOptions: (parser) {
     parser.addFlag('reconfigure');
+    parser.addOption(
+      'prefill',
+      valueHelp: 'prefill',
+      help: '''
```
jpnurmi, line 13: I'd suggest also `hide: true` since it's not for users (same as `initial-route`) which is for testing.

jpnurmi (approved):
LGTM.

Feel free to ignore the suggestion if you prefer to keep the option visible in `--help`.

packages/ubuntu_wsl_setup/lib/main.dart
CarlosNihelton: Keeping it visible might be useful for development/testing. Let's keep it as is.

Thank you again, JP!

Issue #453, opened 2021-10-29
Incorrect keyboard navigation on lists


seb128:
Unsure if that's a flutter issue or an installer one. 

On the welcome screen for example, using the TAB key the focus moves to the next item in the languagelist. The traditional behaviour would be to have TAB switching to the next widget, so the next button

On the keyboard layout screen, TAB alternate between the lists by moving to the next items

PR #454, opened 2021-10-29
Improve keyboard layout tab navigation

jpnurmi:
Traverse lists from top-to-bottom instead of jumping between lists. The
real fix would be to make the lists act like a singular focus node, but
that's hard to fix outside of Flutter. This fix makes the navigation a
bit less embarrassing, meanwhile. :)

Ref: #453


robert-ancell (approved):
LGTM

PR #457, opened 2021-11-01
Specify country code in .arb files

jpnurmi:
Instead of supplying multiple .arb files per language to get a full
locale, construct full locales with help of a country code field
specified in the base .arb file.

This allows us to remove a lot of extra .arb files that were basically
unused, but had to be there to satisfy Flutter (flutter/flutter#92731).


packages/ubuntu_wizard/lib/src/l10n/ubuntu_pt_BR.arb
```diff
@@ -1,5 +1,6 @@
 {
   "@@locale": "pt_BR",
+  "countryCode": "BR",
```
jpnurmi, line 3: Technically, the country code is not necessary for .arb files that already specify a full locale. I specified it here regardless, to make it consistent with others.

packages/ubuntu_wizard/lib/l10n.dart
```diff
@@ -52,12 +52,15 @@ class LocalizedLanguage {
 /// [locales] must contain the base locale i.e. the template .arb locale.
 Future<List<LocalizedLanguage>> loadLocalizedLanguages(
     List<Locale> locales) async {
-  assert(locales.contains(_kBaseLocale));
```
jpnurmi, line 55: I left this out because it wasn't really important and the base locale is "en_US", but AppLocalizations.supportedLocales no longer contains full locales (because .arb has only the country code). We extract the full locale below instead.

packages/ubuntu_wizard/lib/src/l10n/ubuntu_pt.arb
```diff
@@ -1,5 +1,6 @@
 {
   "@@locale": "pt",
+  "countryCode": "PT",
 
-  "languageName": ""
+  "languageName": "Português"
```
oSoMoN, line 5: The language name was originally left empty to ensure the language wouldn't be displayed in the UI, because there currently are no translations for it (only for Brazilian Portuguese).
It feels a bit weird that I can select Portuguese and the UI is 100% English.

However my understanding is that we're heading towards a list of languages that reflects the availability of langpacks, not the current translations for the app itself. So I suppose it doesn't hurt to keep this entry?

jpnurmi: A mass import of translations from Ubiquity will follow. (with help of https://github.com/canonical/ubuntu-desktop-installer/pull/463)

oSoMoN:
LGTM, thanks!

PR #460, opened 2021-11-01
Enable the prefer_const_constructors linter rule and fix findings

jpnurmi:
This is a useful linter rule that makes sure we take advantage of Dart's compile-time constants where possible.


oSoMoN:
LGTM, thanks!

PR #462, opened 2021-11-02
Fix wrong assumptions in language list tests

jpnurmi:
- Fix the widget test to not assume that all tested languages
  (English, French, Italian) are on the screen at the same time,
  because they won't be after the list of languages grows

- Fix the model test to not assume that en_US is the first language


oSoMoN:
LGTM, thanks!

PR #467, opened 2021-11-04
Turn off the public_member_api_docs linter rule

jpnurmi:
First of all, this linter rule is so noisy that it hides useful linter
issues. There are  linter issues slipping in due to `--no-fatal-infos`,
which is used in the CI to ignore `public_member_api_docs`.

Secondly, in an app project where there tend to be areas that don't need
docs, this linter rule results in not-so-useful docs that basically
repeat the same what the member name states, just to silence the linter
without adding any real value. For example:

    /// Does something.
    void doSomething() { ... }

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -13,3 +13,4 @@ linter:
     constant_identifier_names: false
     lines_longer_than_80_chars: false
     prefer_const_constructors: true
+    public_member_api_docs: false


Feichtmeier:
I appreciate this change. Currently the warnings are flooding the workspace and one is distracted by actually new warnings maybe created by yourself or just new ones in general that should be removed fast. :)

CarlosNihelton:
Endorsed. More important warnings are being shadowed out by this `public_member_api_docs` rule and the `effective_dart` depracation notice recommends `package:lints` for Dart-only and `package:flutter_lints` which extends the previous for Flutter, with up-to-date rules.

robert-ancell (approved):
LGTM

jpnurmi:
Instead of switching to `flutter_lints` in the same PR, I have reduced the scope of this PR to turn off `public_member_api_docs` and nothing else.

oSoMoN:
LGTM, thanks!

jpnurmi:
Thanks!

Issue #469, opened 2021-11-04
User can go back on first step


jibel:
**Description**
On first step of the installer, it it possible under certain conditions to go back, which leads to a blank page

**Test Case**
1. Launch the installer
2. Select a language 
3. Click on next
4. Click on back twice

**Actual result**
The installer displays a blank page

**Expected result**
The go back button is greyed and the user cannot go back.

jpnurmi:
Fixed by #438

PR #471, opened 2021-11-04
WSL: disable the back button on the Select your language -page

jpnurmi:
This is supposed to be automatically taken care of by `wizard_router`,
but it doesn't trigger a rebuild as it should when navigating back, so
the back button is left enabled. It might not be easy to fix, so this
change explicitly disables the back button to make sure the user can't
attempt to navigate backwards to a page that doesn't exist.

Fixes: #469

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/pages/select_language/select_language_page.dart
+++ b/packages/ubuntu_wsl_setup/lib/pages/select_language/select_language_page.dart
@@ -84,7 +84,7 @@ class _SelectLanguagePageState extends State<SelectLanguagePage> {
         ),
       ),
       actions: [
-        WizardAction.back(context),
+        WizardAction.back(context, enabled: false),
         WizardAction.next(
           context,
           onActivated: () {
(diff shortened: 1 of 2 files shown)


jpnurmi:
https://user-images.githubusercontent.com/140617/140335362-e9202aae-ea60-486f-a626-a87c57a83381.mp4

didrocks:
This looks good to me BUT what about the reconfiguration mode where the first page is the wsl base option one?

(btw, we don’t get it running yet as you probably saw on the other bug report)

jpnurmi:
> This looks good to me BUT what about the reconfiguration mode where the first page is the wsl base option one?

Thanks, good catch. I'm looking into fixing it properly in the `wizard_router` package.

jpnurmi:
The root cause was fixed in the [`wizard_router`](https://github.com/jpnurmi/wizard_router/commit/ffbe822755094be0916fed8392f1535fd1fac8a1) package. The issue will be fixed by #438 that upgrades the package to the latest version.

PR #472, opened 2021-11-04
UdevService: prepare for network devices

jpnurmi:
While storage devices are looked up using sysnames, network devices are
instead looked up using syspaths provided by Network Manager. This
commit makes it possible to specify whether it's a sysname or syspath
when querying model and vendor names.

Extracted from https://github.com/canonical/ubuntu-desktop-installer/pull/77.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/services/udev_service.dart
+++ b/packages/ubuntu_desktop_installer/lib/services/udev_service.dart
@@ -12,20 +12,30 @@ class UdevService {
   void dispose() => _udev.dispose();
 
   /// The combined model and vendor name of the device.
-  String fullName(String sysname) {
+  String fullName({String? sysname, String? syspath}) {
     return <String?>[
-      modelName(sysname),
-      vendorName(sysname),
+      modelName(sysname: sysname, syspath: syspath),
+      vendorName(sysname: sysname, syspath: syspath),
     ].where((p) => p?.isNotEmpty == true).join(' ');
   }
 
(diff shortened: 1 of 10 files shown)


packages/ubuntu_desktop_installer/lib/services/udev_service.dart
```diff
@@ -12,20 +12,30 @@ class UdevService {
...
   /// The vendor of the specified device (e.g. 'sda' -> 'SanDisk').
-  String? vendorName(String sysname) => _getDevice(sysname)?.vendor;
+  String? vendorName({String? sysname, String? syspath}) {
+    return _getDevice(sysname: sysname, syspath: syspath)?.vendor;
+  }
 
-  UdevDevice? _getDevice(String sysname) {
-    return UdevDevice.fromSysname(_udev, subsystem: 'block', sysname: sysname);
+  UdevDevice? _getDevice({String? sysname, String? syspath}) {
+    assert(sysname != null || syspath != null);
```
robert-ancell, line 33: It seems you can only provide either a sysname or a syspath. This suggests either:
- There should be an assertion that both weren't provided.
- Instead of two parameters, use one with an enum, e.g. `_getPath(syspath, Format.syspath)
- Have two methods, `_getPathFromSyspath`, `_getPathFromSysname`

jpnurmi: I ended up using polymorphism to avoid moving the if-clause from one place to another. :)

robert-ancell (approved):
LGTM

Issue #468, opened 2021-11-04
When running WSL OOBE on an already setup system, u-d-i is in setup mode, not reconfigure


didrocks:
The TUI is correctly showing up the reconfigure path.

The GUI though, is always running in setup mode and doesn’t use the mode returned by the backend. The GUI needs to do a GET on client_variant endpoint.

jpnurmi:
Yeah, there must have been a misunderstanding because the UI expects a `--reconfigure` command-line argument.

didrocks:
yeah, basically this is the daemon which is running on the target (you can imagine the client being somewhere else :)), so this is the decision factor. Also, it avoids logic duplication between the GUI and TUI (we already have enough of them :p)

Sorry for the misunderstanding.

jpnurmi:
No worries, it's ok! :) I'll take a look

PR #473, opened 2021-11-05
Add SubiquityClient.variant() for querying the client variant

jpnurmi:
Ref: #468


didrocks:
Thanks for the change! Still puzzled by the dartx package to add extra features to existing class :)

Changes looks good, test pass, I think you did rerun integration tests :p
:+1:

didrocks (approved):

PR #474, opened 2021-11-05
SubiquityServer: pass server environment from the outside

jpnurmi:
In WSL, DRYRUN_RECONF environment variable can be used to control
whether a dry-run instance of Subiquity runs in reconfiguration mode.

Allow WSL OOBE to pass a suitable environment from the main entry-point
where the necessary arguments are available.

### DIFF
--- a/packages/subiquity_client/lib/subiquity_server.dart
+++ b/packages/subiquity_client/lib/subiquity_server.dart
@@ -40,10 +40,9 @@ abstract class SubiquityServer {
       ? '/run/subiquity/socket'
       : p.join(Directory.current.path, 'test/socket');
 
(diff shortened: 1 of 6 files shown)


jpnurmi:
With this we can pass the environment from main() like this:
```dart
serverEnvironment: {
  if (!isLiveRun(options) && options['reconfigure'] == true)
    'DRYRUN_RECONFIG': 'true',
}
```

jpnurmi:
Also, ideally, SubiquityServer wouldn't know which app is running. This is a step towards making it agnostic of the host app.

CarlosNihelton (approved):
LGTM.

PR #475, opened 2021-11-05
WSL setup: query client variant from Subiquity

### LINKED ISSUE #468: When running WSL OOBE on an already setup system, u-d-i is in setup mode, not reconfigure

jpnurmi:
Fixes #468

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/app.dart
+++ b/packages/ubuntu_wsl_setup/lib/app.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:subiquity_client/subiquity_client.dart';
 import 'package:ubuntu_wizard/settings.dart';
 import 'package:ubuntu_wizard/utils.dart';
 import 'package:ubuntu_wizard/widgets.dart';
@@ -9,12 +10,12 @@ import 'wizard.dart';
 class UbuntuWslSetupApp extends StatelessWidget {
   const UbuntuWslSetupApp({
     Key? key,
+    this.variant,
     this.initialRoute,
-    this.reconfigure = false,
   }) : super(key: key);
 
+  final Variant? variant;
   final String? initialRoute;
-  final bool reconfigure;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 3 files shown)


packages/ubuntu_wsl_setup/lib/app.dart
```diff
@@ -31,9 +32,20 @@ class UbuntuWslSetupApp extends StatelessWidget {
...
   }
+
+  Widget buildWizard(BuildContext context) {
+    switch (variant) {
+      case Variant.WSL_SETUP:
+        return UbuntuWslSetupWizard(initialRoute: initialRoute);
+      case Variant.WSL_CONFIGURATION:
+        return UbuntuWslReconfigureWizard(initialRoute: initialRoute);
+      case null:
+        return const SizedBox.shrink();
```
jpnurmi, line 46: For the time being, the window is empty during the client variant query. If we want a "busy wait" screen, we could inject it here.

Here's a plain centered busy indicator with an artificial few seconds client variant query delay for testing.

https://user-images.githubusercontent.com/140617/140540281-fc5bc5bf-d6d3-48d1-a59d-8895a3abdc91.mp4

didrocks: looks great as a placeholder :) Indeed, we can hook that up later on

packages/ubuntu_wsl_setup/lib/main.dart
```diff
@@ -17,23 +18,31 @@ screens, yet allowing user to overwrite any of those during setup.
...
   runWizardApp(
-    UbuntuWslSetupApp(
-      initialRoute: options['initial-route'],
-      reconfigure: options['reconfigure'] == true,
+    ValueListenableBuilder<Variant?>(
+      valueListenable: variant,
+      builder: (context, value, child) {
+        return UbuntuWslSetupApp(
+          variant: value,
+          initialRoute: options['initial-route'],
```
didrocks, line 32: The route name shouldn’t match the mode? (not a blocker for this PR, but just a thought)

didrocks (approved):
Thanks! Just one small question which is more a general one on routing. Otherwise, this looks good to me.

packages/ubuntu_wsl_setup/lib/main.dart
jpnurmi: The --initial-route argument is not used for selecting the mode. The wizard UI is built based on the above variant value. The --initial-route argument allows jumping to a specific page regardless of the mode. It's mostly useful for integration testing (to avoid repeatedly navigating through all pages in every test case).

jpnurmi: For example, this would jump straight to the profile setup:
```
ubuntu-wsl-setup --initial-route="/profile-setup"
```
It's a hidden argument and not useful to users since they don't know the route names that are an internal implementation detail. However, in integration tests, it's useful when there are several different page-specific test cases. For example, if we had multiple test cases specifically for the profile setup, this argument would make it possible to avoid navigating through the language selection page in every profile setup-specific test case.

didrocks: ah, ack, it’s not the name of the route itself but the entry point of the current route, got it! :)

PR #479, opened 2021-11-08
Don't limit the list of architecture for the snap build

seb128:
The selection should be done on the builder instead


kenvandine (approved):
Looks good

PR #480, opened 2021-11-08
Prevent closing the window during installation

### LINKED ISSUE #455: Prevent closing the window during installation
The window close button in the header bar should be removed to avoid closing (and re-opening) the installer while subiquity is busy applying changes. (Ubiquity removes the close button too.)

Originally reported by @guiverc at https://discourse.ubuntu.com/t/new-desktop-installer-preview-build/24765/14

> I did however EXIT the installer after selecting ERASE DISK and INSTALL Ubuntu (just exploring looking for the weird behavior, and as erase disk isn’t what I wanted [I’m a user; we do strange things, change our minds etc]), where I did what I suspect most users would do… hit to RED CROSS top right to close the program.
>
> On re-starting the installer here, (tried twice), and on both I got
> [...]
> my real partitions are not shown !!
>
> If I load gparted it instead shows
> [...]
> gparted did give an error “Libparted Error - Invalid partition table on /dev/sda - wrong signature on it”
>
> That error maybe new; OR the result of last failed install las time; so I’m not worried about it… My intention was to just install into sda3 with format thus fixing any potential issues in that partition.
>
> sda1 & sda2 should be untouched

jpnurmi:
Hide the close button and ignore the delete-event unless the status is
"done" or "error".

Fixes #455

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
@@ -52,6 +52,7 @@ class InstallationSlidesModel extends ChangeNotifier with SystemShutdown {
   void _updateStatus(ApplicationStatus? status) {
     if (state == status?.state) return;
     _status = status;
+    setWindowClosable(isDone || hasError);
     notifyListeners();
   }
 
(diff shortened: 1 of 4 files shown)


robert-ancell (approved):
LGTM!

Issue #452, opened 2021-10-28
Shouldn't hardcode Ubuntu in strings


seb128:
Checking the translations template there are a bunch of strings that mention 'Ubuntu', examples

- Use LVM with the new Ubuntu installation
- Encrypt the new Ubuntu installation for security
- Erase disk and install Ubuntu

Ubiquity has a variable for those, it's not important for now but it will useful as flavor start using the new installer

PR #478, opened 2021-11-08
Add basic theming & asset lookup capabilities for flavors

jpnurmi:
## Creating a flavor

- Create a new Flutter project
- Create symlinks (#477):
  - `<app>/subiquity` -> `/path/to/subiquity`
  - `<app>/test/socket` -> `/path/to/subiquity/.subiquity/socket`
- Add `ubuntu_desktop_installer` dependency (path) to `pubspec.yaml`
- Call `runInstallerApp()` and pass `FlavorData`

## Theming

```dart
import 'package:flutter/material.dart';
import 'package:ubuntu_desktop_installer/installer.dart';

void main(List<String> args) {
  runInstallerApp(
    args,
    flavor: FlavorData(
      name: 'Ubuntu X',
      theme: ThemeData(primarySwatch: Colors.purple),
    ),
  );
}
```

## Assets

Images can be replaced by placing them into `<app>/assets`. The default
assets in `ubuntu_desktop_installer` are used as fallbacks for missing
assets.

```yaml
flutter:
  assets:
    - assets/hard-drive.png
    - assets/repair-wrench.png
    - assets/steering-wheel.png
```

## Preview

Screenshot from 2021-11-08 09-24-10

P.S. By the time of taking the screenshot, the flavor name is not yet used in translations.

PR #485, opened 2021-11-09
Parameterize 'Ubuntu' in translation strings

jpnurmi:
The parameter values are still hard-coded as 'Ubuntu' until the flavor
name available (#478).

The parameters are all capitalized to match Ubiquity's translation
strings. This way, Weblate is able to automatically suggest matching
translations.

Fixes: #452


jpnurmi:
I split the change into two commits. The first commit is the actual change and worth reviewing. The second one is the result of `flutter gen-l10n`.

packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
```diff
@@ -1,34 +1,59 @@
...
+    }
+  },
   "releaseNotesLabel": "You may wish to read the <a href=\"{url}\">release notes</a>.",
   "@releaseNotesLabel": {
     "type": "text",
     "placeholders": {
       "url": {}
     }
   },
-
```
jpnurmi, line 31: the Flutter Intl extension for VS Code wants to remove these line-breaks, and Weblate doesn't retain them either, so...

jpnurmi:
I tested with a local Weblate instance that parameter matching is case-sensitive.
| `{RELEASE}` (ARB) vs. `${RELEASE}` (PO) | `{release}` (ARB) vs. `${RELEASE}` (PO) |
|---|---|
| image | image |

oSoMoN:
Why are some of the parameters named `RELEASE` while others are named `DISTRO`? Are we expecting that some will include a version string or a codename?
As far as I can tell, for the Ubuntu desktop installer, their value is always "Ubuntu".

jpnurmi:
I was wondering the same. :) The translation strings are matched with Ubiquity: https://git.launchpad.net/ubiquity/tree/debian/real-po/fr.po

oSoMoN:
Got it, that's coming from Ubiquity. This sounds like historical inconsistency, rather than an intentional distinction.
I suppose fixing this inconsistency now would make the translations import more difficult, but we can make a note to fix it later?

jpnurmi:
Good idea, I created https://github.com/canonical/ubuntu-desktop-installer/issues/486

oSoMoN:
LGTM, thanks!

PR #490, opened 2021-11-10
Remove extra space in Burmese language name

jpnurmi:
Removing the extra space fixes the sorting order of the language names.
According to Ubiquity, Burmese is not expected to be the first language
but somewhere at the bottom.

There is an extra 0x20 in Ubiquity's languagelist where the language
name was imported from:

https://git.launchpad.net/ubiquity/tree/d-i/source/localechooser/languagelist#n75

### DIFF
--- a/packages/ubuntu_wizard/lib/src/l10n/ubuntu_localizations_my.dart
+++ b/packages/ubuntu_wizard/lib/src/l10n/ubuntu_localizations_my.dart
@@ -11,7 +11,7 @@ class UbuntuLocalizationsMy extends UbuntuLocalizations {
   String get countryCode => 'MM';
 
   @override
-  String get languageName => ' ﻿မြန်မာစာ';
+  String get languageName => '﻿မြန်မာစာ';
 
   @override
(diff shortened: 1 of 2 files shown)


jpnurmi:
| Before | After | Ubiquity |
|---|---|---|
| Screenshot from 2021-11-10 15-22-42 | Screenshot from 2021-11-10 15-18-17 | image |

jpnurmi:
image

jpnurmi:
> fixes the sorting order of the language names

Well, to some degree, at least. The order is still not identical to Ubiquity, but that's a different issue. I don't even know which one is correct and perhaps that's not even important. :)

With this, all the "exotic" language names end up at the bottom. Burmese being at the top caught my attention.

seb128 (approved):
Thanks, that's a nice catch. I wonder if ubiquity strip spaces to get the right sorting, not that it matters for the fix, removing the space is right and it seems Debian did the same in https://salsa.debian.org/installer-team/localechooser/-/commit/c443fdb88

PR #500, opened 2021-11-12
Tests: add shared tester.buildApp() to reduce boilerplate

jpnurmi:
A similar buildApp() helper was in most page tests. Since setting up
the wizard and its routes is an irrelevant detail for widget tests,
share the logic in one place. Also, when FlavorData will be taken into
use, it doesn't need to be sprinkled all around the tests.


oSoMoN:
LGTM, removing a lot of duplicated code is always satisfactory :)

PR #508, opened 2021-11-16
Correct installer page routes and conditions

jpnurmi:


jpnurmi:
Adapted from https://miro.com/app/board/o9J_klzAnOw=/

image

jpnurmi:
P.S. There will be an integration test for navigating through all the possible routes: https://github.com/canonical/ubuntu-desktop-installer/pull/283

packages/ubuntu_desktop_installer/lib/installer.dart
```diff
@@ -200,17 +194,34 @@ class _UbuntuDesktopInstallerWizardState
         ),
         Routes.selectGuidedStorage: WizardRoute(
           builder: SelectGuidedStoragePage.create,
-          onNext: (settings) => Routes.writeChangesToDisk,
+          onNext: (_) =>
+              (!model.hasEncryption) ? Routes.writeChangesToDisk : null,
```
oSoMoN, line 198: Minor style nit: other such conditionals in this file do not enclose the negated attribute in parentheses.

oSoMoN (approved):
Just a minor style nit, otherwise LGTM, feel free to merge.

packages/ubuntu_desktop_installer/lib/installer.dart
jpnurmi: Good spot, thanks. One more character and it would fit on one line. :) I wanted to have it negated to be able to organize the routes so that the conditions can be "if something, skip to route" and it jumps over any routes in between. The null fallback lets it continue to the next route regardless of what it is.

oSoMoN: Yes, that looks good, and I find it quite readable.

PR #509, opened 2021-11-16
Add Routes.initial for convenience

jpnurmi:
Even though there's a --initial-route command-line argument, it's often
easier and faster to edit the initial route on the fly and hot restart
the app, especially if the app is already running anyway.

Definiting the initial route in routes.dart makes it quick to find and
edit the initial route, rather than searching for it in installer.dart.


oSoMoN:
Good idea!

Issue #510, opened 2021-11-16
Compilation errors in test/ dirs


jpnurmi:
There's an issue that e6693ec5ae3eb39e0409a6e46180242bff85f481 missed. It's not possible to disable _only_ linters for specific directories: https://github.com/dart-lang/linter/issues/320. Excluding the `test/` directory in `analysis_options.yaml` results in compilation errors from the analyzer not showing up either, until the test is run.

![image](https://user-images.githubusercontent.com/140617/141970929-ab59af85-76d3-4a28-bc33-1f2fe239bab9.png)

PR #511, opened 2021-11-16
Partial revert of e6693ec5 to restore analyzer for tests/

jpnurmi:
Consequently, the `prefer_const_constructors` linter rule has been
temporarily commented out because there would be a lot of warnings
in `tests/`.

Close: #510


jpnurmi:
image

PR #515, opened 2021-11-16
Update translation instructions

jpnurmi:


jpnurmi:
To be merged after ubuntu-wizard has been replaced by ubuntu-localizations in Weblate.

seb128 (changes requested):
Thanks, as discussed by chat we are going to use a new project for those so the description is going to be need to be updated with the correct naming before landing

jpnurmi:
The link is correct after all because we decided to keep ubuntu-localizations in the installer translation project.

seb128 (approved):
Right!

Issue #512, opened 2021-11-16
Context menu language is wrong in the daily canary image


jpnurmi:
When navigating through the installer without changing the default English language, the log viewer's context menu in the installation slides screen has a wrong language in the daily canary (20211116). In a development environment, the language is correct.

**Live run with the daily canary (20211116):**
![image](https://user-images.githubusercontent.com/140617/141994034-8f9279ce-1d0a-4ce2-a26b-012211fdaad1.png)

vs.

**Dry run in a development environment:**
![image](https://user-images.githubusercontent.com/140617/141994314-de79f2af-7dd1-4ddb-afdd-bcdc6c903cc4.png)

jpnurmi:
FWIW, the problem is reproducible with such a minimal Flutter app:
```dart
import 'package:flutter/material.dart';
import 'package:flutter_localizations/flutter_localizations.dart';

void main() {
  runApp(
    MaterialApp(
      locale: const Locale('C.UTF-8'),
      localizationsDelegates: const [GlobalMaterialLocalizations.delegate],
      supportedLocales: const [Locale('am'), Locale('en')],
      home: Scaffold(
        body: Builder(builder: (context) {
          final localizations = MaterialLocalizations.of(context);
          return Column(
            children: [
              Text('Cut: ${localizations.cutButtonLabel}'),
              Text('Copy: ${localizations.copyButtonLabel}'),
              Text('Paste: ${localizations.pasteButtonLabel}'),
            ],
          );
        }),
      ),
    ),
  );
}
```
![Screenshot from 2021-11-16 17-16-23](https://user-images.githubusercontent.com/140617/142023297-26a616f5-c251-4b17-b670-bbff10daeb79.png)

jpnurmi:
[WidgetsApp.supportedLocales](https://api.flutter.dev/flutter/widgets/WidgetsApp/supportedLocales.html):

> This list must not null. Its default value is just [const Locale('en', 'US')].
>
> The order of the list matters. The default locale resolution algorithm, basicLocaleListResolution, attempts to match by the following priority:
>
> 1. [Locale.languageCode], [Locale.scriptCode], and [Locale.countryCode]
> 2. [Locale.languageCode] and [Locale.scriptCode] only
> 3. [Locale.languageCode] and [Locale.countryCode] only
> 4. [Locale.languageCode] only
> 5. [Locale.countryCode] only when all preferred locales fail to match
> 6. Returns the first element of [supportedLocales] as a fallback

PR #516, opened 2021-11-16
Make sure 'en' is listed as the first supported locale

jpnurmi:
This fixes the issue that Material picks a wrong language when the
system locale (C.UTF-8) cannot be found on the language list. The first
list entry is used as a fallback, which in our case was 'am'.

Close: #512


oSoMoN (approved):
There's a trivial merge conflict to resolve, but otherwise LGTM, and fixes the issue.

PR #523, opened 2021-11-17
snapcraft: bundle util-linux

### LINKED ISSUE #513: Curtin traceback in latest jammy canary image
[subiquity-server-debug.log](https://github.com/canonical/ubuntu-desktop-installer/files/7547490/subiquity-server-debug.log)

dbungert:
Curtin expects a certain output format from util-linux, which seems to
have changed in the util-linux in Jammy.  Bundle util-linux so that we
have a predictable tool version.  Also adjust PATH to find the tools in
the snap before the system ones.

Fixes #513

### DIFF
--- a/snap/local/subiquity-server
+++ b/snap/local/subiquity-server
@@ -15,7 +15,7 @@ export PYTHON=$SNAP/usr/bin/python3.8
 export PY3OR2_PYTHON=$PYTHON
 
 # set the PATH so subiquity finds curtin
-export PATH=$PATH:$SNAP/bin
+export PATH=$SNAP/bin:$PATH
 
 # base directory for subiquity to locate resources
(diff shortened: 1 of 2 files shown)


dbungert:
@seb128 - we have a design decision to make.  We could merge this and bundle util-linux in the snap, but you observed that maybe this and udev should match.  Alternately I've got a fix in flight on the curtin side: https://code.launchpad.net/~dbungert/curtin/+git/curtin/+merge/412027

seb128 (approved):
Curtin is going to need to be fixed anyway but let's merge that one at least for now to get to a working canary daily, we will need to investigate the compatibility question, I will open a bug about it so we don't forget to revisit once the curtin fix has landed

PR #524, opened 2021-11-18
Support partially overridden flavor image assets

jpnurmi:
Before, a flavored installer had to bundle a full set of image assets.
This change makes it possible to override only specific assets and let
the app use the default assets for others.


jpnurmi:
A dumb example that demonstrates how overriding a subset of image assets works. When there will be more image assets, flavors can choose which ones they want to override.

| Before | After |
|---|---|
| Screenshot from 2021-11-18 11-11-22 | Screenshot from 2021-11-18 11-10-29 |

oSoMoN:
The OptionCard's title isn't visible any longer, as evidenced by your before/after screenshot (and I'm seeing this when testing locally, too).

jpnurmi:
| Light | Dark | Flavor |
|---|---|---|
| Screenshot from 2021-12-15 08-27-29 | Screenshot from 2021-12-15 08-27-22 | Screenshot from 2021-12-15 08-28-29 |

jpnurmi:
How could I not notice that...

oSoMoN (approved):
LGTM now, thanks!

PR #528, opened 2021-11-19
Remove --no-fatal-infos for analysis

jpnurmi:
There used to be a lot of false positives due to `public_member_api_docs`. Now that it has been turned off, we can remove `--not-fatal-infos` to avoid other analysis issues slipping in.

Ref: #467

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -13,7 +13,7 @@ scripts:
   # analyze all packages
   analyze: >
     melos exec -c 1 -- \
-      flutter analyze --no-fatal-infos .
+      flutter analyze .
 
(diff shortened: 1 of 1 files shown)


didrocks:
Ack on this, we can’t clean all infos or are there too many false positives?

jpnurmi:
There used to be a lot of false positives due to `public_member_api_docs`. Now that it has been turned off, we can remove `--not-fatal-infos` to avoid other analysis issues slipping in.

didrocks:
Ack then, approving :)

didrocks (approved):

PR #534, opened 2021-11-25
Update subiquity

jpnurmi:


seb128 (approved):
Great, that should give us a working installation again!

PR #535, opened 2021-11-25
subiquity_client: add free_only API

jpnurmi:
For the "Install third-party software" selection on the "Updates and
other software" page (#31).


robert-ancell (approved):
Don't know what these method are, but this seems like appropriate code to access them!

PR #536, opened 2021-11-25
Install third-party software

jpnurmi:
Restore the checkbox on the "Updates and other software" page (#31),
and wire it up with the Subiquity "free only" API.


jpnurmi:
image

oSoMoN (approved):
LGTM. Note that the "install third-party software" checkbox is checked by default, because subiquity returns `false` for the `free_only` endpoint by default, which seems to be intentional, but this differs from the design in #31.

seb128:
Third party drivers and codecs shouldn't be enabled by default since it's the safest default. Is the free-only-ui API in subiquity also disabling restricted and multiverse as in ubiquity or only dealing with drivers and codecs?

jpnurmi:
Closing in favor of https://github.com/canonical/ubuntu-desktop-installer/pull/733

PR #537, opened 2021-11-25
subiquity_client: hide logger object instances

jpnurmi:
Don't leak subiquity_client's logger instance to prevent accidental use of a wrong logger instance in the installer. The issue came up in https://github.com/canonical/ubuntu-desktop-installer/pull/442#discussion_r732729315.


packages/subiquity_client/lib/src/client.dart
```diff
@@ -0,0 +1,468 @@
+import 'dart:async';
+import 'dart:convert';
+import 'package:dartx/dartx.dart';
+import 'package:http/http.dart';
+import 'package:logger/logger.dart';
+import 'http_unix_client.dart';
+import 'types.dart';
+
+/// @internal
+final log = Logger('subiquity_client');
```
robert-ancell, line 10: Wouldn't it have been easier to rename this to `_log`?

jpnurmi: It would be much more straightforward, but I went through the trouble of moving things around just to be able to keep the same pretty `log` name that we're using everywhere. :) There are other places in the project where the log instance is intentionally exposed to make it available within the same module (e.g. all models in page/connect_to_internet use the same instance from the main model).

Not that it really matters for an internal unpublished library like this, but in general, it's also considered good practice (or a matter of taste) to have the library implementation in `src/foo.dart` to have full control over the visibility of exported symbols in `lib/foo.dart`.

robert-ancell (approved):

Issue #41, opened 2021-03-08
Screen 15 - Configure Active Directory


seb128:
The design for screen n°15, Configure Active Directory

![activedir](https://user-images.githubusercontent.com/10974330/110314853-b3d35f80-8008-11eb-93aa-f735f99b75e4.png)

Design details https://people.canonical.com/~platform/desktop/installer/15.%20Configure%20Active%20Directory/

nafallo:
Any chance to support lightweight SSO, like utilising libpam-krb5?

Mannshoch:
May allow use e.g. a sftp server. So If someone uses username and password which works on a certain server (may a family NAS) a user is allowed to login.

seb128:
subiquity is going to need to help with backend work there right? ubiquity is using pam-auth-update on the target env in addition of using realm

jibel:
Since the password input is hidden, a password confirmation field must be added to ensure that the password is correct.

N0rbert:
Really needed.

jpnurmi:
Which indentation looks best?

### Option 1

![image](https://user-images.githubusercontent.com/140617/220862287-f9ddae18-f4ea-4381-b78c-ade9f2bf37a5.png)

### Option 2

![image](https://user-images.githubusercontent.com/140617/220862392-72a1f81a-4f63-49ef-b195-9f2c27956e61.png)

### Option 3

![image](https://user-images.githubusercontent.com/140617/220862497-8ff1ebfc-6148-4733-94d3-4079b34194bd.png)

### Option 4

![image](https://user-images.githubusercontent.com/140617/220862558-93fba301-ab8b-4d38-8db2-fcf084fc82eb.png)

### Ubiquity for reference

![Screenshot from 2023-02-22 11-44-01](https://user-images.githubusercontent.com/140617/220862193-bac56448-e8db-4759-b9b3-e8adafc61015.png)

/cc @elioqoshi

elioqoshi:
Definitely Option 2, Indentation always the same for form elements, description for the same element starting at the same position as its title.

jpnurmi:
![image](https://user-images.githubusercontent.com/140617/222197122-20647eca-0ac5-4aaa-ac0f-800a1e9b75b0.png)

elioqoshi:
Context: We should offer an error message explaining the error in case something goes wrong.

elioqoshi:
@jibel Do you have the title of the error message too?

[...]

PR #538, opened 2021-11-26
Configure Active Directory

jpnurmi:
The actual AD configuration functionality is missing and will be
implemented separately.

Also, the page route is skipped by default. For trying out the page,
pass --initial-route or edit routes.dart.

    flutter run -a "--initial-route=/configureactivedirectory"

Ref: #41


jpnurmi:
Screenshot from 2021-11-26 09-35-55
Screenshot from 2021-11-26 09-36-05
Screenshot from 2021-11-26 09-36-15

jibel:
I added a comment on #41 to add a password confirmation field. The 2 fields must match in order to all AD join.

jpnurmi:
Depends on https://github.com/canonical/ubuntu-desktop-installer/pull/541

Feichtmeier:
A little input for the test connection button:

Does is also test the credentials of the ad admin?
I've recently had the chance to use the 21.10 version of this feature at work (really impressive btw, worked really well). But currently the test connection button in subiquity felt a bit hard to understand. It tested the domain name and it was correct. However the admin credentials were wrong and so I was allowed to proceed with wrong credentials. I would recommend to do two things: move the button to the bottom of the form and make it require all three fields to be filled correctly :)

jpnurmi:
Maybe the _Continue_ button could say _Connect_ or _Test_ instead, and only change to _Continue_ after a successful connection test? So more or less the same behavior as on the internet page.

Feichtmeier:
> Maybe the "Continue" button could say "Connect" or "Test" instead, and only change to "Continue" after a successful connection test? So more or less the same behavior as on the internet page.

Could also be a solution indeed

My concern with the "Test connection" label on the button  is that it does not test the connection as it says but only the domain name (in ubiquity). 
If it would have the same function as in ubiquity I'd say that the label needs to be "Test domain name" or something like this.

Also the position is (independly of the function of the button) is a bit strange because normally you might want to fill out the whole thing before you test it.
Imo the most intuitive solution would be to move the button to the bottom of the Form

```
[........ Domain name ........]
[...... Admin user name ......]
[....... password ............]

[ Test this connection ]


       [ go back ] [ continue ]
```

jpnurmi:
Let's get back to this when Subiquity has support for AD.

Issue #363, opened 2021-09-28
Welcome page reloads the same keyboard layouts on every language change


jpnurmi:
There is no `SubiquityClient.setLocale()` call between loading keyboard layouts.

From `KeyboardService.load()` docs:
>   /// Reload the list of layouts for the current locale.
>  /// Call this when the locale has changed
>  /// (typically after calling SubiquityClient.setLocale(…)).

Steps to reproduce:
- Cherry-pick #356
- Select languages on the welcome page
- Observe the output

Example output:
```
flutter: INFO welcome: Loaded 8 languages
[...]
flutter: DEBUG subiquity_client: ==> keyboard() {"setting": {"layout": "se", "variant": "", "toggle": null}, "layouts": [{"code": "af", "name": "Afghani", "variants": [...
flutter: INFO welcome: Loaded 99 keyboard layouts
[...]
flutter: INFO welcome: Selected es_ES as UI language
flutter: DEBUG subiquity_client: GET http://localhost/keyboard
flutter: DEBUG subiquity_client: ==> keyboard() {"setting": {"layout": "se", "variant": "", "toggle": null}, "layouts": [{"code": "af", "name": "Afghani", "variants": [...
flutter: INFO welcome: Loaded 99 keyboard layouts
flutter: INFO welcome: Selected fr_FR as UI language
flutter: DEBUG subiquity_client: GET http://localhost/keyboard
flutter: DEBUG subiquity_client: ==> keyboard() {"setting": {"layout": "se", "variant": "", "toggle": null}, "layouts": [{"code": "af", "name": "Afghani", "variants": [...
flutter: INFO welcome: Loaded 99 keyboard layouts
```

The locale is not set until pressing `Continue`:
```
flutter: INFO welcome: Set fr_FR as system locale
flutter: DEBUG subiquity_client: POST http://localhost/locale
flutter: DEBUG subiquity_client: ==> setLocale("fr_FR.UTF-8") null
```

jpnurmi:
Why are we loading keyboard layouts on the welcome page, anyway? Shouldn't the layouts be loaded when entering the keyboard layout page? :thinking:

PR #540, opened 2021-11-26
Remove bogus keyboard layout loading from the welcome screen

jpnurmi:
As documented in KeyboardService.load(), the locale must be passed to
subiquity before loading keyboard layouts. However, when browsing
languages on the welcome page, the locale is not passed to subiquity
until the user presses the Continue button. Therefore, the UI loads
the same en_US keyboard layouts again and again every time the user
changes the UI language.

Loading the keyboard layouts is better done in the keyboard layout
page, where the layouts and variants are actually used. At that point,
the locale has been applied too.

Fixes: #363


oSoMoN:
LGTM, thanks!

oSoMoN (approved):

PR #541, opened 2021-11-26
Share username, password etc. form fields

jpnurmi:
In addition to the existing _Who are you?_ and _WSL Profile setup_ pages, these will be used in the [Configure Active Directory](https://github.com/canonical/ubuntu-desktop-installer/issues/41) too.


packages/ubuntu_wizard/pubspec.yaml
```diff
@@ -33,6 +33,7 @@ dependencies:
       path: packages/ubuntu_logger
   wizard_router: ^0.6.0
   yaru: ^0.2.0
+  yaru_icons: ^0.0.8
```
Feichtmeier, line 36: This could be a opportunity to upgrade to the new 0.1.2 yaru_icons where the path of the widgets changed :)

jpnurmi: Ah yeah, thanks for the reminder! :) https://github.com/canonical/ubuntu-desktop-installer/pull/594

oSoMoN (commented):
LGTM, but I see that `kValidUsernamePattern` and `kValidHostnamePattern` are duplicated, and validation happens independently in several places (in the models where it is used to conditionally enable the Next button, and in the form validators attached to the widgets).
There's an opportunity for further de-duplicating code here.

jpnurmi:
I got rid of all the duplicate validation. I started with `onChanged(String value, bool isValid)` callback but it resulted in a large number of booleans in the models, which is where the `ValidatedValue` wrapper stems from.

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_model.dart
```diff
@@ -32,6 +22,12 @@ enum LoginStrategy {
   autoLogin
 }
 
+// extension _ValidatedStringNotifier on ValueNotifier<ValidatedString?> {
+//   bool? get isValid => value?.isValid;
+//   String? get string => value?.value;
+//   set string(String? value) => value = value;
+// }
+
```
oSoMoN, line 30: If this is unused, should it simply be removed?

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_model.dart
```diff
@@ -94,21 +90,25 @@ class WhoAreYouModel extends ChangeNotifier {
...
 
   /// Loads the identity data from the server, and resolves the system hostname.
   Future<void> loadIdentity() async {
     final identity = await _client.identity();
-    _realName.value = identity.realname?.orIfEmpty(null);
-    _hostname.value = identity.hostname?.orIfEmpty(null);
-    _username.value = identity.username?.orIfEmpty(null);
+    _initIdentity(_realName, identity.realname?.orIfEmpty(null));
+    _initIdentity(_hostname, identity.hostname?.orIfEmpty(null));
+    _initIdentity(_username, identity.username?.orIfEmpty(null));
```
oSoMoN, line 111: Instead of calling `orIfEmpty(null)` here, you could check whether the string is empty in `_initIdentity()`.

packages/ubuntu_desktop_installer/test/configure_secure_boot/configure_secure_boot_page_test.dart
```diff
@@ -64,10 +62,12 @@ void main() {
     expect(find.byType(SuccessIcon), findsNWidgets(2));
 
     await tester.enterText(fields.first, 'ubuntu');
-    verify(model.setSecurityKey('ubuntu')).called(1);
+    verify(model.setSecurityKey(ValidatedString('ubuntu', isValid: true)))
+        .called(1);
 
     await tester.enterText(fields.last, 'ubuntu');
-    verify(model.setConfirmKey('ubuntu')).called(1);
+    verify(model.setConfirmKey(ValidatedString('ubuntu', isValid: false)))
```
oSoMoN, line 69: Why is the expectation that the confirmation key isn't valid here?

packages/ubuntu_desktop_installer/test/who_are_you/who_are_you_page_test.dart
```diff
@@ -93,9 +95,12 @@ void main() {
     final textField = find.widgetWithText(TextField, 'passwd');
     expect(textField, findsNWidgets(2));
     await tester.enterText(textField.first, 'ubuntu');
-    verify(model.password = 'ubuntu').called(1);
+    verify(model.setPassword(ValidatedString('ubuntu', isValid: true)))
+        .called(1);
     await tester.enterText(textField.last, 'ubuntu');
-    verify(model.confirmedPassword = 'ubuntu').called(1);
+    verify(model
+            .setConfirmedPassword(ValidatedString('ubuntu', isValid: false)))
```
oSoMoN, line 102: Same question here, why is the expectation that the confirmed password is invalid?

oSoMoN (commented):
The new changes LGTM, but I have a few questions.

packages/ubuntu_desktop_installer/test/configure_secure_boot/configure_secure_boot_page_test.dart
jpnurmi: It was a mismatch against the mocked value from the model, but very confusing indeed because a matching value was entered above. I've made it clear by entering "mismatch" instead :)

oSoMoN:
I'm seeing a functional regression on the "Who are you?" page of the installer: if I enter "Foo Bar" as a real name, the automatically generated username is "f", whereas on main it is "foo-bar".

jpnurmi:
Too many changes piled up here, some even unrelated. I'll try to split this into smaller steps.

Issue #518, opened 2021-11-16
Low resolution support


jpnurmi:
From https://discourse.ubuntu.com/t/new-desktop-installer-preview-build/24765/38
> Only issue noticed was I’d not setup screen resolution on the VM so it was small (800x600 maybe; end-user mistake) and the installers window was too large so I couldn’t progress forward until I corrected screen resolution.
>
> Maybe a pop-up balloon mentioning screen resolution issue maybe helpful for some users? with link maybe to https://help.ubuntu.com/community/Installation/SystemRequirements 1 which states 1024x768)

This is how 800x600 looks:
| Ubuntu Desktop Installer | Ubiquity |
|---|---|
| ![image](https://user-images.githubusercontent.com/140617/142040962-730eaeb7-549b-4948-9dd9-e60b34eca589.png) | ![image](https://user-images.githubusercontent.com/140617/142041643-157f7d87-d92c-4e4c-a164-c63eeacf6548.png) |

jpnurmi:
See also #253

fitojb:
This is the kind of issues caused by pseudo-“touch-friendly”, super-padded out UI design that is so en vogue right now. I already reported this exact issue years ago to Yaru designers, who were stubborn and rejected it

Feichtmeier:
> This is the kind of issues caused by pseudo-“touch-friendly”, super-padded out UI design that is so en vogue right now. I already reported this exact issue years ago to Yaru designers, who were stubborn and rejected it

No this is not kind of the issue.
This problem can be fixed by building this page with a LayoutBuilder which can digest screen constraints, like for example:

https://github.com/ubuntu/yaru_widgets.dart/blob/9a94d220e98625b9a810c30a629a534d6dcd3f44/lib/src/yaru_master_detail_page.dart#L39-L63

jpnurmi:
From what I've tried earlier, the app remains fairly usable even if the window size is significantly smaller. Earlier, 960x680 was seen as a reasonable minimum size, but I think we could improve the window sizing logic to make it fit into a 800x600 desktop's available workarea while the app would still remains usable.

jpnurmi:
The window is now maximized on app startup if the preferred window size doesn't fit the available workarea. As suggested in https://github.com/canonical/ubuntu-desktop-installer/issues/518#issuecomment-971658672, we should consider improving it further by automatically resizing the window to the preferred size if it fits after changing the resolution.

jpnurmi:
[...]

PR #542, opened 2021-11-29
Limit the window size to the available workarea

jpnurmi:
When the available workarea is smaller than the default window size,
maximize the window to let the window manager take care of filling the
available workarea. This ensures that the window fits on the screen
even if the resolution is relatively low. For example, the default in
Virtual Box is 800x600.

Ref: #518

### DIFF
--- a/packages/ubuntu_desktop_installer/linux/my_application.cc
+++ b/packages/ubuntu_desktop_installer/linux/my_application.cc
@@ -30,8 +30,7 @@ static gboolean my_application_fit_to_workarea(GtkWindow* window) {
   gboolean fits_workarea =
       allocation.width < workarea.width || allocation.height < workarea.height;
   if (!fits_workarea) {
-    gtk_window_move(window, workarea.x, workarea.y);
-    gtk_window_resize(window, workarea.width, workarea.height);
+    gtk_window_maximize(window);
   }
   return fits_workarea;
 }
--- a/packages/ubuntu_desktop_installer/linux/my_application.cc
+++ b/packages/ubuntu_desktop_installer/linux/my_application.cc
@@ -11,6 +11,31 @@ struct _MyApplication {
 
 G_DEFINE_TYPE(MyApplication, my_application, GTK_TYPE_APPLICATION)
 
+static void my_application_get_workarea(GtkWindow* window,
+                                        GdkRectangle* workarea) {
+  GdkWindow* gdk_window = gtk_widget_get_window(GTK_WIDGET(window));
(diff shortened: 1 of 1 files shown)


jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2021-11-29 10-41-31 | Screenshot from 2021-11-29 10-39-48 |

packages/ubuntu_desktop_installer/linux/my_application.cc
```diff
@@ -11,6 +11,31 @@ struct _MyApplication {
...
+  gtk_widget_get_allocation(GTK_WIDGET(window), &allocation);
+
+  GdkRectangle workarea;
+  my_application_get_workarea(window, &workarea);
+
+  gboolean fits_workarea =
+      allocation.width < workarea.width || allocation.height < workarea.height;
+  if (!fits_workarea) {
+    gtk_window_move(window, workarea.x, workarea.y);
+    gtk_window_resize(window, workarea.width, workarea.height);
```
robert-ancell, line 34: Perhaps instead of explicitly moving and resizing the window could be maximized instead? That way the window manager is responsible for correct placement.

packages/ubuntu_desktop_installer/linux/my_application.cc
```diff
@@ -34,11 +59,16 @@ static void my_application_activate(GApplication* application) {
   gtk_header_bar_set_decoration_layout(header_bar, ":close");
   gtk_window_set_titlebar(window, GTK_WIDGET(header_bar));
   gtk_window_set_default_size(window, 960, 680);
+  gtk_widget_realize(GTK_WIDGET(window));
+
+  if (my_application_fit_to_workarea(window)) {
```
robert-ancell, line 64: If the resolution is changed later, then this may no longer apply. Not sure if this is applicable in the installer use case. I know when I use VMs the resolution normally defaults to too small, and then I immediately resize it.

robert-ancell (approved):
Changes look good, just left a few suggestions that may or may not be useful.

packages/ubuntu_desktop_installer/linux/my_application.cc
jpnurmi: Thanks! Much better :)

packages/ubuntu_desktop_installer/linux/my_application.cc
jpnurmi: Thanks, this is a great idea. I left #518 open and added a link to this comment.

PR #530, opened 2021-11-23
WSL setup: add ARB stubs

jpnurmi:
This makes all supported languages available on the language list page and in Weblate.

PR #554, opened 2021-12-07
WSL: fix language selection in the integration test

jpnurmi:
The test was assuming that the "French" list item was visible on the
screen by default. However, #530 added a long list of supported
languages, which pushed the "French" item out of the screen when the
default "English" item is selected on startup. Consequently, the test
would not be able to tap the "French" item. This commit fixes the issue
by scrolling the list until the "French" item becomes visible, before
trying to tap it.

### DIFF
--- a/packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
+++ b/packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
@@ -111,7 +111,8 @@ Future<void> testSelectYourLanguagePage(
   if (language != null) {
     final tile = find.widgetWithText(ListTile, language, skipOffstage: false);
     expect(tile, findsOneWidget);
-    await tester.scrollUntilVisible(tile, kMinInteractiveDimension);
+    final viewRect = tester.getRect(find.byType(ListView));
+    await tester.scrollUntilVisible(tile, viewRect.height / 2);
     await tester.pump();
     await tester.tap(tile);
     await tester.pump();
--- a/packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
+++ b/packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
@@ -109,7 +109,11 @@ Future<void> testSelectYourLanguagePage(
   expectPage(tester, SelectLanguagePage, (lang) => lang.selectLanguageTitle);
 
   if (language != null) {
-    await tester.tap(find.widgetWithText(ListTile, language));
+    final tile = find.widgetWithText(ListTile, language, skipOffstage: false);
+    expect(tile, findsOneWidget);
+    await tester.scrollUntilVisible(tile, kMinInteractiveDimension);
+    await tester.pump();
+    await tester.tap(tile);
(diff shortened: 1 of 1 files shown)


packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
```diff
@@ -109,7 +109,11 @@ Future<void> testSelectYourLanguagePage(
   expectPage(tester, SelectLanguagePage, (lang) => lang.selectLanguageTitle);
 
   if (language != null) {
-    await tester.tap(find.widgetWithText(ListTile, language));
+    final tile = find.widgetWithText(ListTile, language, skipOffstage: false);
+    expect(tile, findsOneWidget);
+    await tester.scrollUntilVisible(tile, kMinInteractiveDimension);
```
CarlosNihelton, line 114: Humble question: `kMinInteractiveDimension` is inherited from `Material`? Wondering whether that could be too big or too small.

CarlosNihelton (commented):
Just one sincere question about `kMinInteractiveDimension`. Otherwise, this LGTM.

packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
jpnurmi: Notice that it scrolls repeatedly until the target item becomes visible. This arbitrary number is just the delta for each scroll step, not the total scrolling distance. An arbitrary constant looked nicer than an arbitrary hardcoded value. :grin:

I recall that a dense ListTile has a height of 48 so this would be scrolling max one item at a time, but how would you like scrolling half the view at a time? Would that make it look less arbitrary?

```diff
diff --git a/packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart b/packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
index 0744e81f..36fc79b8 100644
--- a/packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
+++ b/packages/ubuntu_wsl_setup/integration_test/ubuntu_wsl_setup_test.dart
@@ -111,7 +111,8 @@ Future<void> testSelectYourLanguagePage(
   if (language != null) {
     final tile = find.widgetWithText(ListTile, language, skipOffstage: false);
     expect(tile, findsOneWidget);
-    await tester.scrollUntilVisible(tile, kMinInteractiveDimension);
+    final viewRect = tester.getRect(find.byType(ListView));
+    await tester.scrollUntilVisible(tile, viewRect.height / 2);
     await tester.pump();
     await tester.tap(tile);
     await tester.pump();
```

CarlosNihelton: I liked that! Would you mind?

jpnurmi: Alright, let's go with that

CarlosNihelton (approved):
Perfect!

jpnurmi:
Thanks!

PR #560, opened 2021-12-08
Fix WSL setup failing due /sbin not in the PATH

CarlosNihelton:
Hello, @jpnurmi !

Testing the recent images with the installer seeded we failed under WSL environment due useradd, usermod and some locale related utility commands not found in snap PATH. To fix that we need to:
- Add core20/sbin to the PATH.
- Add to the snap two packages related to locale manipulation.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -111,6 +111,9 @@ parts:
       - python3-systemd
       - python3-urwid
       - util-linux
+      # WSL specifics:
+      - language-selector-common
+      - locales
     prime:
       - -lib/systemd/system/*
 
--- a/packages/ubuntu_wsl_setup/ubuntu-wsl-setup
+++ b/packages/ubuntu_wsl_setup/ubuntu-wsl-setup
@@ -571,7 +571,8 @@ if [ -n "$SNAP_DESKTOP_DEBUG" ]; then
   echo "Now running: exec $*"
 fi
 
-PATH="${SNAP}/usr/bin:${SNAP}/bin:${SNAP_CORE_DIR}/bin:${SNAP_CORE_DIR}/usr/bin"
+PATH="${SNAP}/usr/sbin:${SNAP}/usr/bin:${SNAP}/sbin:${SNAP}/bin"
+PATH="${PATH}:${SNAP_CORE_DIR}/usr/sbin:${SNAP_CORE_DIR}/usr/bin:${SNAP_CORE_DIR}/sbin:${SNAP_CORE_DIR}/bin"
 
 # configure python environment
 export PYTHONIOENCODING=utf-8


snap/snapcraft.yaml
```diff
@@ -111,6 +111,9 @@ parts:
       - python3-systemd
       - python3-urwid
       - util-linux
+      # WSL specifics:
+      - language-selector-common
+      - locales
```
jpnurmi, line 116: @seb128 This increases the installer snap size by ~5mb. Is that ok?

Before:
`-rw-r--r--  1 jpnurmi jpnurmi 119783424 Dec 13 10:26 ubuntu-desktop-installer_0+git.f06845df_amd64.snap`

After:
`-rw-r--r--  1 jpnurmi jpnurmi 124923904 Dec 13 10:35 ubuntu-desktop-installer_0+git.ec20911d_amd64.snap`

seb128: Hum, the increase should not be an issue if that's for a valid reason, I'm not convinced we should bundle the locales though. Why aren't those part of the system? It feels wrong having the installer provide a copy of those

CarlosNihelton: Discussing this topic with Didier last friday, he explained to me that binary compatibility problems due system libc not being the same version in the snap and that's why we have adopted having just binaries from the snaps in the PATH in the first place.

seb128: But the locale has to be generated on the target system no? Could you give an example of cmd that were called and not producting the expected result?

CarlosNihelton: In WSL Subiquity Configure controller performs the system installation steps required after rootfs extraction. That means creating users, groups and configuring locales. Cloud-init is not available in that scenario.

Currently we are breaking when attempting to run `useradd`, because `${SNAP_CORE}/usr/sbin` is not on the PATH. Solving that would create another problem because `useradd` links against `libsemanage.so.1` on the snap while in the target system (Jammy) it is linked against `libsemanage.so.2`. The same with `userdmod`. I said "would" because `PATH` and `LD_LIBRARY_PATH` are overriden to point only inside the snaps, thus we'll not face ldd issues at that moment.

So the next crash will be attempting to run the `update-locale` command, because it is not in the PATH. Core20 doesn't ship it. That script is located at `/usr/sbin` and it calls `localedef`, which is a binary located at `/usr/bin`. Adding those to PATH will allow running commands from the target system when we should be running from the snaps and we would go back to the conflicts described above. Besides, Jammy ships glibc version 2.34 and the snap ships 2.31. It doesn't seem to cause conflicts now, but at some point they may diverge and we would have mismatches on libc.

Now, running the locale commands from the snap would create the expected effects on the target system? `localedef` writes to `/usr/lib/locale/`. Being an unconfined snap, and in the case of WSL it is not even an actual snap, it seems to me that it would work as expected. Subiquity's behaviour described above it's only valid under WSL. I wonder how the other components of the desktop installer work with the locales under the snap environment.

CarlosNihelton: Forcing the shell environment to be equals to the desktop installer with `. <(xargs -0 bash -c 'printf "export %q\n" "$@"' -- < /proc/$INSTALLER_PID/environ)` and invoking the locale commands with their full paths didn't fail under a Jammy cloudimage rootfs fresh install and produced the expected effects on the target system. That would fail ofc if system and snap libc mismatched. I couldn't manage to get a snap with the package locales added to it and see if it would be any different from what I just did.

seb128: It's a bit of a tricky topic. If we start staging base utilities we are on our way to turn the snap into a copy of the target system in the container. While it makes sense to bundle components we rely on where we want to lock the version, any ubuntu system should have a working useradd and it's probably fine to rely on that assumption. Could we clean the PATH and LD_LIBRARY_PATH from snap directories to ensure we just use the classic system useradd without ending up loading some wrong libraries like the libsemanage from the snap env?

seb128 (approved):
Let's merge for now to unblock the situation since we can't probably have a meeting to discuss the details with enough people before the holidays. It still feels like wrong to bundle standard utilities and that's something we should try to go back to using the system utilities for such cases

Issue #38, opened 2021-03-08
Screen 13 - Where Are You


seb128:
The design for screen n°13, Where Are You

![where](https://user-images.githubusercontent.com/10974330/110314520-3d366200-8008-11eb-8d78-63a65d1bceba.png)

Design details https://people.canonical.com/~platform/desktop/installer/13.%20Where%20Are%20You/

BillDietrich:
I am an English-speaking US citizen living in Spain.  So "where I am" is not the same as "how I want to see information".  Maybe the installer could handle this more clearly.  "Where I am" could set initial values, but then have a way to override language, date/time format, currency.

madsrh:
This maps has been used since 8.04 and could IMHO really used a visual update to a more modern look. Just changing the colors would be great. Any change that design would have time to look at this?

I would do it myself, but from the ubiquity package I can see that the map is a PNG, so not easily editable.
https://git.launchpad.net/ubuntu/+source/ubiquity/tree/pixmaps/timezone

madsrh:
@oSoMoN / @ long-chung would you consider rethink this screen? Perhaps just a simple search field with autosearching each typed letter and displaying the results in a popup with the flag of the country 🇬🇧  

Or is the plan is to reimplement the design from ubiquity? If so, here's some feedback from team Yaru:

- The colors look very outdated (and aren't taken from the Yaru color palette)
- The current map png is only 400 x 800px (looks blurry on hi-res screens)
- The pin icon is skeuomorphism and should use the pin icon from the Yaru icon set
- For accessibility we need to provide a high-contrast map for visually impaired
- The pointer often doesn't pick the correct location when clicked

I believe that @muqtxdir is already working on a updated version of the map that fixes a couple of the points mentioned above.

long-chung:
@madsrh Totally agree. The map is outdated but was never in scope to redesign. If there is another option being worked on by another team i'm more than happy to relook at this.

Feichtmeier:
@oSoMoN I'd like to work on this :)

Feichtmeier:
@long-chung @madsrh 

I really like the minimal vibe here, WDYT? padding could be improved I guess

(click to enlarge)
![Peek 2021-07-12 16-07](https://user-images.githubusercontent.com/15329494/125301536-5db25800-e32b-11eb-8ec0-a4cf46192aa2.gif)

madsrh:
[...]

PR #573, opened 2021-12-14
Where are you?

jpnurmi:
The map in the original design is not accessible. Therefore, the initial version of the page comes without a map.

https://user-images.githubusercontent.com/140617/148355112-da204953-3a4a-4a25-a0b2-6731648b8a63.mp4

Close: #38


Feichtmeier:
How about a second corresponding text widget that displays the timezone that your location belongs to?

Peek 2021-07-12 16-07

jpnurmi:
@elioqoshi do you have any thoughts on @Feichtmeier's proposal above?

elioqoshi:
@Feichtmeier 's addition is a good proposal! I do feel the whole page for just a dropdown feels a little empty. In my initial explorations I had a flow where one can choose the timezone via a dropdown AND the map (if you'd choose it in one, the other one would be updated accordingly). Do you think this is feasible?

jpnurmi:
I have added a timezone entry and updated the screencast in the PR description. If we decide to add a map, I'd like to leave that as a separate (big) task. :smile:

oSoMoN:
Quick smoke tests: my location is auto-detected to Madrid (Spain), which isn't very accurate, but IIRC ubiquity doesn't do a better job, so that's not a regression. The timezone (Europe/Madrid) is correct.

Searching in the location text field isn't working as I'd expect: when I input "barcelon", I would expect "Barcelona (Catalonia, Spain)" to be in the list of results, but it isn't:

    flutter: DEBUG geo_service: Search "barcelon" matches 1 locations
    flutter: DEBUG geo_service: Search "barcelona" matches 17 locations

And a dummy question, because I haven't read the code yet: can the names of cities and timezones be localized in the selected language, or is the data only available in English?

jpnurmi:
> my location is auto-detected to Madrid (Spain), which isn't very accurate

The result is from https://geoip.ubuntu.com/lookup. I get sometimes even slightly different results when refreshing the page.

> Searching in the location text field isn't working as I'd expect: when I input "barcelon", I would expect "Barcelona (Catalonia, Spain)" to be in the list of results, but it isn't:

I got also surprised that geoname-lookup.ubuntu.com doesn't do a "starts with" search.
- https://geoname-lookup.ubuntu.com/?query=barcelon
vs.
- https://geoname-lookup.ubuntu.com/?query=barcelona

jpnurmi:
> can the names of cities and timezones be localized in the selected language, or is the data only available in English?

Non-English alternate names are included in both online and offline lookups:
- https://geoname-lookup.ubuntu.com/?query=berliini (Berlin in Finnish)
- local offline database with 15000 capital/largest cities (cities15000.txt)

oSoMoN:
> I got also surprised that geoname-lookup.ubuntu.com doesn't do a "starts with" search.
> 
>     * https://geoname-lookup.ubuntu.com/?query=barcelon
>       vs.
> 
>     * https://geoname-lookup.ubuntu.com/?query=barcelona

And that's [a known bug](https://bugs.launchpad.net/ubuntu-geonames/+bug/1045224) which will turn 10 years old later this year, unless we do something about it ;)

oSoMoN:
That's a minor detail, so feel free to ignore, but I find the logging a bit misleading. When I input "barcelona", I see:

    flutter: DEBUG geo_service: Search "barcelona" matches 17 locations

So far so good. Then I click on the first entry in the results, my location is correctly selected, but I see:

    flutter: DEBUG geo_service: Search "Barcelona (Catalonia, Spain)" matches 0 locations

I understand why this happens, but for someone reading the logs it's a tad misleading, maybe we could avoid issuing a request when the user selects an entry in the search results?

jpnurmi:
Yeah, the extra request is a bit annoying.

The problem is that `Autocomplete.optionsBuilder` gets called before the `Autocomplete.onSelected` handler.

1) submit text
2) build (search) options
3) select location

It gets really ugly but something like this would prevent the spurious search.

```diff
diff --git a/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart b/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
index c8f8e641..709df3e3 100644
--- a/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
@@ -80,7 +80,14 @@ class WhereAreYouPageState extends State<WhereAreYouPage> {
                         );
                       },
                       displayStringForOption: formatLocation,
-                      optionsBuilder: (value) {
+                      optionsBuilder: (value) async {
+                        // wait for onSelected to get called
+                        await Future.delayed(const Duration(microseconds: 1));
+                        // avoid searching a selected and formatted location
+                        if (value.text ==
+                            formatLocation(model.selectedLocation)) {
+                          return model.locations;
+                        }
                         return model.searchLocation(value.text);
                       },
                       onSelected: model.selectLocation,
```

jpnurmi:
I guess we can assume that city names don't contain parentheses, so here's another possibility:
```diff
diff --git a/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart b/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
index c8f8e641..a2d64f82 100644
--- a/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
@@ -81,6 +81,10 @@ class WhereAreYouPageState extends State<WhereAreYouPage> {
                       },
                       displayStringForOption: formatLocation,
                       optionsBuilder: (value) {
+                        if (RegExp(r'^.*\(.*\)$').hasMatch(value.text)) {
+                          // avoid searching a selected and formatted location
+                          return model.locations;
+                        }
                         return model.searchLocation(value.text);
                       },
                       onSelected: model.selectLocation,
```

oSoMoN:
> It gets really ugly but something like this would prevent the spurious search.

I have to agree this is rather ugly, I'd prefer if we could devise a cleaner approach.

> I guess we can assume that city names don't contain parentheses, so here's another possibility:

Unfortunately, it appears that the names of some cities listed in cities15000.txt do contain parentheses (e.g. "Newport (Kentucky)", "سقّز (٢)", "Τιμπού (θερινή) Πουνάκα (χειμερινή)", …).

And given the implementation of `GeoLocation.toDisplayString()`, there's no guarantee the display string will always contain parentheses (although I would expect the case without parentheses to be infrequent).

The [documentation for `Autocomplete.onSelected()`](https://api.flutter.dev/flutter/material/Autocomplete/onSelected.html) says « Any TextEditingController listeners will not be called when the user selects an option », so perhaps we could have a listener that saves the value of the last actual search in the model, and if the formatted value for that last search is equal to `value.text`, then avoid issuing a new search query?

packages/ubuntu_desktop_installer/lib/installer.dart
```diff
@@ -15,6 +16,8 @@ import 'services.dart';
 
 export 'package:ubuntu_wizard/widgets.dart' show FlavorData;
 
+const _kGeoIPUrl = 'https://geoip.ubuntu.com/lookup';
+const _kGeonameUrl = 'http://geoname-lookup.ubuntu.com/';
```
oSoMoN, line 20: Is there a reason for this not to be a https URL?

oSoMoN (approved):
LGTM overall, just a few minor comments.

jpnurmi:
> The [documentation for `Autocomplete.onSelected()`](https://api.flutter.dev/flutter/material/Autocomplete/onSelected.html) says « Any TextEditingController listeners will not be called when the user selects an option »

Hmm, this is not true. The first thing it does is to update the text editing controller's value, and then calls onSelected.

https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/widgets/autocomplete.dart#L331-L335

jpnurmi:
I'm afraid adding a listener or checking the editing controller's current value doesn't really help unless we postpone the search in one way or another. The problem is essentially the same as previously that when building options, we don't know yet if an option has been selected, which could be either by clicking an option or by submitting the field using the keyboard.

jpnurmi:
Thanks! I've opened https://github.com/canonical/ubuntu-desktop-installer/issues/652 for the extra search request.

PR #579, opened 2021-12-15
Close and destroy subiquity sockets

jpnurmi:
Close the clients that are used on startup to send status requests
to test whether subiquity is up and running, and fully destroy closed
sockets to ensure that no asynchronous operations are pending in the
background when sockets are closed and reopened at fast pace in
integration tests.

### DIFF
--- a/packages/subiquity_client/lib/src/server.dart
+++ b/packages/subiquity_client/lib/src/server.dart
@@ -113,12 +113,12 @@ abstract class SubiquityServer {
     for (var i = 0; i < 10; ++i) {
       try {
         await client.send(request);
-        await client.close();
         break;
       } on Exception catch (_) {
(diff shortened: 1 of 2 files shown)


packages/subiquity_client/lib/src/server.dart
```diff
@@ -113,6 +113,7 @@ abstract class SubiquityServer {
     for (var i = 0; i < 10; ++i) {
       try {
         await client.send(request);
+        await client.close();
```
oSoMoN, line 116: Wouldn't it be clearer to have this call outside of the for loop?

jpnurmi: Thanks! Yes, and correct too

oSoMoN (approved):
LGTM, thanks!

Issue #415, opened 2021-10-08
Disk model and vendor names from subiquity


jpnurmi:
Currently, the UI looks up disk model and vendor names using libudev. However, local devices typically don't match with dry-run devices.

Even though this is only a development time/dry-run mode problem, it would be nice if subiquity would provide that information in `Disk` objects, so we could get rid of the libudev FFI-bindings (unless it's needed for network devices later).

The values should probably be the `_ENC` variants if they exist, to display special characters correctly in the UI. The one in simple.json has only excessive spaces, but some may have trademark signs etc.

```bash
jq '.storage.blockdev."/dev/sda".ID_MODEL_ENC' examples/simple.json 
"QEMU\\x20HARDDISK\\x20\\x20\\x20"

jq '.storage.blockdev."/dev/sda".ID_MODEL' examples/simple.json 
"QEMU_HARDDISK"
```
```bash
jq '.storage.blockdev."/dev/sda".ID_VENDOR_ENC' examples/simple.json 
"ATA\\x20\\x20\\x20\\x20\\x20"

jq '.storage.blockdev."/dev/sda".ID_VENDOR' examples/simple.json 
"ATA"
```

PR #580, opened 2021-12-15
Use disk model and vendor names from subiquity 

jpnurmi:
Disk model and vendor names were added in canonical/subiquity#1138. It's no longer necessary to lookup libudev through FFI, and as a bonus, the names match the machine config in dry-run mode.

Close: #415

### DIFF
--- a/packages/subiquity_client/lib/src/types.dart
+++ b/packages/subiquity_client/lib/src/types.dart
@@ -237,6 +237,8 @@ class Disk with _$Disk {
     String? ptable,
     bool? preserve,
     bool? bootDevice,
+    String? model,
+    String? vendor,
   }) = _Disk;
 
   factory Disk.fromJson(Map<String, dynamic> json) => _$DiskFromJson(json);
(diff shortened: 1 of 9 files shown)


jpnurmi:
For example, a dry-run with the default simple.json:

| Before, /dev/sda not mounted | Before, /dev/sda mounted | After, /dev/sda from simple.json |
|---|---|---|
| Screenshot from 2021-12-15 10-05-47 | Screenshot from 2021-12-15 10-23-43 | Screenshot from 2021-12-15 10-05-10 |

robert-ancell (approved):
Yay! That looks much tidier :)

jpnurmi:
Hmm, I wish we could get the same information for network devices from NetworkManager. Then we could get rid of the libudev FFI bindings altogether.

There's nm_device_get_product() and nm_device_get_vendor() in libnm but no corresponding properties in the D-Bus API.

robert-ancell:
> Hmm, I wish we could get the same information for network devices from NetworkManager. Then we could get rid of the libudev FFI bindings altogether.
> 
> There's nm_device_get_product() and nm_device_get_vendor() in libnm but no corresponding properties in the D-Bus API.

If that's information that NetworkManager consumers are likely to use then we probably should add helpers like that to nm.dart.

Issue #581, opened 2021-12-15
Please enable (experimental) autoinstalls


paride:
For automated ISO testing we need a way to perform fully automated installs from the images. Subiquity is already able to perform automated installs, however autoinstalls have been explicitly disabled in the desktop installer because the UI for them is missing: https://github.com/canonical/ubuntu-desktop-installer/commit/5d1025408ce7a4dee8339f1ce260577564c8a662. 

What I propose is to modify `snap/local/subiquity-server` to make it grep `/proc/cmdline` and if e.g. `autoinstall-force` is present then subiquity is *not* passed `--autoinstall=""`, and autoinstalls stay enabled. Wdyt?

Cc: @blackboxsw @dbungert

mwhudson:
FWIW it would be preferable for CI to have something that can manipulate the client to mimic a human performing the install. I don't know practical that is with flutter though. (openqa anyone??)

jpnurmi:
Flutter cannot help with a pristine ISO or the installer snap, but if you'd have the installer sources available, Flutter's integration test framework could run through the whole installation wizard. The PR is pretty outdated by now, but I wrote some integration tests a while back: https://github.com/canonical/ubuntu-desktop-installer/pull/283. The screencast is a dry-run, of course, but I experimented also with live runs in a chrooted environment. My dream was to one day run a proper installation in the CI this way, but I guess this is not exactly what we're looking for here...

paride:
> (openqa anyone??)

That's something I'll look into early next year.

paride:
@jpnurmi yes pristine ISO tests are really desirable, but for the moment performing an autoinstall with be enough to exercise some critical components of the install process (the autoinstall path of subiquity-server, curtin, cloud-init), moreover the autoinstall feature itself deserves testing, so this is good in any case.

PR #582, opened 2021-12-15
subiquity-server: permit autoinstall w/ override

dbungert:
If experimental-autoinstall is found in the kernel command line,
allow autoinstall to be enabled.  It was disabled in the first place due
to UI support not yet being implemented.

Fixes: #581


dbungert:
@paride - Note that using this triggers a bug with subiquity and 'netplan apply' but I guess this is still progress.

seb128 (approved):
Sounds also fine to me, thanks!

PR #591, opened 2022-01-04
Add minimal glibc FFI bindings

jpnurmi:
For starters, it provides uptime for the telemetry service in #585.

```dart
import 'package:glibc/glibc.dart';

void main() {
  print('Uptime: ${sysinfo().uptime}');
}
```


robert-ancell (approved):
LGTM

PR #592, opened 2022-01-05
Allow flavors to pass localization delegates

jpnurmi:
Additional flavor-specific localization delegates are necessary for flavors to be able to translate their own custom slides, for example (#42).


CarlosNihelton (approved):
LGTM.

Just out of curiosity: how will those flavor translations be managed (process wise)?

jpnurmi:
> Just out of curiosity: how will those flavor translations be managed (process wise)?

We are using [Weblate](https://hosted.weblate.org/projects/ubuntu-desktop-installer/) and I'm sure flavors are welcome to join the translation project if they want. :smile: Since flavors' translations are not known by the installer, flavors are free to setup any translation service/system they want, of course.

jpnurmi:
As for creating flavors of the installer, they'll create an app with ubuntu_desktop_installer as a dependency, and call the `runInstallerApp()` entry point.

https://github.com/canonical/ubuntu-desktop-installer/pull/478

PR #595, opened 2022-01-07
subiquity_client: remove unused storage v1 API

jpnurmi:
Storage v2 API is used since #334.


jpnurmi:
@dbungert Is this something you could stamp? It's only removing old unused storage v1 methods.

dbungert (approved):
Yes, this is correct, we don't want to use a mix of V1 and V2 methods.  That actually wouldn't be a problem if the client is careful, but it's not what we want.  So removing these is a good idea.

PR #603, opened 2022-01-11
ValidatedFormField: expose focus node

jpnurmi:
Hidden Wi-Fi entries require more fine-grained control over focus. The
existing `autofocus` property is not enough because keyboard focus must
be dynamically requested and removed whenever the hidden Wi-Fi view is
expanded or collapsed.

### DIFF
--- a/packages/ubuntu_wizard/test/validated_form_field_test.dart
+++ b/packages/ubuntu_wizard/test/validated_form_field_test.dart
@@ -201,8 +201,26 @@ void main() {
     expect(find.text('not equal'), findsNothing);
   });
 
-  testWidgets('focus', (tester) async {
+  testWidgets('focus node is attached', (tester) async {
     final focusNode = FocusNode();
+
+    await tester.pumpWidget(
+      MaterialApp(
+        home: Material(
+          child: ValidatedFormField(focusNode: focusNode),
+        ),
+      ),
+    );
+
+    expect(
+      tester.widget<TextField>(find.byType(TextField)).focusNode,
+      equals(focusNode),
+    );
+  });
+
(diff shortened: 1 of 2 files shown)


jpnurmi:
Note to self: should be added to the others too if this is merged before https://github.com/canonical/ubuntu-desktop-installer/pull/541

packages/ubuntu_wizard/test/validated_form_field_test.dart
```diff
@@ -200,4 +200,22 @@ void main() {
...
     expect(find.text('not equal'), findsNothing);
   });
+
+  testWidgets('focus', (tester) async {
+    final focusNode = FocusNode();
+
+    await tester.pumpWidget(
+      MaterialApp(
+        home: Material(
+          child: ValidatedFormField(autofocus: true, focusNode: focusNode),
```
CarlosNihelton, line 210: One minor question: would there be an interesting test case in which we could explore `autofocus: false`? Just wondering the value of testing `autofocus: true` when the purpose of the PR is to ensure fine-grained control over the focus. Or, instead, should we pay more attention to the fact that the API exposes the `focusNode` as the commit message states and checked at the following expectation:

```dart
expect(
      tester.widget<TextField>(find.byType(TextField)).focusNode,
      equals(focusNode),
    );
```

CarlosNihelton (approved):
LGTM

packages/ubuntu_wizard/test/validated_form_field_test.dart
jpnurmi: Thanks for nudging me to add more tests! :smile:

PR #596, opened 2022-01-07
Updated Subiquity to the latest and greatest

CarlosNihelton:
Hello, @jpnurmi !

Per our conversation earlier, we would like to have the next cloud image built with the latest revision of Subiquity, for further testing with WSL.

The proposed update was made with the help of the `update-subiquity` script. The changes generated by the script on the `snapcraft.yaml` file consisted of only space removals and one replacement of double quotes by single quotes, so I didn't include that in this pull request.

PR #607, opened 2022-01-13
snapcraft.yaml: apply formatting changes by scripts/update-subiquity

jpnurmi:
To avoid unrelated formatting changes for Subiquity updates (e.g. #596).


CarlosNihelton (approved):
Thanks for doing that! Everybody else who would ever come accross this script would never push those changes. 😅

PR #608, opened 2022-01-13
Sort pub dependencies

jpnurmi:
Enable `sort_pub_dependencies` to ensure that dependencies stay sorted.
Sorting dependencies makes maintenance easier, and it also plays better
together with Pubspec Assist.


CarlosNihelton (approved):
LGTM. I'll do the same for my splash screen.

Issue #477, opened 2021-11-08
Subiquity symlinks


jpnurmi:
There are several symlinks to Subiquity in the project tree:
```
packages/subiquity_client/test/socket -> ../subiquity/.subiquity/socket
packages/ubuntu_desktop_installer/test/socket -> ../../subiquity_client/test/socket
packages/ubuntu_desktop_installer/subiquity -> ../subiquity_client/subiquity
packages/ubuntu_wsl_setup/test/socket -> ../../subiquity_client/test/socket
packages/ubuntu_wsl_setup/subiquity -> ../subiquity_client/subiquity
```
The purpose of these symlinks is to be able to start Subiquity and resolve the socket path in dry-run mode. We should find a better way to locate Subiquity because the symlinks are causing problems for remote snap builds and any installer flavors would also have to duplicate them.

PR #610, opened 2022-01-13
Resolve the location of subiquity_client without symlinks

jpnurmi:
This change makes use of the `package_config` Dart package to resolve
the location of the `subiquity_client` because the symlinks to
`subiquity_client/subiquity` and `.subiquity/socket` in the project
tree are problematic in many ways.

Not only we've had to duplicate the symlinks for both the desktop
installer and the WSL setup, but any desktop installer flavor would
have to create similar symlinks too, to be able to dry-run the
installer flavor.

Also, the symlinks are problematic for snap builds. The socket symlinks
break local (as in source-type: local) builds. Using Git as the
source-type circumvents this problem, but the drawback is that then
snapcraft doesn't pick local uncommitted changes. Lastly, the symlinks
break remote builds too.

Close: #477

### DIFF
--- a/packages/subiquity_client/lib/src/server.dart
+++ b/packages/subiquity_client/lib/src/server.dart
@@ -52,7 +52,7 @@ abstract class SubiquityServer {
     return _subiquityPath ??= await _findSubiquityPath();
   }
 
-  // Finds local subiquity in relation to the `subiquity_client` Dart package.
+  // Finds local subiquity relative to the `subiquity_client` Dart package.
   Future<String> _findSubiquityPath() async {
     Object? error;
     final config = await findPackageConfig(
--- a/packages/subiquity_client/lib/src/server.dart
+++ b/packages/subiquity_client/lib/src/server.dart
@@ -2,6 +2,7 @@ import 'dart:io';
 
 import 'package:http/http.dart';
 import 'package:meta/meta.dart';
+import 'package:package_config/package_config.dart';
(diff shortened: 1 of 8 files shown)


didrocks:
This looks good to me. Happy for this to be strengthen! I don’t see any immediate negative impact on WSL, but I would appreciate @CarlosNihelton to have a look as he is the one using it in a dev environment from day to day :)

CarlosNihelton:
That's really great! Just tested on Ubuntu 22.04 on WSL2 and it works great. We benefit from hardening the build environment. Great job!

packages/subiquity_client/lib/src/server.dart
```diff
@@ -35,10 +36,41 @@ abstract class SubiquityServer {
...
+  }
+
+  String? _subiquityPath;
+
+  // Returns the location of the local subiquity submodule.
+  Future<String> _getSubiquityPath() async {
+    return _subiquityPath ??= await _findSubiquityPath();
+  }
+
+  // Finds local subiquity in relation to the `subiquity_client` Dart package.
```
CarlosNihelton, line 55:
```suggestion
  // Finds local subiquity relative to the `subiquity_client` Dart package.
```

CarlosNihelton (approved):
Very, very minor nitpick. Otherwise, LGTM.

CarlosNihelton (approved):
+1

jpnurmi:
Thanks!

Issue #583, opened 2021-12-17
Use libhandy to round all four window corners


Feichtmeier:
Most gnome applications preinstalled in Ubuntu 21.10 use libhandy to somehow modify gtk3 that all four window corners are round:

![grafik](https://user-images.githubusercontent.com/15329494/146516094-41c19fcf-8886-450b-a3d3-15d54018f322.png)

We've managed to include and use libhandy inside the settings project:
https://github.com/Feichtmeier/settings/commit/7f62addf697a7eabdb0bf9431298bd9787e009ce

![grafik](https://user-images.githubusercontent.com/15329494/146516409-0e3a1711-00c0-4ffa-9f65-4110ac1ea1f7.png)
![grafik](https://user-images.githubusercontent.com/15329494/146516641-d38d8b1b-e1cc-402d-a8ef-6ebcec34b2ca.png)

Could this also be done in the installer?

I think it does not only "look better"/"look more modern" but also increases the visual consistency with the rest of the Ubuntu desktop after installation.

seb128:
@robert-ancell hey, do you have any opinion about using libhandy?

robert-ancell:
I think it should be OK because:
- libhandy is being used in many apps.
- We're only using a small subset of it.
- For the installer, we'll be able to control the exact version of libhandy, which avoids the sometimes unstable nature of it.

It may be worth investigating how much code libhandy is using for the rounded windows, and moving that into the Flutter shell or the desktop installer, but that can be done at a later stage.

seb128:
@Feichtmeier would you maybe be wanting to work on the PR for the change?

jpnurmi:
I'm sure there's more code to strip off, but here's the result of throwing out everything except `HdyWindow`, `HdyHeaderBar`, and their dependencies:

- lines of code (src/*) went down from ~31k to ~4k
- release build size went down from 896kb to 151kb

Feichtmeier:
> @Feichtmeier would you maybe be wanting to work on the PR for the change?

I can but all the work has been done by @jpnurmi in this case so I can replicate the changes in the Linux directory in the settings app in the installer if you like

jpnurmi:
Using libhandy is straightforward, but getting it in the snap is the laborious part. core20/20.04 has too old libhandy (0.x) without support for rounded corners, so I suppose libhandy 1.x would need to be built as part of the snap.

jpnurmi:
[...]

PR #611, opened 2022-01-13
snapcraft.yaml: use flutter from git

jpnurmi:
This is a pre-requisite for being able to use libhandy for rounded
window corners:

https://github.com/canonical/ubuntu-desktop-installer/issues/583#issuecomment-1010723573


snap/snapcraft.yaml
```diff
@@ -117,59 +117,71 @@ parts:
...
+    stage-packages:
+      - clang
+      - cmake
+      - curl
+      - libgtk-3-dev
+      - ninja-build
+      - unzip
+      - xz-utils
+      - zip
+    override-prime: ''
```
jpnurmi, line 150: This is only for building so we don't want to prime anything. I tried `- -"*"` but it errors out because of some flutter binary cache files. I saw this in Olivier's `firefox-snap-experiments` and it seems to do the trick.

seb128 (commented):
How stable is flutter git stable?

If using libhandy isn't fixing the rendering issue does it bring enough value to justify slowing the build by adding extra parts and switching the flutter serie?

jpnurmi:
> How stable is flutter git stable?

Flutter's channels are managed as git branches. The stable branch/channel always matches the latest stable release. It's the same version as the Flutter snap fetches from the stable channel.

> If using libhandy isn't fixing the rendering issue does it bring enough value to justify slowing the build by adding extra parts and switching the flutter series?

Yeah, it's a pity that it didn't help, so the only gain is eye candy... :(

kenvandine:
I would still be in favor of this.  The build changes of using git for flutter doesn't add any risk and is easy to maintain.

snap/snapcraft.yaml
kenvandine: You should drop all of those stage-packages in the flutter-git part.

jpnurmi:
Thanks for the help. I minimized the changes so that now it's not much more than switching from `build-snaps: flutter` to `after: flutter-git`.

seb128 (approved):
You guys have been convincing, let's use the git vcs ;-)

PR #612, opened 2022-01-13
Adjust welcome page design

### LINKED ISSUE #567: Welcome screen alignment and design
I agree with [this comment](https://discourse.ubuntu.com/t/new-desktop-installer-preview-build/24765/49) on discourse that the alignment of this window is inconsistent with the rest of the pages.

Since everything in the installer is left aligned this ListView in the middle should also be aligned to the left.

grafik

The question is what happens then with the white space that will be on the right? Eventually add the Ubuntu CoF icon?

grafik

Or the current mascot for the current release?

grafik

Feichtmeier:
- add the mascot_white.svg as the current ubuntu mascot logo
- wrap the language selection in a row together with a primary color colored mascot image

TODO:
- [x] update readme with a new picture
- [x] fix test
- [x] loading page

Closes #567


jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2022-01-14 10-32-27 | Screenshot from 2022-01-14 10-32-02 |
| Screenshot from 2022-01-14 10-32-59 | Screenshot from 2022-01-14 10-31-28 |

Feichtmeier:
Thanks @jpnurmi !
Should have posted the result myself :see_no_evil: 

One question would be if you agree with the less saturated Svg for the light theme? I thought it would be too bright for the light theme with full opacity:

grafik

jpnurmi:
Furthermore, the loading screen, which is shown while connecting to subiquity, should be updated accordingly:

https://user-images.githubusercontent.com/140617/149500274-0dd0caa3-1000-49c7-a006-c06d893c726f.mp4

Feichtmeier:
Oh missed the loading screen will do once we've settled for one design

Right, making it left aligned was the initial issue :D 
It's a bit tricky to make the spaces look good though. How about this?

grafik
grafik

jpnurmi:
It's better but there's still a noticeable jump. The two lists have the same rounded frame that shifts and resizes a bit.

Would the layout feel more stable if the welcome page had a similar header (e.g. "Select language:") as the next page, and if the width would match to ensure that the listview stays in the same place instead of jumping a bit? :)

Feichtmeier:
Hm I think it is correct to make them all aligned to a specific position like left or space between etc
But that the pages can look different is totally okay feels natural IMO. The wizard page gives a visual skeleton that stays the same.

Feichtmeier:
ya

jpnurmi:
https://user-images.githubusercontent.com/140617/149503281-b15a115f-bbcf-43ba-a27b-5e285d464500.mp4

vs.

https://user-images.githubusercontent.com/140617/149503273-057dc261-4049-4a5f-891b-9cdc5d2e2595.mp4

vs.

https://user-images.githubusercontent.com/140617/149503297-8dcf01ff-a684-4ba5-a890-9efc5f59bd8b.mp4

jpnurmi:
It wouldn't feel jumpy if it wasn't the exact same rounded frame in the (almost) same place. Center alignment is not really that bad, is it? :)

Feichtmeier:
The "Chose a language" title does not hurt and is nice for accessibility so I can add it ofc
But I don't think it feels jumpy because of the nice transition maybe.

Feichtmeier:
Your suggestion plus full orange mascot in a very small window (800x600)
grafik


Normal resolution:
grafik

I agree this looks like a good result. If you agree, too that is :) (Where are all the designers? :swan: :sweat_smile: )

jpnurmi:
The non-jumpy version with the same position and width looks best in my eyes. :nerd_face:

For the other versions, the navigation transition in between helps, of course. These transitions are considered mobile-centric, though, and the Flutter team is probably going to remove them on desktop. The transitions were already taken out for a while until the change had to be reverted: https://github.com/flutter/flutter/pull/82596 -> https://github.com/flutter/flutter/issues/88221 -> https://github.com/flutter/flutter/pull/89997. If/when it happens again, we could add our own transitions to avoid the jump, but without a jump, there wouldn't be a problem to begin with... :wink:

Feichtmeier:
> The non-jumpy version with the same position and width looks best in my eyes. :nerd_face: (resize the graphics if it gets too packed?)
> 
> For the other versions, the navigation transition in between helps, of course. These transitions are considered mobile-centric, though, and the Flutter team is probably going to remove them on desktop. The transitions were already taken out for a while until the change had to be reverted: https://github.com/flutter/flutter/pull/82596 -> https://github.com/flutter/flutter/issues/88221 -> https://github.com/flutter/flutter/pull/89997. If/when it happens again, we could add our own transitions to avoid the jump, but without a jump, there wouldn't be a problem to begin with... :wink: 

Ok. I'm not sure that you mean though. Does the last version fit for you?

jpnurmi:
> Ok. I'm not sure that you mean though. Does the last version fit for you?

The current has different widths. See the second vs. third video in https://github.com/canonical/ubuntu-desktop-installer/pull/612#issuecomment-1013010864

Feichtmeier:
grafik

ya

Like this @jpnurmi ?
Still need to find where you define that loading screen

jpnurmi:
Looks nicer to me! What do you think?

The loading screen stub is defined in `installer.dart`.

Feichtmeier:
> Looks nicer to me! What do you think?
> 
> The loading screen stub is defined in `installer.dart`.

Well, at some point you will have a jump in the pages anyways but I like both versions so no hard preference.

What's the reason for this funny empty box in the loading screen btw? 
Shouldn't a circular progress indicator fit better?

jpnurmi:
> What's the reason for this funny empty box in the loading screen btw?
> Shouldn't a circular progress indicator fit better?

A ​circular progress indicator wasn't well-received so it was left as a placeholder frame to avoid endless bikeshedding on the design. :smile: Before, the window would stay plain empty until the subiquity connection had been established, which gave a broken feeling that the app would have crashed. We should have a nice design for it, but showing an appbar + frame + disabled buttons alone made it feel like the UI is running but not ready/populated.

jpnurmi:
I'm tempted to propose having the logo in the center and using a hero animation to move it in place on the real welcome screen. Unfortunately, it won't look nice under normal circumstances when the loading phase is short.

Typically, it shouldn't take more than a second to establish the connection. However, on low-end hardware or highly loaded systems, it could take longer. When it only takes a fraction of a second, the busy indicator or other animation flashes so quickly that it doesn't look nice for most users. Maybe we should show the loading screen for at least a couple of seconds... :thinking:

jpnurmi:
Remember that for end-users it's a rare exception that the startup takes a noticeable amount of time. Unless we add an intentional splash screen delay, for most users the startup would look closer to this:

https://user-images.githubusercontent.com/140617/149548112-1a21c466-1b82-4861-aa54-8eaf8f06ada7.mp4

Feichtmeier:
As long as it can pulse for a half a second I don't think that this is a problem. Maybe it is no problem at all if you only see a fast flash of the picture. Well it is hard to tell how the effect is if run in a vm and from source. Would need to see it on a real machine.
So you would prefer the gray square?

ya

jpnurmi:
I would prefer that the logo didn't jump abruptly from the center to the side. There's intentionally no transition from the placeholder to the welcome screen because that way it looks like the page simply gets populated with a delay, as explained before. If the subtle gray lines jumping slightly bothered me earlier, the big orange logo stands out even more... :upside_down_face:

What comes to the pulsing fade animation, it's a cool idea but somehow I personally feel like it would be more appropriate to keep the bling-bling to the minimum in an app like this. Subtle animations or transitions are ok to make the experience smooth, but IMO they should be so subtle that the user doesn't even pay attention to them.

Feichtmeier:
I try gray box of doom + logo pulsing
I think the pulsing logo is rather a progress indicator than a bling bling.
I think some sort of progress indication is a must, the gray box alone looks totally disabled

Feichtmeier:
@jpnurmi if you do not want to have an animation what about making the mascot gray in the loading screen:


ya

jpnurmi:
This is the best proposal so far. I'm still skeptical about the strong color change that will cause flickering in production, though. I have a feeling that development time dry-runs and hot reloads are misleading here. We should optimize for end-users, not for developers. :)

In production, the subiquity daemon is already kicked off at system startup. When the installer starts, normally it's able to connect to subiquity roughly around the same time Flutter is ready to render, give or take. If and when Flutter gets there first, a matching empty loading screen doesn't cause flickering but makes it look like the UI merely gets populated with a tiny delay. In other words, the content doesn't noticeably change after a few frames of the first rendering.

Dry-run startups and hot reloads are vastly different from live startups. The installer has to kill an existing subiquity server process and start a new one after the Dart entry point has already been reached. In this setup, Flutter wins easily and gets to render way before subiquity is ready. This is why the loading screen is visible for a good while during dry-run startups and hot reloads.

Feichtmeier:
> This is the best proposal so far. I'm still skeptical about the strong color change that will cause flickering in production, though. I have a feeling that development time dry-runs and hot reloads are misleading here. We should optimize for end-users, not for developers. :)
> 
> In production, the subiquity daemon is already kicked off at system startup. When the installer starts, normally it's able to connect to subiquity roughly around the same time Flutter is ready to render, give or take. If and when Flutter gets there first, a matching empty loading screen doesn't cause flickering but makes it look like the UI merely gets populated with a tiny delay. In other words, the content doesn't noticeably change after a few frames of the first rendering.
> 
> Dry-run startups and hot reloads are vastly different from live startups. The installer has to kill an existing subiquity server process and start a new one after the Dart entry point has already been reached. In this setup, Flutter wins easily and gets to render way before subiquity is ready. This is why the loading screen is visible for a good while during dry-run startups and hot reloads.

I understand that you want to maximize the stability, but I don't agree that animations in the installer, the very first application users consume which makes it very important from a marketing perspective, are "bling bling". In the new windows 11 installer, every single page has an animation of the graphics on the left (and yes, every page has a big fancy graphic on the left).

Anyways, I pushed the last proposal and also an attempt to fix the test. It looks like something is wrong with the word "English" because I can tap the italian tile and it is selected and the french one but not the english one :D

seb128:
@elioqoshi could you or someone else from design review the proposed changes and give us guidance?

elioqoshi:
Thanks @Feichtmeier !

Ideally we should get some input from the brand team here as well since it's using brand elements (@waynecrosby ).

> I understand that you want to maximize the stability, but I don't agree that animations in the installer, the very first application users consume which makes it very important from a marketing perspective, are "bling bling". In the new windows 11 installer, every single page has an animation of the graphics on the left (and yes, every page has a big fancy graphic on the left).

I agree with the sentiment. Making the installer experience more engaging would definitely be welcome. We would need to put some planning and further efforts on things like animations however. Having animations for the sake of animations could eventually lead to accessibility issues however (some people find constant movements on screen triggering for example). 

I think best here is to have an appropriate illustration for both light and dark theme. The Impish Indri illustration for example is not ideal for light theme as it can't simply be inverted (inverted eyes look weird in this case) so ideally we should get some input from the brand team on here.

jpnurmi:
Can we focus on the design of the welcome screen and leave the "loading" screen out of this discussion?

It's an internal implementation detail that there may be a placeholder screen visible for a blink of an eye. The sole purpose of that is to bring the content visible as soon as possible. Please don't try to insert an animation (or even a large logo with a different color) that is visible for a frame or two. In production, it won't look like an animation but more like a rendering glitch. :smile: The videos above have been taken in a development environment that has a large delay and does not match how it looks in production.

Feichtmeier:
> Can we focus on the design of the welcome screen and leave the "loading" screen out of this discussion?
> 
> It's an internal implementation detail that there may be a placeholder screen visible for a blink of an eye. The sole purpose of that is to bring the content visible as soon as possible. Please don't try to insert an animation (or even a large logo with a different color) that is visible for a frame or two. In production, it won't look like an animation but more like a rendering glitch. smile The videos above have been taken in a development environment that has a large delay and does not match how it looks in production.

Okay so the only difference between the loading screen and the welcome screen is that the list is empty, right?
Should I also keep the header "Chose your language" then?

jpnurmi:
> Okay so the only difference between the loading screen and the welcome screen is that the list is empty, right?
> Should I also keep the header "Chose your language" then?

That would be great! :+1:

Feichtmeier:
ya1
ya

Ok

jpnurmi:
Hi @waynecrosby, could you confirm if these proposed design changes are ok for the desktop installer's welcome page? The main motivation for the change is that other pages are left-aligned.

| Before | After |
|---|---|
| Screenshot from 2022-02-01 10-26-00 | Screenshot from 2022-02-01 10-24-42 |
| Screenshot from 2022-02-01 10-25-53 | Screenshot from 2022-02-01 10-24-35 |

waynecrosby:
@jpnurmi We do have another version of the mascot (see below)
install-3
However, I like the use of the same colour mascot for light and dark as it allows the user to just concentrate on the theme colour change. So the long and short is yes it looks good.

jpnurmi (approved):
Thanks, @Feichtmeier! And everyone else involved! :smile:

Feichtmeier:
yw and thanks for thre time-consuming review!
should I resolve conflicts again or will you do it?

jpnurmi:
Done! Should be good to go once the CI passes. Design changes are clearly the hardest changes. :smile:

PR #621, opened 2022-01-18
CI: build snap

jpnurmi:


seb128 (commented):
Thanks, it is useful to have. Small nitpick questions before giving the +1

1. What's the point on having workflow_dispatch? Do we expect to need to trigger a snap build manually sometimes?

2. Should we limit to pull_request against main? That's probably the only target we use at the moment but the job is not trivial and it might help not wasting resources in the future?

seb128:
Thanks! As mentioned on the chat we should keep an eye on the impact that has on our workflow quota since each change to a merge request is going to create another 120Mb artifact. We can easily lower the artifact retention period for the project but maybe we could also try to clean artifacts on merge if needed

seb128 (approved):
Thanks!

PR #622, opened 2022-01-19
subiquity: update for apt umount workaround

dbungert:


dbungert:
I was able to do a smoketest with default options and the install completed in my KVM setup.

jpnurmi (approved):
Success! Thanks for resolving the issues! :+1:

PR #623, opened 2022-01-19
Clean up explicit scrollbars that are no longer needed

jpnurmi:
Scrollbars are automatically applied on desktop since flutter/flutter#78588

### DIFF
--- a/packages/ubuntu_wizard/test/rounded_list_view_test.dart
+++ b/packages/ubuntu_wizard/test/rounded_list_view_test.dart
@@ -28,15 +28,15 @@ void main() {
       findsOneWidget,
     );
 
-    // listview inside scrollbar
+    // listview inside rounded container
     expect(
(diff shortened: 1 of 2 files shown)


packages/ubuntu_wizard/test/rounded_list_view_test.dart
```diff
@@ -28,15 +28,15 @@ void main() {
       findsOneWidget,
     );
 
-    // listview inside scrollbar
+    // listview inside rounded container
```
jpnurmi, line 31: It's not visible in the diff, but there's a "scrollbar inside rounded container" test above. Since the internal ListView structure is different than what we had, I changed the "container -> scrollbar -> listview" test to "container -> scrollbar & container -> listview")

oSoMoN (approved):
LGTM, thanks!

PR #626, opened 2022-01-21
WhoAreYouPage: ignore the "show password" button for tab focus traversal

jpnurmi:
Split off from #541.

Before, it took 2 steps to move keyboard focus from the password field to the password confirmation field. Now, the icon button is skipped so that the password fields are adjacent nodes in the focus chain. The downside is that there's no way to access the "show password" button via keyboard...

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_widgets.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_widgets.dart
@@ -135,12 +135,16 @@ class _PasswordFormField extends StatelessWidget {
         final model = Provider.of<WhoAreYouModel>(context, listen: false);
         model.password = value;
       },
-      suffixIcon: IconButton(
-        onPressed: () {
-          final model = Provider.of<WhoAreYouModel>(context, listen: false);
-          model.obscureText = !model.obscureText;
-        },
-        icon: Icon(obscureText ? YaruIcons.hide : YaruIcons.view),
+      suffixIcon: Focus(
+        canRequestFocus: false,
+        descendantsAreFocusable: false,
+        child: IconButton(
+          onPressed: () {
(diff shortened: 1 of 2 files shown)


jpnurmi:
## Before
https://user-images.githubusercontent.com/140617/150534617-f151f433-0af2-48a4-bb36-93024c81d479.mp4

## After
https://user-images.githubusercontent.com/140617/150534632-d1e2cb32-1e26-41f1-ba4e-ec23f8258e2f.mp4

oSoMoN:
This behaviour makes more sense to me, however we're sacrificing accessibility here.
How does it currently work in ubiquity?
Are you aware of some sort of convention on a keyboard shortcut to toggle password visibility in obscured text fields in other applications / widget toolkits?

jpnurmi:
I can't see any common convention. Google has a separate checkbox which would be easy to handle. Online services often have nothing, some have focusable icons/buttons, and some don't allow focus.

### Ubiquity (nothing)
ubiquity
I could swear I've seen a "[x] Show password" checkbox in Ubiquity before, but it's not there even when going back to 15.10.

### GNOME Settings (nothing)
settings

### Ubuntu One (nothing)
ubuntu

### Google (checkbox)
google

### Twitter (focusable icon)
twitter

### Facebook (non-focusable icon)
facebook

### LinkedIn (focusable button)
linkedin

### Microsoft (nothing)
microsoft

### Apple (nothing)
apple

### Discord (nothing)
discord

### Zoom (nothing)
zoom

### Instagram (focusable button)
instagram

### Trello (nothing)
trello

oSoMoN (approved):
Thanks for the thorough research!
It looks like very few of these actually care about accessibility.
Let's go with this approach for now, we can still revisit/revert later on.

jpnurmi:
Here's a couple of more desktop cases:

### Windows 11 settings (non-focusable icon)
windows

### macOS system preferences (nothing)
macos

PR #627, opened 2022-01-21
ValidatedFormField: test initial value rebuild

jpnurmi:
This is an additional test cherry-picked from #541 for a currently untested use case that regressed during the refactoring.


oSoMoN (approved):
LGTM, thanks!

PR #629, opened 2022-01-21
ValidatedFormField: only set default spacing if appropriate

jpnurmi:
If there's no "success widget" at all, don't inject empty spacing at the
end because it would cause misalignment in a column together with other
widgets.


jpnurmi:
| Before | After |
|---|---|
| image | image |

oSoMoN (approved):
LGTM, thanks!

Issue #347, opened 2021-09-24
Share validated fields between Who are you page and WSL profile setup


jpnurmi:
The two pages were developed in parallel and there were issues sharing translations. We have now the infrastructure for shared translations in `ubuntu_wizard` so we should consider sharing as much of these as possible.

jpnurmi:
The validation logic in `WhoAreYouModel` and `ProfileSetupModel` should be shared too.

PR #633, opened 2022-01-25
Promote the show/hide password button to ValidatedFormField

jpnurmi:
This PR promotes the show/hide password logic from the _Who are you?_ -page to the `ValidatedFormField` widget.

Currently, only the password field in the _Who are you?_ -page has a show/hide password button. All other password and passphrase fields, including the WSL profile password, active directory password, and encryption passphrase should also have the same "show password" button. Now, simply specifying `obscureText: true` is enough to enable the built-in show/hide password functionality.

Ref: #347

### DIFF
--- a/packages/ubuntu_wizard/lib/src/widgets/validated_form_field.dart
+++ b/packages/ubuntu_wizard/lib/src/widgets/validated_form_field.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:flutter/material.dart';
 import 'package:form_field_validator/form_field_validator.dart';
+import 'package:yaru_icons/yaru_icons.dart';
 
 // The spacing between the form field and the success icon.
 const _kIconSpacing = 10.0;
@@ -39,9 +40,10 @@ class ValidatedFormField extends StatefulWidget {
   /// The label below the [TextField]
   final String? helperText;
 
-  /// This boolean is forwarded to the [TextField] to decide
-  /// if the digits should be obscured.
-  final bool obscureText;
+  /// If specified, this boolean is forwarded to the [TextField] to decide
+  /// if the digits should be obscured, and a show/hide icon button is shown
+  /// to toggle between the obscured and non-obscured state.
+  final bool? obscureText;
 
   /// The specific widget shown right to the [TextField] if the
   /// input value is valid.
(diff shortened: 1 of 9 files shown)


packages/ubuntu_wizard/lib/src/widgets/validated_form_field.dart
```diff
@@ -39,9 +40,10 @@ class ValidatedFormField extends StatefulWidget {
   /// The label below the [TextField]
   final String? helperText;
 
-  /// This boolean is forwarded to the [TextField] to decide
-  /// if the digits should be obscured.
-  final bool obscureText;
+  /// If specified, this boolean is forwarded to the [TextField] to decide
+  /// if the digits should be obscured, and a show/hide icon button is shown
```
oSoMoN, line 44: Picking nits: the text field can contain all sorts of characters, not just digits.

oSoMoN (commented):
This results in a different behaviour on the "Who are you" page: currently on main, there's only one toggle button for the two password fields, and it toggles the visibility in both fields. With this change, each password field has its own button.

Both behaviours look acceptable to me, but can you clarify whether this change is intended?

jpnurmi:
Sorry, the behavior change got buried in the commit messages. We looked at some apps and services in https://github.com/canonical/ubuntu-desktop-installer/pull/626 and found out that very few of them actually have a "show password" button. For those that have, I can't see a second password field for verification.

When a show/hide password button is embedded, I actually find both options somehow strange. Having multiple show/hide buttons next to each other makes it look a bit cluttered. On the other hand, if the show/hide button is inside a specific password field, it feels somehow strange that it toggles another field. I must admit I'm biased and hate the second option because the implementation is clumsy. :smile:

Having a separate checkbox, as Google and Microsoft have, looks actually pretty clean and intuitive.
| Google | MS |
|---|---|
| Screenshot from 2022-02-02 15-32-38 | Screenshot from 2022-02-02 15-33-02 |

oSoMoN:
Indeed, I had missed the behaviour change explanation in the commit message, my bad.
I do agree that both behaviours feel awkward, and the checkbox option would solve the problem nicely. @elioqoshi can we have your opinion on this?

jpnurmi:
we went with a checkbox instead: https://github.com/canonical/ubuntu-desktop-installer/pull/674

PR #635, opened 2022-01-25
Prepare StorageSelector for manual LVM group creation

jpnurmi:
Refactor _BootDiskSelector out of _widgets and make it a standalone
widget suitable for storage selection in the new partition vs. LVM
group selection dialog when creating new storage entries.


packages/ubuntu_desktop_installer/test/allocate_disk_space/storage_selector_test.dart
```diff
@@ -0,0 +1,61 @@
...
+            ],
+            onSelected: (v) => selected = v,
+          ),
+        ),
+      ),
+    ));
+
+    await tester.tap(find.byType(DropdownButton<int>));
+    await tester.pumpAndSettle();
+    await tester.pump();
```
oSoMoN, line 50: Is the extra `pump()` needed?

oSoMoN (commented):
The changes look sound to me (just one minor question inline), but I've tried to test them by setting `machine-config` to e.g. `examples/many-nics-and-disks.json`, and it looks like subiquity doesn't let me select any disk as a boot device, so the behaviour of the dropdown looks buggy. This is already the case on main though, it doesn't appear to be a regression caused by these changes.

oSoMoN (approved):
LGTM, thanks!

PR #636, opened 2022-01-25
snapcraft: update for os-prober

dbungert:
An os-prober field will now appear on storage v2 partition responses.
If non-null, contains info about the OS on the partition.
Also bundle os-prober into the snap.


jpnurmi:
No installation issues :+1:

seb128 (approved):
wfm

Issue #428, opened 2021-10-12
Existing OS installations


jpnurmi:
A list of existing OS installations is needed for the installation type screen:

![image](https://user-images.githubusercontent.com/140617/136984834-784643fd-3d66-4a6b-932c-ab03b23a683c.png)

https://github.com/canonical/ubuntu-desktop-installer/issues/33

corradoventu:
screen ‘Installation type’ says: ‘The computer currently has no detected operating systems…’ while I have 4 different Ubuntu versions installed in various partitions of nvme disk of my PC
Tested with subiquity 203, 209, 214, 226

jpnurmi:
We are still missing the "Reinstall/upgrade" option but existing OSes are probed and listed.

corradoventu:
I don't see other OS trying install with ubuntu-desktop-installer   0+git.74cbf061      560  from iso dated jul 23

jpnurmi:
Let's take it case by case if there are still issues detecting existing OSes.

PR #640, opened 2022-01-26
subiquity_client: add OsProber

jpnurmi:
Ref: #428

### DIFF
--- a/packages/subiquity_client/lib/src/types.dart
+++ b/packages/subiquity_client/lib/src/types.dart
@@ -196,6 +196,21 @@ enum ProbeStatus { PROBING, FAILED, DONE }
 
 enum Bootloader { NONE, BIOS, UEFI, PREP }
 
+@freezed
+class OsProber with _$OsProber {
+  @JsonSerializable(fieldRename: FieldRename.snake)
+  const factory OsProber({
+    required String long,
+    required String label,
+    required String type,
+    String? subpath,
+    String? version,
+  }) = _OsProber;
+
+  factory OsProber.fromJson(Map<String, dynamic> json) =>
+      _$OsProberFromJson(json);
+}
+
 bool? _wipeFromString(String? value) =>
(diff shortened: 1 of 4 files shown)


packages/subiquity_client/lib/src/types.dart
```diff
@@ -196,14 +196,29 @@ enum ProbeStatus { PROBING, FAILED, DONE }
 
 enum Bootloader { NONE, BIOS, UEFI, PREP }
 
+@freezed
+class OsProber with _$OsProber {
+  @JsonSerializable(fieldRename: FieldRename.snake)
+  const factory OsProber({
```
jpnurmi, line 202: P.S. Missing coverage for this const constructor is a bug in Dart: https://github.com/flutter/flutter/issues/95008 (already fixed in the latest Flutter master)

packages/subiquity_client/test/types_test.dart
```diff
@@ -115,6 +115,25 @@ void main() {
...
+      'type': 'efi',
+      'subpath': '/efi/Microsoft/Boot/bootmgfw.efi',
+      'version': '11',
+    };
+    expect(os.toJson(), equals(json));
+    expect(OsProber.fromJson(json), os);
+  });
+
   test('partition', () {
     const keyboard = Partition(
```
oSoMoN, line 138: Not strictly related to this change, but the name of the test object is misleading, it should be `partition`, not `keyboard`. Probably a copy-paste leftover from another test…

oSoMoN (approved):
LGTM, see just one minor comment for something unrelated to this change.

packages/subiquity_client/test/types_test.dart
jpnurmi: i hope it's fine as a separate commit

oSoMoN: Yes, that's actually better, thanks!

PR #643, opened 2022-01-28
Allow re-registration of services for integration tests

jpnurmi:
Under normal circumstances, services are never supposed to be registered
multiple times.

Integration tests are an exception because they are run by calling the
app entry point multiple times in the same process. Since each app
instance registers its services, the re-registration assertion would
fail.

Relax the rules by removing the unimportant assert that was originally
added there for sanity checking the expected behavior described above,
without taking integration tests into account.


CarlosNihelton (approved):
👍

PR #645, opened 2022-01-28
Integration tests: update resolving of the .subiquity directory

jpnurmi:
Integration tests were accidentally broken by #610. The tests are no
longer correctly locating the .subiquity directory. Exposing the lookup
logic in SubiquityServer allows integration tests to locate .subiquity
for verifying output files.

PR #648, opened 2022-01-31
Fix the WSL Setup integration test

jpnurmi:
- update goldens (the default user is no longer written to wsl.conf)
- fix verifying of state and config files (wait for subiquity to be ready writing them)


jpnurmi:
This will be rebased after the dependencies are in. Only the last two commits are relevant. :)

CarlosNihelton:
Well, the last two are pretty straightforward. No objections.

jpnurmi:
I had to follow up #645 with a couple of fixes because I screwed it up while applying some totally innocent final cleanups... :man_facepalming:

Issue #546, opened 2021-11-30
Increase the visual density?


jpnurmi:
The default visual density of Flutter desktop apps is relatively low. As mentioned in the Material [density guidelines](https://material.io/design/layout/applying-density.html#usage), increasing the density could be beneficial especially for those screens that have lists and tables (languages, keyboard layouts, access points, partitions).

> By making more content available on-screen, increasing the density of lists, tables, and long forms makes information easier to scan, view, and compare.

Flutter allows specifying the visual density on a scale from -4.0 to 4.0, where a low scale means high density. On desktop platforms, the default density scale is -2.0 aka. "compact". The following screenshots illustrate the difference with increased densities:

| -2.0 | -3.0 | -4.0 |
|---|---|---|
| ![Screenshot from 2021-11-30 19-39-49](https://user-images.githubusercontent.com/140617/144112217-045d4f55-2c02-48b7-bb15-5ed06db8832f.png) | ![Screenshot from 2021-11-30 19-40-05](https://user-images.githubusercontent.com/140617/144112233-e350211a-ae75-461e-83e1-71bd21ce96b4.png) | ![Screenshot from 2021-11-30 19-40-20](https://user-images.githubusercontent.com/140617/144112276-f0cd3b48-0f0c-4a86-ab2a-31eff5174d1b.png) |
| ![Screenshot from 2021-11-30 20-49-12](https://user-images.githubusercontent.com/140617/144117559-dfc624a5-e5b2-4559-9d48-f4b39c33f9c6.png) | ![Screenshot from 2021-11-30 20-01-24](https://user-images.githubusercontent.com/140617/144117151-28b06da8-b961-478d-96ae-a93f07914a14.png) | ![Screenshot from 2021-11-30 20-01-53](https://user-images.githubusercontent.com/140617/144117149-c927e2f0-7233-47ce-b735-4bd774b3c7a6.png) |

oSoMoN:
I am no designer but to my untrained eye a value of -4.0 looks way more desktop-friendly indeed.

Feichtmeier:
If you limit this to list tiles I think this is a good idea.
However I would leave buttons as-is

You could try this yaru.dart branch and see if this fits for you:
https://github.com/ubuntu/yaru.dart/tree/dense_tiles

jpnurmi:
Some random remarks:
[...]

PR #654, opened 2022-02-01
Installer: increase the visual density of lists and tables

jpnurmi:
As suggested in [Material design guidelines](https://material.io/design/layout/applying-density.html#usage):

> By making more content available on-screen, increasing the density of lists, tables, and long forms makes information easier to scan, view, and compare.

## Desktop Installer

| Before | After |
|---|---|
| Screenshot from 2022-02-01 10-50-07 | Screenshot from 2022-02-01 10-50-26 |
| Screenshot from 2022-02-01 10-50-38 | Screenshot from 2022-02-01 10-50-46 |
|  Screenshot from 2022-02-01 10-51-12 | Screenshot from 2022-02-01 10-51-19 |
| Screenshot from 2022-02-01 10-51-53 | Screenshot from 2022-02-01 10-51-59 |

## WSL Setup

| Before | After |
|---|---|
| Screenshot from 2022-02-01 10-54-43 | Screenshot from 2022-02-01 10-55-04 |

Close: #546

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/installer.dart
+++ b/packages/ubuntu_desktop_installer/lib/installer.dart
@@ -106,8 +106,12 @@ class UbuntuDesktopInstallerApp extends StatelessWidget {
   static FlavorData get defaultFlavor {
     return FlavorData(
       name: 'Ubuntu',
-      theme: yaruLight,
-      darkTheme: yaruDark,
+      theme: yaruLight.copyWith(
+        listTileTheme: ListTileThemeData(dense: true),
+      ),
+      darkTheme: yaruDark.copyWith(
+        listTileTheme: ListTileThemeData(dense: true),
+      ),
     );
   }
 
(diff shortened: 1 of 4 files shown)


jpnurmi:
It's not just you. :) It may be a matter of getting used to, but the small font does look strange at first sight.

Perhaps it would be better to choose the [`visualDensity`](https://github.com/canonical/ubuntu-desktop-installer/issues/546) approach instead? I recall it had some undesired side effects on other UI elements, though, but maybe we can scope it to ListViews only. What do you think, @Feichtmeier?

Feichtmeier:
Sorry for not answering directly here - but I agree that the tiny text looks strange.
I made the inputs dense in yaru.dart https://github.com/ubuntu/yaru.dart/pull/131 which did not change the text size and only made those super huge inputs smaller. (need to release this yaru version)

Eventually we could try to extract a design into yaru_widgets with the right size we can agree on?

jpnurmi:
We can manually wrap the installer's long lists with a dense theme, and leave the default in yaru less dense as it fits shorter lists fine, or what do you think?

Feichtmeier:
sounds good - then maybe also try to include a normal text size inside this local theme (local in terms of locally applied in the tree)?

jpnurmi:
Yeah, given that ListTileThemeData.dense forces a small font, we'll go with ThemeData.visualDensity instead, which looked great in #546.

Issue #483, opened 2021-11-09
Should the theme selection titles be changed?


seb128:
It was pointed on by translators on https://hosted.weblate.org/translate/ubuntu-desktop-installer/ubuntu-desktop-installer/en/?checksum=c6aef533d958e29d#comments

> This phrase (which nowadays has been turned into a meme) must be a placeholder, otherwise it isn't a right fit for the UI of a professional operating system, and should be replaced by something else.

jpnurmi:
I'd be in favor of removing the page altogether. It would be a better fit for a first-run wizard.

jpnurmi:
Alternatively, make it a toggle somewhere in the corner to make it available right away in the beginning. Currently, theme selection comes pretty late in the wizard, when the user is almost done, and dark theme lovers have already hurt their eyes. :sunglasses:

jpnurmi:
One more point for discussion:
> making it configurable for flavors gets actually more complicated because every desktop has their own way of configuring it

yarons:
I put a lot of effort into translating this to Hebrew, I think that the best solution would be adding the "formal" string as a comment and allow the translators to choose their flavors, I find it very poetic.

Muqtxdir:
> I'd be in favor of removing the page altogether. It would be a better fit for a first-run wizard.

I agree with this and moreover the opted theme preference is not remembered after install so the purpose of page is not achieved. 

regarding the description of optioncards for translation s @Feichtmeier could add more to this since he implemented this page

Feichtmeier:
A simple toggle could be enough like in the first section here

![](https://cdn.discordapp.com/attachments/860793709912915988/906213513800867850/unknown.png)

![](https://cdn.discordapp.com/attachments/860793709912915988/906213425695297586/unknown.png)

fitojb:
This issue is closed but the string isn’t removed.

jpnurmi:
@fitojb My bad, I removed the UI elements but forgot the translation strings. Thanks for bringing it up! https://github.com/canonical/ubuntu-desktop-installer/pull/1105

PR #655, opened 2022-02-01
Choose your look: remove theme descriptions

jpnurmi:
They are not suitable for translation.

Close: #483

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
+++ b/packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
@@ -316,8 +316,6 @@
   "chooseYourLookPageHeader": "You can always change this later in the appearance settings.",
   "chooseYourLookPageDarkSetting": "Dark",
   "chooseYourLookPageLightSetting": "Light",
-  "chooseYourLookPageLightBodyText": "Everything is light and bright",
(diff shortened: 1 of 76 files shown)


jpnurmi:
The page is not visible in today's installer. The route was commented out in #404.

oSoMoN (approved):
LGTM, thanks!

elioqoshi:
@jpnurmi This is great, thanks! Would there be any opportunity to change the size of the selection buttons as well?

elioqoshi:
I was just thinking we are not utilizing all the space we have there. Maybe: 
grafik

or

grafik

Feichtmeier:
YaruImageTile could be used for @elioqoshi 's suggestion, even if it is not very hard to create it on your own :)

https://github.com/ubuntu/yaru_widgets.dart/blob/main/lib/src/yaru_image_tile.dart

(even if the selection color should be less transparentized)
grafik
grafik

If yet another dep should be avoided you can ofc just copy paste the code :)

jpnurmi:
@Feichtmeier Looks very nice! Do you want to take this over? I was struggling with the looks and wouldn't mind at all :D

Feichtmeier:
@jpnurmi sure I can :+1: 
Would probably be a good opportunity to finally make a first pub release for yaru_widgets.dart

jpnurmi:
Awesome, thanks!

PR #656, opened 2022-02-01
ChooseYourLook: use yaru_widgets.dart for images

### LINKED ISSUE #483: Should the theme selection titles be changed?

Feichtmeier:
| Before | After with 1/3 of the screenwidth | After with 1/4 of the screen width |
| --- | --- | --- |
|grafik | grafik | grafik |

CC @jpnurmi @elioqoshi please have a look concerning the size of the images
~~I convert the PR to draft because I realized that I use the Yaru* suffix for all widgets in yaru_widgets.dart ... except the ImageTile~~ (fixed)

Closes #483

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
@@ -32,7 +32,7 @@ class ChooseYourLookPage extends StatelessWidget {
                   children: [
                     SizedBox(
                       width: width / 3,
-                      child: ImageTile(
+                      child: YaruImageTile(
                         path: 'assets/Theme_thumbnails-Light.png',
                         currentlySelected:
                             Theme.of(context).brightness == Brightness.light,
@@ -55,7 +55,7 @@ class ChooseYourLookPage extends StatelessWidget {
                   children: [
                     SizedBox(
                       width: width / 3,
-                      child: ImageTile(
+                      child: YaruImageTile(
                         path: 'assets/Theme_thumbnails-Dark.png',
                         currentlySelected:
                             Theme.of(context).brightness == Brightness.dark,
--- a/packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
@@ -14,6 +14,7 @@ class ChooseYourLookPage extends StatelessWidget {
   Widget build(BuildContext context) {
     final lang = AppLocalizations.of(context);
     final width = MediaQuery.of(context).size.width;
+    final settings = Settings.of(context, listen: false);
     return WizardPage(
       header: Text(lang.chooseYourLookPageHeader),
       actions: <WizardAction>[
@@ -36,7 +37,6 @@ class ChooseYourLookPage extends StatelessWidget {
                         currentlySelected:
                             Theme.of(context).brightness == Brightness.light,
                         onTap: () {
-                          final settings = Settings.of(context, listen: false);
                           settings.applyTheme(Brightness.light);
                         },
                       ),
@@ -60,7 +60,6 @@ class ChooseYourLookPage extends StatelessWidget {
                         currentlySelected:
                             Theme.of(context).brightness == Brightness.dark,
                         onTap: () {
-                          final settings = Settings.of(context, listen: false);
                           settings.applyTheme(Brightness.dark);
                         },
                       ),
--- a/packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/material.dart';
 import 'package:ubuntu_wizard/settings.dart';
 import 'package:ubuntu_wizard/widgets.dart';
+import 'package:yaru_widgets/yaru_widgets.dart';
 
 import '../../l10n.dart';
 
@@ -12,6 +13,7 @@ class ChooseYourLookPage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final lang = AppLocalizations.of(context);
+    final width = MediaQuery.of(context).size.width;
     return WizardPage(
       header: Text(lang.chooseYourLookPageHeader),
       actions: <WizardAction>[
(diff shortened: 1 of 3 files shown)


packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
```diff
@@ -12,46 +13,66 @@ class ChooseYourLookPage extends StatelessWidget {
...
-                child: OptionCard(
-                  image: FlavorImage.asset('assets/Theme_thumbnails-Light.png'),
-                  title: Text(lang.chooseYourLookPageLightSetting),
-                  body: Text(lang.chooseYourLookPageLightBodyText),
-                  selected: Theme.of(context).brightness == Brightness.light,
-                  onSelected: () {
-                    final settings = Settings.of(context, listen: false);
-                    settings.applyTheme(Brightness.light);
-                  },
+        child: SingleChildScrollView(
```
jpnurmi, line 26: the scroll view can be dropped because children's sizes are calculated below to make them fit

packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
```diff
@@ -12,46 +13,66 @@ class ChooseYourLookPage extends StatelessWidget {
...
+          child: Row(
+              crossAxisAlignment: CrossAxisAlignment.center,
```
jpnurmi, line 28: Can drop this because Rows' and Columns' cross-axis alignment is already centered by default.

packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
```diff
@@ -12,46 +13,66 @@ class ChooseYourLookPage extends StatelessWidget {
...
+              mainAxisAlignment: MainAxisAlignment.center,
```
jpnurmi, line 29: Add this to make centering work:

          mainAxisSize: MainAxisSize.min,

packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
```diff
@@ -12,46 +13,66 @@ class ChooseYourLookPage extends StatelessWidget {
...
+              children: [
+                Column(
```
jpnurmi, line 31: Add for centering:

    mainAxisSize: MainAxisSize.min,

packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
```diff
@@ -12,46 +13,66 @@ class ChooseYourLookPage extends StatelessWidget {
...
-                  onSelected: () {
-                    final settings = Settings.of(context, listen: false);
-                    settings.applyTheme(Brightness.dark);
-                  },
+                SizedBox(
+                  width: width / 20,
                 ),
-              )
-            ]),
+                Column(
```
jpnurmi, line 54: Add for centering:

    mainAxisSize: MainAxisSize.min,

jpnurmi (commented):
Thanks, @Feichtmeier! It's very nice!

@elioqoshi, which one do you like better; the smaller or larger thumbnails?

packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
Feichtmeier: I thought so too 😃 but unless the installer won't have a min size defined in my_application.cc you can produce overflow if you make the window very small. That's why I've added the scroll view. Should I remove it and make a second PR for the min sizes?

packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
```diff
@@ -12,46 +13,66 @@ class ChooseYourLookPage extends StatelessWidget {
...
+          child: Row(
+              crossAxisAlignment: CrossAxisAlignment.center,
+              mainAxisAlignment: MainAxisAlignment.center,
+              children: [
+                Column(
+                  children: [
+                    SizedBox(
+                      width: width / 3,
+                      child: YaruImageTile(
+                        path: 'assets/Theme_thumbnails-Light.png',
```
jpnurmi, line 36: `FlavorImage.asset` needs to be restored to be able to support flavored installers. I opened a suggestion for YaruImageTile: https://github.com/ubuntu/yaru_widgets.dart/issues/68.

packages/ubuntu_desktop_installer/lib/pages/choose_your_look_page.dart
jpnurmi: The window has a fixed size in production. :smile:

jpnurmi: The window is only resizable in debug mode for testing purposes. Layouts should be reasonably scalable even if the window has a certain fixed size right now. So just like you made it calculate the size. :+1: 

If the preferred size changes in the future, the UI is less likely to require much refactoring. It's easiest to prepare those things while working on the page now, instead of coming back to it one year later. :smile:

Muqtxdir:
@Feichtmeier could you please use the svgs from this branch: https://gitlab.gnome.org/Community/Ubuntu/gnome-control-center/-/merge_requests/5

Feichtmeier:
@Muqtxdir thanks I will!

I am currently implementing @jpnurmi 's suggestion to wrap a Widget rather than always an Image

For the future we could stick to the svg format for assets because they can be easily recolored by flutter svg

jpnurmi:
@Feichtmeier Here are the required test changes with YaruImageTile (needs adapting to YaruSelectionTile once available)
```diff
diff --git a/packages/ubuntu_desktop_installer/test/choose_your_look/choose_your_look_page_test.dart b/packages/ubuntu_desktop_installer/test/choose_your_look/choose_your_look_page_tes>
index 2f5fff0e..915a422e 100644
--- a/packages/ubuntu_desktop_installer/test/choose_your_look/choose_your_look_page_test.dart
+++ b/packages/ubuntu_desktop_installer/test/choose_your_look/choose_your_look_page_test.dart
@@ -7,6 +7,7 @@ import 'package:ubuntu_desktop_installer/l10n.dart';
 import 'package:ubuntu_desktop_installer/pages/choose_your_look_page.dart';
 import 'package:ubuntu_wizard/settings.dart';
 import 'package:ubuntu_wizard/widgets.dart';
+import 'package:yaru_widgets/yaru_widgets.dart';
 
 import '../widget_tester_extensions.dart';
 import 'choose_your_look_page_test.mocks.dart';
@@ -28,20 +29,20 @@ void main() {
       ),
     );
 
-    final lightOptionCard = find.widgetWithText(
-      OptionCard,
-      lang(tester).chooseYourLookPageLightSetting,
+    final lightImageTile = find.widgetWithImage(
+      YaruImageTile,
+      AssetImage('assets/Theme_thumbnails-Light.png'),
     );
     expect(lightImageTile, findsOneWidget);
     await tester.tap(lightImageTile);
     verify(settings.applyTheme(Brightness.light));
 
-    final darkOptionCard = find.widgetWithText(
-      OptionCard,
-      lang(tester).chooseYourLookPageDarkSetting,
+    final darkImageTile = find.widgetWithImage(
+      YaruImageTile,
+      AssetImage('assets/Theme_thumbnails-Dark.png'),
     );
-    expect(darkOptionCard, findsOneWidget);
-    await tester.tap(darkOptionCard);
+    expect(darkImageTile, findsOneWidget);
+    await tester.tap(darkImageTile);
     verify(settings.applyTheme(Brightness.dark));
   });
 }
```

Feichtmeier:
Thanks @jpnurmi - I've included your suggested changes and the test. But I've decided to extract those Columns to a private Widget to make the page more readable, hope this was okay?

Feichtmeier:
> @Feichtmeier could you please use the svgs from this branch: https://gitlab.gnome.org/Community/Ubuntu/gnome-control-center/-/merge_requests/5

Nice updated pictures!
However could we eventually make those previews even more abstract? I think we do not need to draw the buttons in the headerbar, etc. 
Maybe we could just create those things with flutter :thinking: Let me try

Feichtmeier:
> 1/3 of the screenwidth looks best, thanks @Feichtmeier !
> 
> In case we will add a 3rd option in the future (Automatic?) we can adjust the sizes again I guess.

yes :) Here is a faked third option (wrong image)
grafik
we only need to adapt the width to screenwidth / 4 for 3 cards, etc

jpnurmi:
For reference, the final looks with the currently available options:
Screenshot from 2022-02-07 12-02-00

PR #662, opened 2022-02-07
Fix analysis issues for the latest Flutter stable 2.10

jpnurmi:
body_might_complete_normally_nullable:

    info • This function has a nullable return type of 'Future<dynamic>?', but ends without returning a value


oSoMoN (approved):
LGTM, thanks!

PR #665, opened 2022-02-07
Fix slow subiquity startup

jpnurmi:
The socket path that was passed to `_waitSubiquity()` was absolute, not
relative. If the socket was deep enough in the directory hierarchy, the
connect would refuse to establish which caused a 10s startup delay (full
10x wait loop). Yet, it would connect fine after the wait loop because
at that point, a relative path was used.


mwhudson:
Well, TIL!

oSoMoN (approved):
LGTM, thanks!

PR #674, opened 2022-02-10
Show/hide password checkbox

jpnurmi:
A show/hide toggle-button embedded into a password field is problematic
in a few ways that have been discussed in https://github.com/canonical/ubuntu-desktop-installer/pull/626 and https://github.com/canonical/ubuntu-desktop-installer/pull/633. This commit
changes the embedded toggle-button to a separate checkbox placed below
both password fields, similarly to when creating a Google or
Microsoft account, for example.

A summary of the issues this solves:

- Tab focus traversal

  - Either an embedded toggle-button breaks natural tab focus traversal,
    or if made non-focusable to fix tab focus traversal, then it won't
    be accessible by keyboard at all.

    vs.

  - A separate checkbox can be focusable and results in a natural tab
    focus traversal order.

- Look'n'feel

  - It feels strange when the toggle button is embedded inside a specific
    password field, but then it also toggles the visibility of another
    password elsewhere in the confirmation field.

  - Alternatively, if the above issue was fixed by embedding the toggle
    button into both password fields, the UI starts feeling cluttered
    with multiple show/hide password buttons side by side.

    vs.

  - A separate checkbox is not strictly tied/embedded to either password
    field. Yet, it has a clear and easy-to-understand association with
    both password fields.

Screenshot from 2022-02-10 16-12-28

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/l10n/app_localizations.dart
+++ b/packages/ubuntu_desktop_installer/lib/l10n/app_localizations.dart
@@ -1185,6 +1185,12 @@ abstract class AppLocalizations {
   /// **'The passwords do not match'**
   String get whoAreYouPagePasswordMismatch;
 
+  /// No description provided for @whoAreYouPageShowPassword.
+  ///
+  /// In en, this message translates to:
+  /// **'Show password'**
+  String get whoAreYouPageShowPassword;
+
   /// No description provided for @writeChangesToDisk.
   ///
   /// In en, this message translates to:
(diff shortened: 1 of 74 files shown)


jpnurmi:
@elioqoshi, what do you think about using a show password checkbox?

- [ ] Show password

elioqoshi:
I think a checkbox would be a good solution!

oSoMoN (approved):
LGTM, thanks!

PR #686, opened 2022-02-16
CI/generate: swap diff & warnings

jpnurmi:
When you click the details link for the generate step, GitHub
automatically scrolls all the way to the bottom. Showing the diff
first takes you straight to the summary of warnings, which is the
important part of the detailed output.


oSoMoN (approved):
LGTM, thanks!

Issue #698, opened 2022-02-23
Screen 16 - Enable Ubuntu Pro


jpnurmi:
The design for screen n°16, Enable Ubuntu Pro

![image](https://user-images.githubusercontent.com/140617/155300951-5d91d471-6fe2-438b-918d-1be9ab84b4b5.png)

Design details https://people.canonical.com/~platform/desktop/installer/ (TBD)

jpnurmi:
> Ubuntu Pro delivers 10 years of additional security features, including:
>
> - Coverage for over 30,000 packages (extended from 2,500 packages)
> - Kernel Livepatch security patching to avoid unscheduled reboots
> - Ubuntu Security Guide (USG) hardening and compliance profiles 
>
> Ubuntu Pro is **free for personal use on up to 3 machines.**
>
> For more information on how to register, visit [ubuntu.com/pro](http://ubuntu.com/pro)

To enable Ubuntu Pro, add your token or skip this step.

- [ ] Enable Ubuntu Pro now
  __ Your token ______________________________________________________
You can go to [ubuntu.com/pro](http://ubuntu.com/pro) to retrieve your token or register a new one.    

- [ ] Skip this step and upgrade later   
You can enable Ubuntu Pro via Settings or the ‘ua attach’ command.

jpnurmi:
![image](https://user-images.githubusercontent.com/140617/174793889-cc4afcb8-852a-4b3c-9cdd-51d099eb557c.png)

PR #702, opened 2022-02-24
Promote CallbackValidator from ubuntu_wsl_setup to ubuntu_wizard

jpnurmi:
This comes in handy for validating Ubuntu Pro token input (#698).


CarlosNihelton (approved):
LGTM.
The workflow complains about the tests, but honestly, I don't see much to add for testing on this specific subject. The callback validator is very straightforward.

jpnurmi:
Oh, that's interesting. I was thinking that it's fine to move the class because it's implicitly tested by the WSL advanced setup page. However, moving it to another package that does not have its own tests does cause the coverage to be lost. :)

jpnurmi:
I added tests to make the CI happy (and secretly slipped in a tiny change to move another validator to the same place :P).

PR #678, opened 2022-02-14
Bump url_launcher version & fix tests

jpnurmi:
The plugins in the flutter/plugins repo have been moved away from shared
method channels in native plugins. In other words, the channel names
have been renamed from `url_launcher` to `url_launcher_<platform>`.

See flutter/plugins#4719 and flutter/flutter#94224 for details.

PR #703, opened 2022-02-24
Mockable URL launcher

jpnurmi:
As suggested in #678, to avoid assumptions on url_launcher internals.


jpnurmi:
>  92.30% of diff hit (target 93.64%)

argh, the test coverage bar is getting high :D

packages/ubuntu_wizard/lib/src/utils/url_launcher.dart
```diff
@@ -0,0 +1,20 @@
...
+Future<bool> launchUrl(String url) {
+  return getService<UrlLauncher>().launchUrl(url);
+}
+
+class UrlLauncher {
+  Future<bool> launchUrl(String url) async {
+    if (!await canLaunch(url)) {
+      log.error('Unable to launch $url');
+      return false;
+    }
```
oSoMoN, line 17: Would it be useful to log a debug message when launching a URL ?

oSoMoN (approved):
LGTM. Just one very minor suggestion.

packages/ubuntu_wizard/lib/src/utils/url_launcher.dart
jpnurmi: Sure, could be useful :+1:

PR #704, opened 2022-02-25
Speed up allocate_disk_space_page_test

jpnurmi:
For some reason, find.widgetWithText() takes awfully long for finding
an OutlinedButton with specific text in the confirmation dialog's
widget tree. Replacing it with find.text() makes the operation almost
instant. This does the job fine in this context because tap events
propagate from the non-interactive text label to the OutlinedButton
ancestor.

Before, ~20s:

    $ flutter test test/allocate_disk_space
    00:20 +40: All tests passed!

After, ~5s:

    $ flutter test test/allocate_disk_space
    00:05 +40: All tests passed!


oSoMoN (approved):
The gain is even greater on my machine, where the test takes ~38s to run on main, and ~7s with your patch. Very nice!

PR #708, opened 2022-02-25
Ubuntu Pro

jpnurmi:
https://user-images.githubusercontent.com/140617/155756755-d3c933d3-0de5-4a34-bf0c-f689aa8c55b0.mp4

Close: #698


jpnurmi:
The token needs to be copied on the target system. Another use case for the upcoming file copy API in subiquity?

PR #715, opened 2022-03-01
Define SNAP_XXX environment variables for subiquity in dry-run mode

jpnurmi:
The environment was only defined when running the UI, not when running
subiquity_client tests. Therefore, running subiquity_client tests from
VS Code snap would fail:

    FAIL: Don't know how to fake GET response to ('v2/snaps/code', {})


oSoMoN (approved):
LGTM, thanks!

PR #716, opened 2022-03-02
subiquity_client_test: remove assumption on subiquity's location

jpnurmi:
Pass full path to the source catalog so that subiquity can find it
regardless of where subiquity is run from.


oSoMoN (approved):
LGTM, thanks!

PR #722, opened 2022-03-03
Let subiquity decide the initial timezone

jpnurmi:
Request the initial timezone from subiquity, and let it do the GeoIP
lookup instead of looking up geoip.ubuntu.com directly from the UI.

This removes a nice chunk of GeoIP lookup code, and fixes the issue
that the installer did not remember a previously applied timezone, and
fixes the issue that the installer is currently doing geoip lookups when
running integration tests.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
@@ -70,7 +70,6 @@ class WhereAreYouPageState extends State<WhereAreYouPage> {
                               formatLocation(model.selectedLocation);
                         }
                         return TextFormField(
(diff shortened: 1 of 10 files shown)


packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
```diff
@@ -70,7 +70,6 @@ class WhereAreYouPageState extends State<WhereAreYouPage> {
                               formatLocation(model.selectedLocation);
                         }
                         return TextFormField(
-                          autofocus: true,
```
jpnurmi, line 73: in offline mode, the field is initialized to the system locale and no longer left empty

oSoMoN (approved):
LGTM, thanks!

PR #725, opened 2022-03-07
Remove close button in activities overview

jpnurmi:
Even though the close button in the window title was hidden, activities
overview would still show its own close button. This commit gets rid of
that misleading button as the window is not really closable because the
close event is ignored.

### DIFF
--- a/packages/ubuntu_wizard/linux/ubuntu_wizard_plugin.cc
+++ b/packages/ubuntu_wizard/linux/ubuntu_wizard_plugin.cc
@@ -46,16 +46,12 @@ static gboolean set_window_title(GtkWindow* window, const gchar* title) {
 }
 
 static gboolean is_window_closable(GtkWindow* window) {
-  GtkWidget* header_bar = get_header_bar(GTK_WINDOW(window));
-#ifdef HAVE_LIBHANDY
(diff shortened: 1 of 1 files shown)


jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2022-03-07 10-51-21 | Screenshot from 2022-03-07 10-49-29 |

CarlosNihelton (approved):
LGTM.

Honest question: could it matter setting the window as non-deletable after the window being visible in any possible use case for Ubuntu?

jpnurmi:
I haven't noticed any issues with GDM. It works as expected even if I stay in the activities overview during the installation. The close button returns as soon as the installation finishes. There must be a reason why the note was added to the docs, though, so there must be a display manager that doesn't respect the flag. However, it would be merely a cosmetic issue since we ignore the close event anyway.

For reference https://docs.gtk.org/gtk3/method.Window.set_deletable.html

> Depending on the system, this function may not have any effect when called on a window that is already visible, so you should call it before calling gtk_widget_show().

jpnurmi:
I just noticed that the note has been there for 17 years. Perhaps it's not an issue with today's window managers. :thinking: 

https://github.com/GNOME/gtk/commit/adb16c5b89e606ec6293d9e96295ab0b88b655eb

PR #727, opened 2022-03-07
Update subiquity and adapt to changes

jpnurmi:
- canonical/subiquity#1176
- canonical/subiquity#1181 (`storage/v2/guided` payload)

### DIFF
--- a/packages/subiquity_client/lib/src/types.dart
+++ b/packages/subiquity_client/lib/src/types.dart
@@ -216,10 +216,11 @@ bool? _wipeFromString(String? value) =>
 String? _wipeToString(bool? value) =>
     value != null && value == true ? 'superblock' : null;
 
-@freezed
-class Partition with _$Partition {
+@Freezed(unionKey: '\$type', unionValueCase: FreezedUnionCase.pascal)
+class DiskObject with _$DiskObject {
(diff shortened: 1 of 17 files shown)


jpnurmi:
There are a few more issues while updating to the very latest subiquity. I'm turning this into a draft for further investigation.

jpnurmi:
- no permission to write logs
- apt configurer fails in the CI
- setxkbmap fails in the CI

packages/subiquity_client/lib/src/types.dart
```diff
@@ -255,6 +265,8 @@ class Disk with _$Disk {
     String? vendor,
   }) = _Disk;
 
+  List<Partition>? get partitions => objects?.whereType<Partition>().toList();
```
jpnurmi, line 268: Note: The `partitions` list from subiquity includes now both, `Partition` and `Gap` objects, which breaks the UI and a lot of tests. Renaming the mixed object list as `objects` and providing `partitions` for backward compatibility reduces the amount of diff a lot.

oSoMoN: That's very considerate, thanks!

oSoMoN (approved):
LGTM, thanks!

PR #730, opened 2022-03-08
CI: run integration tests

jpnurmi:


CarlosNihelton (approved):
LGTM!

jpnurmi:
I'm fairly confident we have finally reached the level of stability with these tests that enabling them in the CI won't cause too much flakiness. :smile:

Issue #681, opened 2022-02-15
Going back & closing the window allowed before the slides


jpnurmi:
Subiquity starts the installation process when the user presses the "Start installing" button on the "Write changes to disk" page. Once the installation is ongoing and the user proceeds on the next page, the installer should not allow the user to navigate back or close the window. Currently, these actions are only prevented during the installation slides, but not on the Who/Where are you? pages or whatever comes in between.

jpnurmi:
| u-d-i | ubiquity |
|---|---|
| ![image](https://user-images.githubusercontent.com/140617/154051174-f4ff9128-35e4-4a8e-bf3f-6d6cf6216991.png) | ![image](https://user-images.githubusercontent.com/140617/154051080-9f3a2b4f-0e66-48f8-ba46-5003d171cfc1.png) |

PR #734, opened 2022-03-10
Add SubiquityStatusMonitor

jpnurmi:
A simple and standalone, thoroughly tested status monitor that can run in the background and uses its own socket to send (blocking) status change requests to subiquity. Having access to subiquity status, errors, and log IDs has many potential use cases. For starters, it will be used to ensure that the window is closable when appropriate regardless of the page (#681).


jpnurmi:
Included in #736

PR #736, opened 2022-03-11
Prevent closing the window during installation

jpnurmi:
This is a partial fix to #681. Keep the window non-closable, regardless
of the screen, whenever subiquity reports a status that indicates that
installation is ongoing.


packages/ubuntu_wizard/lib/app.dart
```diff
@@ -71,12 +73,16 @@ Future<void> runWizardApp(
...
+    registerServiceInstance(subiquityMonitor);
+  }
 
   WidgetsFlutterBinding.ensureInitialized();
   await setupAppLocalizations();
 
   onWindowClosed().then((_) async {
     await interfaceSettings.close();
+    await subiquityMonitor?.stop();
     await subiquityClient.close();
```
CarlosNihelton, line 86: Hey, @jpnurmi ! This looks very promising! Currently one can quickly go thru the wizard steps for WSL and close the window by pushing the "Finish" button on the last page, causing the Flutter app to end and stop Subiquity in the middle of some task. Could we use this `subiquityMonitor` to show some sort of progress indicator before transitioning to the last page to avoid user closing the window while Subiquity still has work to do?

jpnurmi: That sounds like a great idea!

CarlosNihelton: Do we need to add a new page that automatically transition to `SetupCompletePage` when `subiquityMonitor` reports status DONE ? Would you have a better idea?

CarlosNihelton: > Depends on how you want to visualize the state, I guess. smile Perhaps one option could be to make `SetupCompletePage` visualize the unfinished state and disable the _Finish_ button?

Sounds doable as well. Adding a reference to the `subiquityMonitor` in the `SetupCompleteModel` and a getter to indicate whether it's busy or not plus changing the `SetupCompletePage` to render different contents based on that getter sounds good enough to me.

jpnurmi: Ok, I was wondering about the progress indicator part. If the operation would most often take something like a fraction of a second, then a progress indicator that only momentarily flashes on the screen might not look good. It sounds like it might be worth it entertaining the user with animation in this case, though! :+1:

CarlosNihelton: Here is a dirty POC. We could replace the `Center(child: CircularProgressIndicator())` with a proper widget showing progress and Bob's your uncle! How to you feel about that?

```diff
diff --git a/packages/ubuntu_wsl_setup/lib/main.dart b/packages/ubuntu_wsl_setup/lib/main.dart
index 90c0497..21cc129 100644
--- a/packages/ubuntu_wsl_setup/lib/main.dart
+++ b/packages/ubuntu_wsl_setup/lib/main.dart
@@ -26,6 +26,7 @@ screens, yet allowing user to overwrite any of those during setup.
     serverArgs = ['--prefill', options['prefill']];
   }
   registerService(UrlLauncher.new);
+  registerService(SubiquityStatusMonitor.new);
   runWizardApp(
     ValueListenableBuilder<Variant?>(
       valueListenable: variant,
@@ -39,6 +40,7 @@ screens, yet allowing user to overwrite any of those during setup.
     options: options,
     subiquityClient: SubiquityClient(),
     subiquityServer: SubiquityServer.wsl(),
+    subiquityMonitor: SubiquityStatusMonitor(),
     onInitSubiquity: (client) {
       client.variant().then((value) => variant.value = value);
       client.markConfigured(['filesystem']);
diff --git a/packages/ubuntu_wsl_setup/lib/pages/setup_complete/setup_complete_model.dart b/packages/ubuntu_wsl_setup/lib/pages/setup_complete/setup_complete_model.dart
index b24b7a8..a2a9fa1 100644
--- a/packages/ubuntu_wsl_setup/lib/pages/setup_complete/setup_complete_model.dart
+++ b/packages/ubuntu_wsl_setup/lib/pages/setup_complete/setup_complete_model.dart
@@ -2,16 +2,29 @@ import 'package:flutter/foundation.dart';
 import 'package:subiquity_client/subiquity_client.dart';
 import 'package:ubuntu_wizard/utils.dart';

+extension BusyState on ApplicationState {
+  bool get isInstalling =>
+      index > ApplicationState.STARTING_UP.index &&
+      index < ApplicationState.DONE.index;
+}
+
 /// Implements the business logic of the WSL Setup complete page.
 class SetupCompleteModel extends ChangeNotifier with SystemShutdown {
   /// Creates a setup complete model.
-  SetupCompleteModel(this.client) {
+  SetupCompleteModel(this.client, this.monitor) {
     _identity.addListener(notifyListeners);
+    _isInstalling = monitor.onStatusChanged
+        .map((status) => status?.state?.isInstalling == true);
   }

   @override
   final SubiquityClient client;

+  final SubiquityStatusMonitor monitor;
+
+  late final Stream<bool> _isInstalling;
+  Stream<bool> isInstalling() => _isInstalling;
+
   final _identity = ValueNotifier(const IdentityData());

   /// The username for the profile.
diff --git a/packages/ubuntu_wsl_setup/lib/pages/setup_complete/setup_complete_page.dart b/packages/ubuntu_wsl_setup/lib/pages/setup_complete/setup_complete_page.dart
index 81b74cf..cf73464 100644
--- a/packages/ubuntu_wsl_setup/lib/pages/setup_complete/setup_complete_page.dart
+++ b/packages/ubuntu_wsl_setup/lib/pages/setup_complete/setup_complete_page.dart
@@ -24,8 +24,9 @@ class SetupCompletePage extends StatefulWidget {
   /// Creates an instance with [SetupCompleteModel].
   static Widget create(BuildContext context) {
     final client = getService<SubiquityClient>();
+    final monitor = getService<SubiquityStatusMonitor>();
     return ChangeNotifierProvider(
-      create: (_) => SetupCompleteModel(client),
+      create: (_) => SetupCompleteModel(client, monitor),
       child: const SetupCompletePage(),
     );
   }
@@ -47,30 +48,39 @@ class _SetupCompletePageState extends State<SetupCompletePage> {
   Widget build(BuildContext context) {
     final lang = AppLocalizations.of(context);
     final model = Provider.of<SetupCompleteModel>(context);
-    return WizardPage(
-      title: Text(lang.setupCompleteTitle),
-      header: Text(lang.setupCompleteHeader(model.username)),
-      content: ListView(
-        children: <Widget>[
-          Text(lang.setupCompleteUpdate),
-          const SizedBox(height: 16),
-          const _CodeLabel('\$ sudo apt update'),
-          const SizedBox(height: 16),
-          const _CodeLabel('\$ sudo apt upgrade'),
-          const SizedBox(height: kContentSpacing),
-          Text(lang.setupCompleteManage),
-          const SizedBox(height: 16),
-          const _CodeLabel('\$ sudo ubuntuwsl...'),
-          const SizedBox(height: 56),
-          Text(lang.setupCompleteRestart),
-        ],
-      ),
-      actions: <WizardAction>[
-        WizardAction(
-          label: lang.finishButton,
-          onActivated: () => model.reboot(immediate: false),
-        ),
-      ],
-    );
+    return StreamBuilder<bool>(
+        stream: model.isInstalling(),
+        initialData: true,
+        builder: (context, snapshot) {
+          print(snapshot.data);
+          if (snapshot.data == null || snapshot.data == true) {
+            return Center(child: CircularProgressIndicator());
+          }
+          return WizardPage(
+            title: Text(lang.setupCompleteTitle),
+            header: Text(lang.setupCompleteHeader(model.username)),
+            content: ListView(
+              children: <Widget>[
+                Text(lang.setupCompleteUpdate),
+                const SizedBox(height: 16),
+                const _CodeLabel('\$ sudo apt update'),
+                const SizedBox(height: 16),
+                const _CodeLabel('\$ sudo apt upgrade'),
+                const SizedBox(height: kContentSpacing),
+                Text(lang.setupCompleteManage),
+                const SizedBox(height: 16),
+                const _CodeLabel('\$ sudo ubuntuwsl...'),
+                const SizedBox(height: 56),
+                Text(lang.setupCompleteRestart),
+              ],
+            ),
+            actions: <WizardAction>[
+              WizardAction(
+                label: lang.finishButton,
+                onActivated: () => model.reboot(immediate: false),
+              ),
+            ],
+          );
+        });
   }
 }

```

CarlosNihelton: You are absolutely right. This only proves that we can properly react to the application state.

https://user-images.githubusercontent.com/11138291/159755283-cb7187a2-b890-40e1-a73d-c00b84692961.mp4

CarlosNihelton: Yet the last page says everything is done, so it would be a contradiction to show a progress indicator there. We will need to be more creative. :D

jpnurmi: Oh wow, it takes a lot longer than I had ever imagined! Thanks for the video. Feels almost like the original design is missing a step?

CarlosNihelton: No! I forgot to mention that I added a delay in Subiquity to simulate a poor connection or something alike. Yet it could get as long as that. In dry-run we don't actually download and install anything with apt. Instead we do `apt install --simulate`, which just write logs. So I needed the delay, otherwise it would be blazing fast. For real, though, is not fast enough to ensure users will never close the window before Subiquity finishes its background tasks.

CarlosNihelton: For reference, here is the real experience with subiquity TUI as is, not cuts, no speed ups. Notice we have one more screen there, showing "Applying changes". It took a little while to do apt update and apt install language-pack-fr-base/jammy,now 1:22.04+20211115, language-pack-fr/jammy,now 1:22.04+20220321 and wfrench/jammy,now 1.2.6-1.

jpnurmi: Thanks for the videos. There were efforts to unify the two, so following the same approach as the TUI might be good?

CarlosNihelton: I believe so.

jpnurmi:
@CarlosNihelton Would you be willing to tick this off since you have basically tested the code? :slightly_smiling_face:

CarlosNihelton:
> @CarlosNihelton Would you be willing to tick this off since you have basically tested the code? 🙂

I'll experiment it with the Ubuntu Desktop Installer to have a complete view and submit a review.

CarlosNihelton (approved):
LGTM! We really need to take advantage of `SubiquityMonitor` for the WSL case.

PR #737, opened 2022-03-11
Prevent going back from the Where are you? page

jpnurmi:
At that point, the installation is already ongoing.

Ref: #681


oSoMoN (approved):
LGTM, thanks!

PR #746, opened 2022-03-21
Subiquity subtree

jpnurmi:
To make it possible for flavors to have a git-dependency on the installer (https://github.com/ubuntu-mate/ubuntu-mate-installer/pull/1) that includes the appropriate subiquity revision:
```
dependencies:
  ubuntu_desktop_installer:
    git: ...
    ref: ...
```
If subiquity was a submodule, pub would not fetch it and it would be necessary to manually initialize/update/sync the submodule somewhere deep in the pub cache.

`scripts/update-subiquity` has been updated:
```bash
 ./scripts/update-subiquity 
git subtree pull --prefix packages/subiquity_client/subiquity https://github.com/canonical/subiquity.git main --squash
From https://github.com/canonical/subiquity
 * branch                main       -> FETCH_HEAD
Subtree is already at commit 80c3c7879c18c9ccc4826b34763a08ef9181e82c.
```


jpnurmi:
The idea was not well received. Another option is to bring flavors into the same repo.

PR #749, opened 2022-03-21
subiquity: send --use-os-prober arg

dbungert:
Subiquity is being changed to not use os-prober by default, turn that
back on by way of this argument.


dbungert:
The associated MP on Subiquity has been merged.  Please merge this MP when Subiquity is next updated.

jpnurmi (approved):
Thanks! I included a minimal Subiquity update to canonical/subiquity#1241. The argument is being passed according to the logs and the installation worked. :+1:

PR #752, opened 2022-03-22
Fix slideshow interval

jpnurmi:
1 hour was probably meant to be 1 minute. :) Anyway, Ubiquity is using
50s so use the same. See `SLIDESHOW_OPTIONS = { timeout:50000 }` in:

https://bazaar.launchpad.net/~ubiquity-slideshow/ubiquity-slideshow-ubuntu/html/view/head:/slideshows/ubuntu/slides/index.html#L19


CarlosNihelton (approved):
LGTM.

PR #761, opened 2022-03-29
subiquity_client: increase response logging length limit

jpnurmi:
120 is way too short and results in unhelpful messages like this:

```
DEBUG subiquity_client: ==> status() {"state": "ERROR", "confirming_tty": "/dev/tty1", "error": {"state": "DONE", "base": "1648559745.870598793.install_fail"...
```

It's good to have some limits because some responses are huge, such
as the keyboard layout response. Increasing 10x is long enough to fit
typical storage v2 responses with one disk and a few partitions.


CarlosNihelton (approved):
LGTM!

PR #733, opened 2022-03-09
Updates and other software: restore third-party drivers option

jpnurmi:
There used to be a non-functional "install third-party software" checkbox that was commented out in #373. This PR brings the checkbox back so that it's connected to the "drivers" API of Subiquity. Another checkbox for codecs follows in a separate PR.

image

Design: https://people.canonical.com/~platform/desktop/installer/6.%20Updates%20and%20Other%20Software/

PR #765, opened 2022-03-30
Temporarily mark "drivers" as configured

jpnurmi:
The "drivers" controller doesn't get automatically marked as configured
if available drivers are detected. Explicitly mark it configured until
we have UI (#733) for it, to ensure that the installer doesn't get stuck
on systems where available drivers are detected.


ogayot (approved):
Thanks!

PR #768, opened 2022-03-30
WhoAreYouModel: respect existing values

jpnurmi:
It may happen in integration tests that UI values are entered before
receiving a response from subiquity. If/when that happens, don't reset
the values back to null.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_model.dart
@@ -106,9 +106,9 @@ class WhoAreYouModel extends ChangeNotifier {
   /// Loads the identity data from the server, and resolves the system hostname.
   Future<void> loadIdentity() async {
     final identity = await _client.identity();
-    _realName.value = identity.realname?.orIfEmpty(_realName.value);
-    _hostname.value = identity.hostname?.orIfEmpty(_hostname.value);
-    _username.value = identity.username?.orIfEmpty(_username.value);
+    _realName.value ??= identity.realname?.orIfEmpty(null);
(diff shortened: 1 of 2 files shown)


CarlosNihelton (approved):
LGTM. Are you looking into similar situations for other models as well?

jpnurmi:
The chances are high there are other similar cases

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_model.dart
```diff
@@ -106,9 +106,9 @@ class WhoAreYouModel extends ChangeNotifier {
   /// Loads the identity data from the server, and resolves the system hostname.
   Future<void> loadIdentity() async {
     final identity = await _client.identity();
-    _realName.value = identity.realname?.orIfEmpty(_realName.value);
-    _hostname.value = identity.hostname?.orIfEmpty(_hostname.value);
-    _username.value = identity.username?.orIfEmpty(_username.value);
+    _realName.value ??= identity.realname?.orIfEmpty(null);
+    _hostname.value ??= identity.hostname?.orIfEmpty(null);
+    _username.value ??= identity.username?.orIfEmpty(null);
```
CarlosNihelton, line 111: The `??=` indeed makes more sense. Now we have the existing value, if set, as the top precedence, the result from the client call or null at last. Correct?

CarlosNihelton (approved):
👍

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_model.dart
jpnurmi: Yep! And empty values from the client are kept as null to distinguish between initially empty vs. manually erased empty states.

PR #774, opened 2022-04-01
Fix clunky dropdown focus visuals

jpnurmi:
The default focus highlight from the Material design DropdownButton does not match well with the input decoration that is used to style our dropdowns. Remove the default focus color and let the input decorator highlight focus instead. This way, the focus is consistently highlighted in the same way as input fields do.


jpnurmi:
## Before

https://user-images.githubusercontent.com/140617/161223727-1f12331f-f6a2-4b6f-a5cb-64202c21a700.mp4

## After

https://user-images.githubusercontent.com/140617/161223743-1070a87b-445d-45b7-90a3-b58897074690.mp4

CarlosNihelton (approved):
My first take on this would be just to set theme colors. You had to do a real work here... The end result looks like we would expect. LGTM.

PR #775, opened 2022-04-01
Fix partition dialog layouts

jpnurmi:
The manual tile-height based layouts broke after densities were changed. This PR adds a simple tabular FormLayout with support for configurable row and column spacings to address the problem.

| Before | After |
|---|---|
| Screenshot from 2022-04-01 12-39-53 | Screenshot from 2022-04-01 12-37-21 |
| Screenshot from 2022-04-01 12-40-00 | Screenshot from 2022-04-01 12-37-29 |


CarlosNihelton (approved):
LGTM

PR #782, opened 2022-04-06
Remove old typeOf() workarounds

jpnurmi:
This PR removes the `typeOf<T>()` and `find.byTypeOf()` helpers in favor of the
standard `find.byType(T)`. The restriction on passing generic type arguments was
removed in Dart 2.14.

https://medium.com/dartlang/announcing-dart-2-14-b48b9bb2fb67

> We’ve also removed an old restriction on type arguments, which
> disallowed using generic function types as a type argument. All of
> the following were invalid before 2.14, but are now allowed:

Dart 2.14 was shipped with Flutter 2.5. The current stable versions are
2.16 and 2.10, respectively.


CarlosNihelton (approved):
LGTM!

PR #785, opened 2022-04-06
Fix $SNAP path for WSL

CarlosNihelton:
Since Subiquity server is started by this for WSL, the SNAP environment variable previously set by the preparation scripts was overwritten by the current directory which can even be a Windows directory.

And since the environment herein set was only meant for DRY_RUN, let's check for that before setting it up.


jpnurmi (approved):
Thanks, LGTM :+1:

PR #786, opened 2022-04-07
Prevents user closing the installer window for WSL

CarlosNihelton:
Per our conversation in #736 currently there is no way to prevent users from closing the application during WSL setup while the server is still busy due lack of monitoring before presenting the `SetupCompletePage`.

This draft presents a concept built upon `SubiquityStatusMonitor` that presents a simple wait page  inspired by the current state of the Subiquity TUI application, that prevents going to the last route of the wizard when the server is not done. `ApplicationState.Error` is also interpreted as done for the sake of this page controller.


elioqoshi:
This is a very sensible thing to do and looks good to me!

I'd just check with @jpnurmi  to make sure the loading screen is using a pattern we already use. Is that in line with what is used in the GUI Installer @jpnurmi ?

elioqoshi:
@CarlosNihelton @jpnurmi onm second thought, a loading bar is already used for a determinate waiting time in the installation slide show:

image

But since this is an undetermined time, I'd suggest we use a spinner icon here instead of a loading bar.

packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_page.dart
```diff
@@ -0,0 +1,67 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final model = Provider.of<ApplyingChangesModel>(context, listen: false);
+    final lang = AppLocalizations.of(context);
+    return WizardPage(
+      title: Text(lang.setupCompleteTitle),
+      content: StreamBuilder<bool>(
+          stream: model.isInstalling(),
+          builder: (context, snapshot) {
+            if (snapshot.data == false && previousData == true) {
```
jpnurmi, line 46: this smells like doing business logic in the view :)

packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_page.dart
```diff
@@ -0,0 +1,67 @@
...
+              mainAxisAlignment: MainAxisAlignment.center,
+              children: [
+                const CircularProgressIndicator(),
+                Padding(
+                  padding: const EdgeInsets.only(top: kContentSpacing),
+                  child: Text(lang.applyingChanges),
+                ),
+              ],
+            );
+          }),
```
jpnurmi, line 64: nit: fix indentation

packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_page.dart
```diff
@@ -0,0 +1,67 @@
...
+                });
+                Wizard.of(context).next();
+              });
+            }
+            return Column(
+              mainAxisAlignment: MainAxisAlignment.center,
+              children: [
+                const CircularProgressIndicator(),
+                Padding(
+                  padding: const EdgeInsets.only(top: kContentSpacing),
```
jpnurmi, line 59: the intention is to add spacing between the two, the following is more efficient and perhaps easier to read?
```dart
children: [
  const CircularProgressIndicator(),
  const SizedBox(height: kContentSpacing),
  Text(lang.applyingChanges),
]
```

packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_model.dart
```diff
@@ -0,0 +1,19 @@
...
+import '../../installing_state.dart';
+
+/// Implements the business logic of the WSL Applying Changes Page.
+///
+/// See also:
+///  * [ApplyingChangesPage]
+class ApplyingChangesModel {
+  /// Creates a model for the 'applying changes' page.
+  ApplyingChangesModel(this._monitor) {
+    _isInstalling = _monitor.onStatusChanged
```
jpnurmi, line 12: moving this from the body to the initializer list should allow removing "late"

jpnurmi (commented):
Just to double-check, is the window supposed to be non-closable already on startup? It sounds potentially annoying if there's no way to close the window without going through the wizard.

packages/ubuntu_wsl_setup/lib/main.dart
```diff
@@ -39,8 +41,17 @@ screens, yet allowing user to overwrite any of those during setup.
     options: options,
     subiquityClient: SubiquityClient(),
     subiquityServer: SubiquityServer.wsl(),
+    subiquityMonitor: subiquityMonitor,
     onInitSubiquity: (client) {
-      client.variant().then((value) => variant.value = value);
+      client.variant().then((value) {
+        variant.value = value;
+        if (value == Variant.WSL_SETUP) {
+          setWindowClosable(false);
```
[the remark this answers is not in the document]
CarlosNihelton: It was, but I guess you're right. I'll remove this.

packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_page.dart
CarlosNihelton: Well, you're right. And you might be able to help with understand why this was necessary.

Without this, I get an unhandled exception at the end of the wizard due the last route calling `Wizard.next()` when I click too fast at the exit button. The same behavior can be more easily observed by running the integration tests.

I thought that could be some sort of race condition caused by the Streambuilder scheduling the post frame callback more than once. Lift this logic to the model should be able to avoid the stream to issue `false` more than once, so the post frame callback should never be scheduled more than once. But that's not what happen. Making this widget stateless makes things worse. So apparently the only way to avoid the StreamBuilder to schedule the post frame callback more than once is by keeping some state in the widget.

Even if I do this in the ApplyingChanges in the model, keep the widget stateful but remove the `previousData` from the comparison in the if-statement, I still get the referred exceptions.

Model:
```diff
-  Stream<bool> isInstalling() => _isInstalling;
+  bool previous = true;
+
+  // Returns false only once, after the server first leaving the
+  // [InstallingState] (according to that extension definition).
+  // This mitigates the risk of Wizard.next() being called by the last page due
+  // a fast transition on this stream.
+  Stream<bool> isInstalling() => _isInstalling.map((event) {
+        if (event == false && previous == true) {
+          previous = false;
+          return false;
+        }
+        return true;
+      });

```

View:
```diff
-          if (snapshot.data == false && previousData == true) {
+          if (snapshot.data == false) {
```

The kind of stack trace I got:


```
flutter: DEBUG subiquity_client: GET http://localhost/identity
flutter: DEBUG subiquity_client: ==> identity() {"realname": "u", "username": "u", "crypted_password": "", "hostname": ""}
flutter: INFO system_shutdown: request reboot (immediate: false)
flutter: INFO window: request window close
starting server
flutter: INFO window: window closed
flutter: INFO subiquity_client: Closing socket /home/cnihelton/Dev/Flutter_UDI/main/packages/ubuntu_wsl_setup/../subiquity_client/subiquity/.subiquity/socket
flutter: DEBUG [CN]: Scheduling post frame callback
flutter: DEBUG [CN]: Running the Post frame callback
flutter: ERROR ubuntu_wsl_setup: Unhandled exception
        _AssertionError: 'package:wizard_router/src/wizard.dart': Failed assertion: line 324 pos 14: 'index < widget._routes.length - 1': `Wizard.next()` called from the last route /setup-complete.
\#0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:51:61)
\#1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
\#2      WizardScopeState.next.nextRoute (package:wizard_router/src/wizard.dart:324:14)
\#3      WizardScopeState.next (package:wizard_router/src/wizard.dart:330:25)
\#4      ApplyingChangesPage.build.<anonymous closure>.<anonymous closure> (package:ubuntu_wsl_setup/pages/applying_changes/applying_changes_page.dart:47:36)
\#5      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1144:15)
\#6      SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1089:9)
\#7      SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:995:5)
\#8      _rootRun (dart:async/zone.dart:1426:13)
\#9      _CustomZone.run (dart:async/zone.dart:1328:19)
\#10     _CustomZone.runGuarded (dart:async/zone.dart:1236:7)
\#11     _invoke (dart:ui/hooks.dart:151:10)
\#12     PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:308:5)
\#13     _drawFrame (dart:ui/hooks.dart:115:31)

Lost connection to device.
```

Would you have a better idea other than attempting to replace the StramBuilder by some hand-written similar widget?

packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_page.dart
CarlosNihelton: Great tip!

packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_model.dart
CarlosNihelton: Great tip was well. Even got rid of the monitor reference, since it was only used to initialize this Stream.

packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_page.dart
jpnurmi: In a way, the view model is not doing its job but merely exposing a raw stream and that way pushing the processing logic to the view. What if we'd process the stream in the view model instead, and drop the stream from the view?

```dart
class ApplyingChangesModel extends ChangeNotifier {
  ApplyingChangesModel(this._monitor);

  final SubiquityStatusMonitor _monitor;
  StreamSubscription<ApplicationStatus?>? _sub;

  bool get isInstalling => _monitor.status?.state?.isInstalling == true;

  void init({required VoidCallback onDone}) {
    _sub = _monitor.onStatusChanged.listen((status) {
      if (status?.state == ApplicationState.DONE ||
          status?.state == ApplicationState.ERROR) {
        onDone();
      } else {
        notifyListeners();
      }
    });
  }

  @override
  void dispose() {
    _sub?.cancel();
    super.dispose();
  }
}
```

With this, the widget tree becomes fully declarative. In fact, `isInstalling` is no longer even needed because this view only visualizes that state.
```dart
class _ApplyingChangesPageState extends State<ApplyingChangesPage> {
  @override
  void initState() {
    super.initState();
    final model = Provider.of<ApplyingChangesModel>(context, listen: false);
    model.init(onDone: Wizard.of(context).next);
  }

  @override
    return WizardPage(
      title: Text(lang.setupCompleteTitle),
      content: Column(
        mainAxisAlignment: MainAxisAlignment.center,
        children: [
          const CircularProgressIndicator(),
          const SizedBox(height: kContentSpacing),
          Text(lang.applyingChanges),
        ],
      ),
    );
  }
}
```

CarlosNihelton: You're absolutely right. This approach is still simple enough and prevents the issue I mentioned. I was too biased into using StreamBuilder. Thanks for the input.

CarlosNihelton:
> Hey, I forgot the integration test. Could you update that?

Yeap. I wanted to check whether the window is closable during the period of time the new page is displayed but it appears to me that a new method call would be necessary to just return a bool, which seems overkill.

jpnurmi (approved):

PR #798, opened 2022-04-20
[WSL] Disclaimer for slow internet speeds

CarlosNihelton:
Testing with slow machines revealed that the 'Applying changes' screen can stay for more time than we desire. Thus adding a simple disclaimer about speed of internet connection is a little UX enhancement.

image

A small change for the Subiquity TUI motivated updating the submodule version. I didn't notice any breaking changes while testing locally.

Also removed references to the `ubuntuwsl` command no longer supported nor supplied.


jpnurmi (approved):
LGTM :+1:

PR #807, opened 2022-04-25
Upgrade mockito

jpnurmi:
We have mockito pinned to a specific version to ensure deterministic output. This PR bumps up the version to avoid staying too much behind.


CarlosNihelton (approved):
+1

PR #813, opened 2022-04-26
Installation slides page: allow showing and hiding the logs

jpnurmi:
https://user-images.githubusercontent.com/140617/165674413-2d73dd14-e41d-4299-8e2e-3b7e730da73c.mp4

Ref: #42


CarlosNihelton (approved):
This looks good to me. I missed two things tough, but I'll leave just for your consideration:

1. Integration tests could click on the terminal icon to display the journal, IMHO.
2. I'd expect some marging around the black area of the log view. I'm not sure if this is the final design.

CarlosNihelton:
> I pushed the l10n mass update as a separate commit to make it more pleasant to review.

Indeed it helps a lot.

jpnurmi:
@CarlosNihelton Thanks for the suggestions!

> 1. Integration tests could click on the terminal icon to display the journal, IMHO.

There's already a widget test that tests the terminal button + log view behavior. :smile: We're utilizing integration tests for testing the UI's integration with Subiquity by running different installation variants and verifying the installation results. Showing and hiding the log view is a "UI-only" action that doesn't affect the installation result. Widget tests are faster, less fragile, and more appropriate for testing widget-level features. :wink: 

> 2. I'd expect some marging around the black area of the log view. I'm not sure if this is the final design.

I was experimenting with margins but I couldn't make it look and feel great together with the slide animation and the arrow buttons below. It's a matter of taste, but the variant without margins looks most attractive and feels most integrated into my untrained eyes. Another option could be to make it a modal popup but then it becomes a whole different UI concept because it cannot slide in from the middle of the screen, it needs a close button, and... ugh :grin:

## With margins on all sides:

https://user-images.githubusercontent.com/140617/165458355-a274cf50-2ec9-4a70-8623-b2a84aaaf86f.mp4

## With the bottom margin removed:

https://user-images.githubusercontent.com/140617/165458369-9cb02121-4d56-4d9e-93ae-7aa77f5d404a.mp4

## With all margins removed:

https://user-images.githubusercontent.com/140617/165458376-812e30fa-0e35-426e-b599-eed375b5946e.mp4

What do you think, @elioqoshi?

CarlosNihelton:
> Here's one more variant with buttons taken out:
> 
>  4_no_buttons.mp4

To me this feels better, but it could be even more if the marging background was the same as the window default color, overriding the slideshow completely as you did earlier, but having a container with a background color matching the window's and the log view with the marging you show in `1. With margins on all sides`. The absence of the buttons at this point seems a must to me. I'm curious to see what @elioqoshi thinks about this.

jpnurmi:
> To me this feels better, but it could be even more if the marging background was the same as the window default color, overriding the slideshow completely as you did earlier, but having a container with a background color matching the window's and the log view with the marging you show in `1. With margins on all sides`. The absence of the buttons at this point seems a must to me. I'm curious to see what @elioqoshi thinks about this.

Thanks for clarifying. Great suggestion! Here's a clip of this variant - not bad at all.

https://user-images.githubusercontent.com/140617/165674413-2d73dd14-e41d-4299-8e2e-3b7e730da73c.mp4

CarlosNihelton:
> > To me this feels better, but it could be even more if the marging background was the same as the window default color, overriding the slideshow completely as you did earlier, but having a container with a background color matching the window's and the log view with the marging you show in `1. With margins on all sides`. The absence of the buttons at this point seems a must to me. I'm curious to see what @elioqoshi thinks about this.
> 
> Thanks for clarifying. Great suggestion! Here's a clip of this variant - not bad at all.
> 
> https://user-images.githubusercontent.com/140617/165674413-2d73dd14-e41d-4299-8e2e-3b7e730da73c.mp4

Looks gorgeous to me !!!

jibel:
This is excellent! Thank you both!
Did you get an approval from design?

jpnurmi:
@waynecrosby, we're working on making it possible to show and hide installation logs in the new desktop installer's installation slides screen. We run into some issues and came up with a proposal which is presented below. Does it look ok to you?

# Ubiquity

For reference, this is how the current desktop installer shows and hides logs by expanding and collapsing the window.

https://user-images.githubusercontent.com/140617/166195864-1675c5f8-972d-4e67-a663-636c2dbce2bb.mp4

# New design

The new desktop installer's design specs look very similar to Ubiquity but don't include a state where the logs are visible.

slides

https://people.canonical.com/~platform/desktop/installer/16.%20Installation%20Slides/

# Issues

We started by replicating Ubiquity's behavior in the new installer but run into some issues.

- Unappealing visual artifacts and flickering during window resize.
- When working with lower resolutions, there's no available space to expand the window.
- The number of visible log lines is low even when there's enough space to expand.

# Proposal

We came up with an alternative proposal inspired by the Pop!_OS installer that doesn't resize the window but makes use of the area reserved for the slides above. This avoids issues with visual artifacts and flickering because there's no need to resize the window, and the number of visible log lines is significantly larger, too. What do you think?

https://user-images.githubusercontent.com/140617/165674413-2d73dd14-e41d-4299-8e2e-3b7e730da73c.mp4

elioqoshi:
@jpnurmi As mentioned, this is a great improvement and I couldn't propose a better solution! :)

waynecrosby:
@jpnurmi I think the solution works great, I would just try and keep the contents within the installer window away from the Suru on the right hand side as much as possible. Thanks

PR #817, opened 2022-04-29
WSL: show "applying changes..." at the end of reconfiguration

jpnurmi:
Same as #786 but for reconfiguration mode.

This solves the issue found by the integration tests that Subiquity
would get killed before it was done with writing the config.


CarlosNihelton (approved):
LGTM! Indeed in a slow virtual machine the server died at different points of execution of the configure controller. That's a very sane way to go.

PR #818, opened 2022-05-03
Temporarily disables the profile page hyperlink.

CarlosNihelton:
That hyperlink renders and reacts correclty but there are no browsers available at the context where user would click the link, so it fails to launch the webpage. I took an approach that favors easily roll back in part to avoid translators to spend time reworking he strings, but also because we expect to make it work soon.

Here is how it looks like with the proposed changes:

image

This closes https://github.com/ubuntu/WSL/issues/173.


jpnurmi (approved):
LGTM :+1:

PR #829, opened 2022-05-12
Prevents UI to translate itself for languages lacking font support in WSL

CarlosNihelton:
There are a couple of languges that require adding specific font packages to better support them for WSL. Unfortunately we didn't detect it earlier and the lack of such fonts is causing issues to users. Adding the required fonts will increase too much the image size, so the short term approach is to prevent the UI from translating itself into any of those languages. They will still show up in the "Select your language" page and selecting any of those will still be meaningfull for Subiquity, so they must be preserved.

With the following changes we go from this (notice the first line and last two):
image

into this:
image

This closes https://github.com/ubuntu/WSL/issues/172

### DIFF
--- a/packages/ubuntu_wsl_setup/test/select_language_model_test.dart
+++ b/packages/ubuntu_wsl_setup/test/select_language_model_test.dart
@@ -81,25 +81,15 @@ void main() {
     final model = SelectLanguageModel(MockSubiquityClient());
     await model.loadLanguages();
     final languages = List.generate(model.languageCount, model.language);
-    var blacklisted = 'Uyghur';
-    var index = languages.indexOf(blacklisted);
-    var loc = model.uiLocale(index);
-    expect(loc, kDefaultLocale);
-    blacklisted = 'Sinhala';
-    index = languages.indexOf(blacklisted);
-    loc = model.uiLocale(index);
-    expect(loc, kDefaultLocale);
+    expect(model.uiLocale(languages.indexOf('Uyghur')), kDefaultLocale);
+    expect(model.uiLocale(languages.indexOf('Sinhala')), kDefaultLocale);
   });
 
   test('language wont return unsupported chars', () async {
     final model = SelectLanguageModel(MockSubiquityClient());
     await model.loadLanguages();
     final languages = List.generate(model.languageCount, model.language);
-    var blacklisted = '\u{0626}';
-    var index = languages.indexWhere((lang) => lang.contains(blacklisted));
-    expect(index, -1);
-    blacklisted = '\u{0DC3}';
-    index = languages.indexWhere((element) => element.contains(blacklisted));
-    expect(index, -1);
+    expect(languages.any((e) => e.contains('\u{0626}')), isFalse);
+    expect(languages.any((e) => e.contains('\u{0DC3}')), isFalse);
   });
 }
--- a/packages/ubuntu_wsl_setup/test/select_language_model_test.dart
+++ b/packages/ubuntu_wsl_setup/test/select_language_model_test.dart
@@ -76,4 +76,30 @@ void main() {
     expect(model.selectedLanguageIndex, equals(1));
     expect(wasNotified, isTrue);
   });
+
+  test('uiLocale returns default for unsupported langs', () async {
+    final model = SelectLanguageModel(MockSubiquityClient());
+    await model.loadLanguages();
+    final languages = List.generate(model.languageCount, model.language);
+    var blacklisted = 'Uyghur';
+    var index = languages.indexOf(blacklisted);
+    var loc = model.uiLocale(index);
+    expect(loc, kDefaultLocale);
+    blacklisted = 'Sinhala';
+    index = languages.indexOf(blacklisted);
+    loc = model.uiLocale(index);
+    expect(loc, kDefaultLocale);
+  });
+
+  test('language wont return unsupported chars', () async {
+    final model = SelectLanguageModel(MockSubiquityClient());
+    await model.loadLanguages();
+    final languages = List.generate(model.languageCount, model.language);
+    var blacklisted = '\u{0626}';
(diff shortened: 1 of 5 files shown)


jpnurmi:
You could get away without any code changes if you'd remove, rename, or move the undesired ARB files in `ubuntu_wsl_setup/lib/l10n` and re-run `flutter gen-l10n`.

a) `rm app_xx.arb`, or
b) `mv app_xx.arb  app_xx.arb.bak`, or
c) `mkdir bak && mv app_xx.arb bak/`

CarlosNihelton:
> You could get away without any code changes if you'd remove, rename, or move the undesired ARB files in `ubuntu_wsl_setup/lib/l10n` and re-run `flutter gen-l10n`.
> 
> a) `rm app_xx.arb`, or b) `mv app_xx.arb app_xx.arb.bak`, or c) `mkdir bak && mv app_xx.arb bak/`

Well, if I do that the referred language disappears from the selection list. It's desirable to preserve it, because Subiquity will apply locale and install language packs for that language once we finish.

packages/ubuntu_wsl_setup/test/select_language_model_test.dart
```diff
@@ -76,4 +76,30 @@ void main() {
...
+  test('language wont return unsupported chars', () async {
+    final model = SelectLanguageModel(MockSubiquityClient());
+    await model.loadLanguages();
+    final languages = List.generate(model.languageCount, model.language);
+    var blacklisted = '\u{0626}';
+    var index = languages.indexWhere((lang) => lang.contains(blacklisted));
+    expect(index, -1);
+    blacklisted = '\u{0DC3}';
+    index = languages.indexWhere((element) => element.contains(blacklisted));
+    expect(index, -1);
```
jpnurmi, line 103:
```suggestion
    expect(languages.any((e) => e.contains('\u{0626}')), isFalse);
    expect(languages.any((e) => e.contains('\u{0DC3}')), isFalse);
```

packages/ubuntu_wsl_setup/test/select_language_model_test.dart
```diff
@@ -76,4 +76,30 @@ void main() {
...
+    await model.loadLanguages();
+    final languages = List.generate(model.languageCount, model.language);
+    var blacklisted = 'Uyghur';
+    var index = languages.indexOf(blacklisted);
+    var loc = model.uiLocale(index);
+    expect(loc, kDefaultLocale);
+    blacklisted = 'Sinhala';
+    index = languages.indexOf(blacklisted);
+    loc = model.uiLocale(index);
+    expect(loc, kDefaultLocale);
```
jpnurmi, line 91:
```suggestion
    expect(model.uiLocale(languages.indexOf('Uyghur')), kDefaultLocale);
    expect(model.uiLocale(languages.indexOf('Sinhala')), kDefaultLocale);
```

jpnurmi (approved):
Ah, I should've read the description more carefully, sorry. LGTM! :+1:

packages/ubuntu_wsl_setup/test/select_language_model_test.dart
CarlosNihelton: Thanks for the suggestion!

packages/ubuntu_wsl_setup/test/select_language_model_test.dart
CarlosNihelton: Thanks for the suggestion!

jpnurmi:
It looks like this change broke automatic scrolling to the selected language on app startup. The selected logical language index doesn't match with visual UI language index?

CarlosNihelton:
The list is reordered based on the new display name.

jpnurmi:
I was looking into updating scroll_to_index because there's a new version available that adapts to the non-nullable WidgetsBinding.instance API change in Flutter but then I noticed that WSL no longer scrolls on startup:

| Before d39951e | After d39951e |
|---|---|
| Screenshot from 2022-05-16 12-08-11 | Screenshot from 2022-05-16 12-07-33 |

CarlosNihelton:
I'm looking into it.

CarlosNihelton:
@jpnurmi everything is behaving as expected by checking out current `main`. It's scrolling to match my system's language as it should.

jpnurmi:
Ok, so auto-scrolling didn't really work as expected on my system even before that commit. English just happened to fit on the screen by default. :)

CarlosNihelton:
Do you see why it doesn't work in your system but works on mine? English is my default too.

jpnurmi:
For me, it bails out because of the "server locale == UI locale" condition and doesn't even try to scroll: https://github.com/canonical/ubuntu-desktop-installer/pull/839

Is it reproducible with this?
```
LANG="en_US.UTF-8" LANGUAGE="en_US:en" flutter run
```

jpnurmi:
There's some locale detection and parsing involved on multiple levels including Subiquity, GTK, Flutter, Dart, and the UI so I'm not sure what exactly is needed to reproduce this, but it's bailing out because "en_US == en_US" and then doesn't scroll at all.

CarlosNihelton:
> 

Good catch. It selects English, but won't scroll.

jpnurmi:
> Before [d39951e](https://github.com/canonical/ubuntu-desktop-installer/commit/d39951eabd177d64bdbffbcb2bb03fcd81797c7c)	After [d39951e](https://github.com/canonical/ubuntu-desktop-installer/commit/d39951eabd177d64bdbffbcb2bb03fcd81797c7c)
> Screenshot from 2022-05-16 12-08-11	Screenshot from 2022-05-16 12-07-33

My bad, d39951e is not to blame. These are both wrong. :) We are passing `AutoScrollPosition.middle` so if scrolling had worked, English would've been scrolled to the exact center on the left. It was a pure coincidence that English was on the screen by default.

PR #833, opened 2022-05-13
Use SafeChangeNotifier for all view models

jpnurmi:
[SafeChangeNotifier](https://pub.dev/packages/safe_change_notifier) was developed to protect from the common issue that
view models may get disposed while waiting for an async operation to
complete. Unlike the built-in ChangeNotifier that throws assertion
errors, SafeChangeNotifier handles the situation gracefully by simply
not notifying the listeners if called after disposal.

This change is motivated by an attempt to fix the recently disabled
`use_build_context_synchronously` lint failures.

P.S. SafeChangeNotifier is not a new dependency but is already used in the "connect the internet" module. :)


CarlosNihelton (approved):
That SafeChangeNotifier sounds interesting!

PR #841, opened 2022-05-16
Remove superfluous KeyboardService

jpnurmi:
Keyboard layout loading used to be a much more complicated process that
involved non-trivial data structures. The KeyboardService class (or
KeyboardModel as it was called then) was introduced to help with that
complexity.

These days, loading the layouts is so straightforward that a separate
KeyboardService no longer serves a purpose. Any view model can get the
list of layouts by calling SubiquityClient.keyboard(), which returns a
KeyboardSetup that contains a List&lt;KeyboardLayout&gt; that can be used as
is. There's no parsing, tuples, or maps involved anymore, where some
help was needed. :)


CarlosNihelton (approved):
LGTM

PR #848, opened 2022-05-17
Split wizard actions into two phases - activation & execution 

jpnurmi:
The activation step allows pages to perform asynchronous tasks such as
sending information to Subiquity. The execution step gets only called if
the context is still valid after the asynchronous activation step. This
ensures that e.g. `Wizard.of(context).next()` is not incorrectly called
after navigating back.

This is a partial fix to issues exposed by the recently enabled
`use_build_context_synchronously` lint.


CarlosNihelton (approved):
This looks great! The onNext and onDone wizard API's make the code look lean and straight to the point.

PR #850, opened 2022-05-18
[WSL] Server side username validation

CarlosNihelton:
With the merge of #852 we can validate usernames with the server. The intent behind it is centralize the logic to prevent users from selecting usernames reserved for the system. For WSL specifically, since the installer runs on the target, there is also the possibility of the username already exists.

So, with this PR we make use of that feature in the profile setup model.

Since WSL installer has fewer pages, I opted for pushing it first. Easier to test. Should the current approach be successfull I'll push the modifications I already experimented with for U-D-I. The end result can be seen below:

https://user-images.githubusercontent.com/11138291/169035544-deaa9eee-9478-4295-bfb1-1271463d5e4a.mp4

This closes https://github.com/ubuntu/WSL/issues/179

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/pages/profile_setup/profile_setup_model.dart
+++ b/packages/ubuntu_wsl_setup/lib/pages/profile_setup/profile_setup_model.dart
@@ -1,7 +1,9 @@
 import 'package:flutter/foundation.dart';
+import 'package:form_field_validator/form_field_validator.dart';
 import 'package:safe_change_notifier/safe_change_notifier.dart';
 import 'package:subiquity_client/subiquity_client.dart';
 import 'package:ubuntu_wizard/utils.dart';
+import 'package:ubuntu_wizard/widgets.dart';
 
 /// The regular expression pattern for valid usernames:
 /// - must start with a lowercase letter
@@ -15,6 +17,7 @@ class ProfileSetupModel extends SafeChangeNotifier {
     Listenable.merge([
       _realname,
       _username,
+      _usernameValidation,
       _password,
       _confirmedPassword,
       _showAdvanced,
@@ -24,6 +27,8 @@ class ProfileSetupModel extends SafeChangeNotifier {
   final SubiquityClient _client;
   final _realname = ValueNotifier<String?>(null);
   final _username = ValueNotifier<String?>(null);
+  final _usernameValidation =
+      ValueNotifier<UsernameValidation>(UsernameValidation.OK);
   final _password = ValueNotifier('');
   final _confirmedPassword = ValueNotifier('');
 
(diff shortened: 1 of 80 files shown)


jpnurmi:
Would you mind pushing the subiquity update and client changes + tests in a separate PR? I'll promise to review it quickly so we'll get it out of the way. :)

CarlosNihelton:
> Would you mind pushing the subiquity update and client changes + tests in a separate PR? I'll promise to review it quickly so we'll get it out of the way. :)

Sorry! Just now I notice this comment. I'll do it.

CarlosNihelton:
The business resides in the last commit. The previous two are related to l10n only.

packages/ubuntu_wsl_setup/lib/pages/profile_setup/profile_setup_widgets.dart
```diff
@@ -41,6 +41,7 @@ class _UsernameFormField extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final lang = AppLocalizations.of(context);
+    final model = context.read<ProfileSetupModel>();
     final username =
         context.select<ProfileSetupModel, String>((model) => model.username);
```
jpnurmi, line 47: we could use the same selective rebuild approach?
```dart
final validation = context.select<..., UsernameValidation>(model) => model.usernameValidation)
```

packages/ubuntu_wsl_setup/lib/pages/profile_setup/profile_setup_widgets.dart
```diff
@@ -55,11 +56,13 @@ class _UsernameFormField extends StatelessWidget {
         PatternValidator(
           kValidUsernamePattern,
           errorText: lang.profileSetupUsernameInvalid,
-        )
+        ),
+        ...model.validators,
```
jpnurmi, line 60: ...and then together with an extension for localizing enum values, `...model.validators` could be replaced by a single validator that checks the current validation status from the model?

```dart
CallbackValidator(
  (_) => model.usernameOk,
  errorText: validation.localize(lang),
),
```

packages/ubuntu_wsl_setup/lib/pages/profile_setup/profile_setup_model.dart
```diff
@@ -64,10 +69,43 @@ class ProfileSetupModel extends SafeChangeNotifier {
...
+    }
+    errorValueToText.forEach((key, value) {
+      _validators.add(
+        CallbackValidator(
+          (_) => usernameValidation != key,
+          errorText: value,
+        ),
+      );
+    });
+  }
```
jpnurmi, line 107: I'd propose replacing this with a CallbackValidator placed in the username field widget to avoid widget dependencies in the model.

packages/ubuntu_wsl_setup/lib/pages/profile_setup/profile_setup_page.dart
```diff
@@ -42,6 +42,19 @@ class _ProfileSetupPageState extends State<ProfileSetupPage> {
...
+        {
+          UsernameValidation.ALREADY_IN_USE: lang.profileSetupUsernameInUse,
+          UsernameValidation.SYSTEM_RESERVED:
+              lang.profileSetupUsernameSystemReserved,
+          UsernameValidation.INVALID_CHARS:
+              lang.profileSetupUsernameInvalidChars,
+          UsernameValidation.TOO_LONG: lang.profileSetupUsernameTooLong,
+        },
+      );
+    });
```
jpnurmi, line 57: enum extensions work great for localizing enum values. for example, https://github.com/canonical/ubuntu-desktop-installer/blob/main/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart#L129

jpnurmi (commented):
great work! :+1: i'd like to propose restructuring the code so that we can eliminate widget dependencies from the view model :)

packages/ubuntu_wsl_setup/lib/pages/profile_setup/profile_setup_widgets.dart
CarlosNihelton: That suggestion blew my mind! Got where I wanted in a much cleaner way. Thank you very much for the suggestion.

jpnurmi (approved):
So nice to have proper validation! LGTM :+1:

PR #852, opened 2022-05-18
Subiquity username validation endpoint

cnihelton:
Since https://github.com/canonical/subiquity/pull/1296 Subiquity offers an endpoint to validate usernames against a set of names reserved for system usage, inspired by Debian installer, but baked into the server.

This PR brings knowledge of this endpoint to the client.

### DIFF
--- a/packages/subiquity_client/lib/src/client.dart
+++ b/packages/subiquity_client/lib/src/client.dart
@@ -29,12 +29,6 @@ extension VariantString on Variant {
   String toVariantString() => name.toLowerCase();
 }
 
-extension UVString on UsernameValidation {
-  static UsernameValidation fromString(String value) {
-    return UsernameValidation.values.firstWhere((v) => value == v.name);
-  }
-}
-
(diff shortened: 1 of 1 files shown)


packages/subiquity_client/lib/src/client.dart
```diff
@@ -208,6 +214,21 @@ class SubiquityClient {
...
+      Uri.http(
+        'localhost',
+        'identity/validate_username',
+        {'username': '"$username"'},
+      ),
+    );
+    final response = await request.close();
+
+    final respStr = await _receive("identity/validate_username()", response);
+    return UVString.fromString(respStr.removePrefix('"').removeSuffix('"'));
```
jpnurmi, line 229: you can replace `UVString.fromString()` with `UsernameValidation.values.byName()`

https://api.dart.dev/stable/2.17.1/dart-core/EnumByName.html

jpnurmi (commented):
we can get rid of the extension but otherwise looks good! :+1:

PR #856, opened 2022-05-19
[UDI] Server side username validation

CarlosNihelton:
Following the same structure of #850 here we leverage the server endpoint for username validation on Ubuntu Desktop Installer.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
+++ b/packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
@@ -446,4 +446,3 @@
   "copyingFiles": "Copying files...",
   "installationFailed": "Installation failed"
 }
-}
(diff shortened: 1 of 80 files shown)


packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
```diff
@@ -441,4 +445,4 @@
   "supportedSoftware": "Supported software",
   "copyingFiles": "Copying files...",
   "installationFailed": "Installation failed"
-}
```
jpnurmi, line 444: The Flutter Intl extension for VS Code provides support for ARB files and insists on removing this line break. :)

jpnurmi (approved):
LGTM :+1:

packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
CarlosNihelton: My Vim keeps adding it. Every tool, every choice. :smile:

Issue #412, opened 2021-10-08
Sync client data classes with Subiquity's


jpnurmi:
All fields in `subiquity_client/lib/src/types.dart` are nullable types. These should be synced with `subiquity/common/types.py`.

Using nullable types for fields that cannot be null in subiquity
a) gives a false sense of safety in the UI because they cannot be null in subiquity so it will throw when passing null defaults, and
b) clutters the UI code with unnecessary null-aware operators and repeated defaults.

It would be best to define those defaults in one place, in `types.dart`, and keep that in sync with subiquity's expectations.

Related to #411.

jpnurmi:
September 29, 2021, dbungert:
> in general, the concept of which values you have to send has been revisited on the API. in apidef.py, you might see a field marked Optional. This is some unfortunate syntax which means that it can be assigned a null as a value. The true values that can be omitted from the incoming json data are marked with a default value
> example - guided choice
> ```python
> class GuidedChoice:
>     disk_id: str                                                                                                                  
>     use_lvm: bool = False                                                                                                         
>     password: Optional[str] = attr.ib(default=None, repr=False)
> ```
> the only required value is disk_id.  This was always how it was intended to work but a bug in the deserializer prevented that and clients had to send all fields.

PR #858, opened 2022-05-20
Sync IdentityData with Subiquity

jpnurmi:
All the IdentityData fields are non-nullable i.e. none of them is
Optional in Subiquity. Making every field nullable only gives a false
sense of safety and clutters the UI code with unnecessary null-aware
operators and repeated imaginary defaults that are never used because
the values coming from Subiquity cannot be null.

Ref: https://github.com/canonical/ubuntu-desktop-installer/issues/412


jpnurmi:
https://github.com/canonical/subiquity/blob/a97333bebb4c4ddaf78a43d9c540667b68d15f1e/subiquity/common/types.py#L351-L356

CarlosNihelton (approved):
Good work!

PR #859, opened 2022-05-20
Sync keyboard detection steps with Subiquity

jpnurmi:
All the StepPressKey, StepKeyPresent, and StepResult fields are non-
nullable i.e. none of them is Optional in Subiquity. Making every field
nullable only gives a false sense of safety and clutters the UI code
with unnecessary null-aware operators and repeated imaginary defaults
that are never used because the values coming from Subiquity cannot be
null.

Ref: https://github.com/canonical/ubuntu-desktop-installer/issues/412

### DIFF
--- a/packages/subiquity_client/test/subiquity_client_test.dart
+++ b/packages/subiquity_client/test/subiquity_client_test.dart
@@ -427,9 +427,6 @@ void main() {
       final stepPressKey = step0 as StepPressKey;
       expect(stepPressKey.symbols, isNotEmpty);
       expect(stepPressKey.keycodes, isNotEmpty);
-      expect(stepPressKey.keycodes.first, hasLength(2));
-      expect(stepPressKey.keycodes.first.first, isA<int>());
-      expect(stepPressKey.keycodes.first.last, isA<String>());
 
       // assumes that step 2 is a result step
       final step2 = await client.getKeyboardStep('2');
--- a/packages/subiquity_client/test/subiquity_client_test.dart
+++ b/packages/subiquity_client/test/subiquity_client_test.dart
@@ -427,9 +427,9 @@ void main() {
       final stepPressKey = step0 as StepPressKey;
       expect(stepPressKey.symbols, isNotEmpty);
       expect(stepPressKey.keycodes, isNotEmpty);
-      expect(stepPressKey.keycodes!.first, hasLength(2));
-      expect(stepPressKey.keycodes!.first.first, isA<int>());
-      expect(stepPressKey.keycodes!.first.last, isA<String>());
+      expect(stepPressKey.keycodes.first, hasLength(2));
+      expect(stepPressKey.keycodes.first.first, isA<int>());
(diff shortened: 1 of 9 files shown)


jpnurmi:
https://github.com/canonical/subiquity/blob/a97333bebb4c4ddaf78a43d9c540667b68d15f1e/subiquity/common/types.py#L121-L143
```python
@attr.s(auto_attribs=True)
class StepPressKey:
    # "Press a key with one of the following symbols"
    symbols: List[str]
    keycodes: Dict[int, str]


@attr.s(auto_attribs=True)
class StepKeyPresent:
    # "Is this symbol present on your keyboard"
    symbol: str
    yes: str
    no: str


@attr.s(auto_attribs=True)
class StepResult:
    # "This is the autodetected layout"
    layout: str
    variant: str


AnyStep = Union[StepPressKey, StepKeyPresent, StepResult]
```

jpnurmi:
This needs more work. Keycodes are `Dict[int, str]` vs. `List<List<dynamic>>`. :slightly_frowning_face: 

It seems that the conversion from `Dict[int, str]` to `List[List[int, str]]` is intentional in Subiquity. The problem is just that a list of mixed int and string types becomes `List<dynamic>` in Dart. This means that `Dict[int, str]` turns to `List<List<dynamic>>` which is too far from the "expected" type which would be `Map<int, String>`.

https://github.com/canonical/subiquity/blob/a97333bebb4c4ddaf78a43d9c540667b68d15f1e/subiquity/common/serialize.py#L92-L108

```py
    def _walk_Dict(self, meth, args, value, m, p, serializing):
        k_ann, v_ann = args
        if k_ann is str:
            return {
                meth(k_ann, k, m, f'{p}/{k}'): meth(v_ann, v, m, f'{p}[{k}]')
                for k, v in value.items()
                }
        elif serializing:
            return [
                [meth(k_ann, k, m, f'{p}/{k}'), meth(v_ann, v, m, f'{p}[{k}]')]
                for k, v in value.items()
                ]
        else:
            return dict([
                (meth(k_ann, k, m, f'{p}/{k}'), meth(v_ann, v, m, f'{p}[{k}]'))
                for k, v in value
                ])
```

packages/subiquity_client/test/subiquity_client_test.dart
```diff
@@ -427,9 +427,6 @@ void main() {
       final stepPressKey = step0 as StepPressKey;
       expect(stepPressKey.symbols, isNotEmpty);
       expect(stepPressKey.keycodes, isNotEmpty);
-      expect(stepPressKey.keycodes!.first, hasLength(2));
-      expect(stepPressKey.keycodes!.first.first, isA<int>());
-      expect(stepPressKey.keycodes!.first.last, isA<String>());
```
jpnurmi, line 432: `stepPressKey.keycodes` was previously `List<List<dynamic>>` and this code tested that it had int and String elements. Now, `stepPressKey` is `Map<int, String>` so testing the types is no longer necessary.

CarlosNihelton (approved):
LGTM.

PR #862, opened 2022-05-20
Removes ProxyAssetBundle dependency on /proc/self/exe symbolic link.

CarlosNihelton:
Hello! While doing my research I stumbled accross this class with a big note saying that it only works on Linux due dependency on `/proc/self/exe` symlink. It turns out Dart offers a suitable API for the resolved path of the app executable: `Platform.resolvedExecutable`. If I understood correctly the intent here, the following set of modifications makes this class portable. To prove the concept, I took a screenshot of the same codebase running Linux (over WSL) and Windows versions (compiled in release mode).

portable_proxyassetbundle


jpnurmi (approved):
Thanks, LGTM! :+1:

PR #867, opened 2022-05-23
Prevents window getting unclosable on server exit

CarlosNihelton:
Hello! I'm not quite sure how, but latest Windows builds revealed a race condition on the window getting non-closable after Subiquity server exiting. Checking Subiquity status monitor current state for null and allowing the window to be closed in that situation prevents the window from getting unclosable after the server reporting DONE (or ERROR) and exiting.

Closes https://github.com/ubuntu/WSL/issues/201

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/main.dart
+++ b/packages/ubuntu_wsl_setup/lib/main.dart
@@ -47,7 +47,9 @@ screens, yet allowing user to overwrite any of those during setup.
         variant.value = value;
         if (value == Variant.WSL_SETUP) {
           subiquityMonitor.onStatusChanged.listen((status) {
-            setWindowClosable(status?.state?.isInstalling == false);
+            final closable =
+                status == null || status.state?.isInstalling == false;
+            setWindowClosable(closable);
           });
         }
       });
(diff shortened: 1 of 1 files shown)


packages/ubuntu_wsl_setup/lib/main.dart
```diff
@@ -47,7 +47,9 @@ screens, yet allowing user to overwrite any of those during setup.
         variant.value = value;
         if (value == Variant.WSL_SETUP) {
           subiquityMonitor.onStatusChanged.listen((status) {
-            setWindowClosable(status?.state?.isInstalling == false);
+            final closable =
+                status == null || status.state?.isInstalling == false;
```
jpnurmi, line 51: both `status` and `status.state` are nullable. this would still make the window non-closable if `status.state` is null. it would be safer to use `status?.state?.isInstalling != true` because then it wouldn't matter if either of them is null.

CarlosNihelton: You're right. Thanks for that!

jpnurmi (approved):
LGTM :+1:

PR #871, opened 2022-05-24
Sync source selection classes with Subiquity

jpnurmi:
Ref: #412


jpnurmi:
https://github.com/canonical/subiquity/blob/a97333bebb4c4ddaf78a43d9c540667b68d15f1e/subiquity/common/types.py#L172-L186

```py
@attr.s(auto_attribs=True)
class SourceSelection:
    name: str
    description: str
    id: str
    size: int
    variant: str
    default: bool


@attr.s(auto_attribs=True)
class SourceSelectionAndSetting:
    sources: List[SourceSelection]
    current_id: str
    search_drivers: bool
```

PR #874, opened 2022-05-24
IntegrationTester.pumpUntil() return immediately if found

jpnurmi:
Don't enter the async loop at all if the condition is already met.


CarlosNihelton (approved):
This looks even more promising.

PR #879, opened 2022-05-25
Prepare tests for Disk API changes

jpnurmi:
This helps to take the next step, which is to make the majority of Disk
arguments required and non-nullable to match the Subiquity API. With a
centralized helper method we can stick some dummy defaults in one place
instead of repeating them all over the place.

Ref: #412


CarlosNihelton (approved):
LGTM.

PR #882, opened 2022-05-27
Sync Disk API with Subiquity 

jpnurmi:
Ref: #412


jpnurmi:
https://github.com/canonical/subiquity/blob/a97333bebb4c4ddaf78a43d9c540667b68d15f1e/subiquity/common/types.py#L281-L295

```py
@attr.s(auto_attribs=True)
class Disk:
    id: str
    label: str
    type: str
    size: int
    usage_labels: List[str]
    partitions: List[Union[Partition, Gap]]
    ok_for_guided: bool
    ptable: Optional[str]
    preserve: bool
    path: Optional[str]
    boot_device: bool
    model: Optional[str] = None
    vendor: Optional[str] = None
```

CarlosNihelton (approved):
LGTM

PR #888, opened 2022-05-30
Specify JsonSerializable() defaults in build.yaml

jpnurmi:
Apply the defaults once in build.yaml instead of repeating for every
generated class. The generated types.g.dart is identical. This cleans
up types.dart in a way that makes it much nicer to generate later.


CarlosNihelton (approved):
Nice! I didn't know about this possibility with JsonSerializable.

PR #889, opened 2022-05-30
[subiquity_client] import types.dart generated from subiquity/types.py

jpnurmi:
Close: #412


packages/subiquity_client/lib/src/types.dart
```diff
@@ -5,6 +5,36 @@ part 'types.g.dart';
 
 // ignore_for_file: invalid_annotation_target
 
+// subiquity serializes `Dict[int, Any]` as `List[List[int, Any]]`
+class JsonMapConverter
```
jpnurmi, line 9: I chose a more generator-friendly and property-agnostic syntax for the `Map<int, String>` JSON conversion workaround.

packages/subiquity_client/lib/src/types.dart
```diff
@@ -5,6 +5,36 @@ part 'types.g.dart';
...
+  @override
+  List<dynamic> toJson(Map<int, String> values) {
+    return [
+      for (final entry in values.entries) [entry.key, entry.value],
+    ];
+  }
+}
+
+// subiquitycore.models.network
+class NetDevInfo {
```
jpnurmi, line 29: This is an unfortunate external dependency for `types.py`. This stub ensures that the generated code below won't choke on a missing type.

packages/subiquity_client/lib/src/types.dart
```diff
@@ -5,6 +5,36 @@ part 'types.g.dart';
...
+  const NetDevInfo._();
+  factory NetDevInfo.fromJson(Map<String, dynamic> json) => NetDevInfo._();
+  Map<String, dynamic> toJson() => {};
+}
+
+// Union[Partition, Gap]
+typedef DiskObject = PartitionOrGap;
```
jpnurmi, line 36: the generator doesn't know that `Union[Partition, Gap] == DiskObject` so this is for backward compatibility

packages/subiquity_client/lib/src/types.dart
```diff
@@ -71,23 +101,63 @@ class ApplicationStatus with _$ApplicationStatus {
...
+@freezed
+class KeyFingerprint with _$KeyFingerprint {
+  const factory KeyFingerprint({
+    required String keytype,
+    required String fingerprint,
+  }) = _KeyFingerprint;
+
+  factory KeyFingerprint.fromJson(Map<String, dynamic> json) =>
+      _$KeyFingerprintFromJson(json);
 }
```
jpnurmi, line 119: there are a few new types that aren't in use in the GUI today that codecov complains about. the total coverage goes only down by 0.7% so i think it's not a big deal :)

CarlosNihelton (approved):
This looks good to me. But it's still marked as draft. Do you plan on doing anything else?

jpnurmi:
Sorry, my bad. :) I was feeling unsure if I dare to enable the generator check in the CI. It's ok now that it was left out.

PR #890, opened 2022-05-30
Shared service locator and widgets

jpnurmi:
The same service locator had already been copied to several apps so a small standalone package was made out of it.


CarlosNihelton (approved):
This looks great!
Curiously I was exploring the repositories of the ubuntu-flutter-community org when this PR appeared. :D

PR #896, opened 2022-05-31
Update Subiquity to latest

CarlosNihelton:
There are some interesting fixes for WSL in Subiquity that we'd like to get into the snap and rootfs soon.


jpnurmi (approved):
The changes look pretty subtle. I have not done a live installation. LGTM :+1:

PR #897, opened 2022-06-01
Let WSL have its own locale setting

jpnurmi:
This allows us to move the installer-specific Settings class from
ubuntu_wizard to ubuntu_desktop_installer.


CarlosNihelton (approved):
That was blazing fast! Thank you very much!

PR #900, opened 2022-06-02
Replaces `String socketPath`  by an Endpoint

CarlosNihelton:
Hello! This is the first attempt to flexibilize the relationship between the Flutter GUI clients and Subiquity server. Replacing the socket paths passed around in the API surface by a higher level construct enables us soon to compose other connections, like TCP, without changing the API.

### DIFF
--- a/packages/subiquity_client/lib/src/client.dart
+++ b/packages/subiquity_client/lib/src/client.dart
@@ -2,8 +2,8 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 import 'package:dartx/dartx.dart';
-import 'package:subiquity_client/src/server.dart';
 import 'package:ubuntu_logger/ubuntu_logger.dart';
+import 'endpoint.dart';
 import 'types.dart';
 
 /// @internal
@@ -47,7 +47,7 @@ class SubiquityClient {
   Future<bool> get isOpen => _isOpen.future;
 
   void open(Endpoint endpoint) {
-    log.info('Opening socket address ${endpoint.address}');
+    log.info('Opening socket to $endpoint');
     _endpoint = endpoint;
     _client.connectionFactory = (uri, proxyHost, proxyPort) async {
       return Socket.startConnect(endpoint.address, endpoint.port);
@@ -56,7 +56,7 @@ class SubiquityClient {
   }
 
   Future<void> close() async {
-    log.info('Closing socket address ${_endpoint.address}');
+    log.info('Closing socket to $_endpoint');
     _client.close();
   }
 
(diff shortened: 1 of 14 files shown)


packages/subiquity_client/lib/src/server.dart
```diff
@@ -11,6 +11,25 @@ enum ServerMode { LIVE, DRY_RUN }
 /// @internal
 final log = Logger('subiquity_server');
 
+/// Vocabulary type that combines an [InternetAddress] and a [port] number
+class Endpoint {
```
jpnurmi, line 15: could we have the `Endpoint` class in a separate `endpoint.dart` file so the client and server don't need to import each other?

packages/subiquity_client/lib/src/server.dart
```diff
@@ -11,6 +11,25 @@ enum ServerMode { LIVE, DRY_RUN }
...
+
+  /// Creates an Unix domain socket endpoint.
+  factory Endpoint.unix(String socketPath) => Endpoint(
+        InternetAddress(socketPath, type: InternetAddressType.unix),
+        port: 0,
+      );
+
+  /// Creates a localhost TCP endpoint on port number [port].
+  factory Endpoint.localhost(int port) => Endpoint(
+        InternetAddress("127.0.0.1", type: InternetAddressType.IPv4),
```
jpnurmi, line 27: i thought we had prefer_single_quotes enabled :D

packages/subiquity_client/lib/src/client.dart
```diff
@@ -41,23 +41,22 @@ class SubiquityException implements Exception {
...
+  late Endpoint _endpoint;
   final _isOpen = Completer<bool>();
 
   Future<bool> get isOpen => _isOpen.future;
 
-  void open(String socketPath) {
-    log.info('Opening socket ${p.absolute(socketPath)}');
-    _socketPath = socketPath;
+  void open(Endpoint endpoint) {
+    log.info('Opening socket address ${endpoint.address}');
```
jpnurmi, line 50: we could even implement Endpoint.toString()? :)

jpnurmi (commented):
Some nitpick but overall looks good! :+1:

packages/subiquity_client/lib/src/server.dart
CarlosNihelton: Silly me not noticing the server was imported into the client automatically. Sorry about that. Fixed now.

packages/subiquity_client/lib/src/server.dart
CarlosNihelton: That deserves a PR on its own :D

image

jpnurmi (approved):
LGTM :+1:

PR #917, opened 2022-06-08
Ensures port is used on Uri.http

CarlosNihelton:
In my experimentations setting the destination port on the connection factory (in Socket.startConnect) is not enough  when [openUrl](https://api.flutter.dev/flutter/dart-io/HttpClient/openUrl.html) is used, because it expects the full authority (i.e. "host:port") specified in the Uri paramter.
This creates the authority getter on the endpoint to facilitate formatting uri's.
Also subiquity client, server and monitor are modified to take this new method in consideration when building HTTP uri's.

It took me a while to convince myself about why this is necessary in the first place.


jpnurmi (approved):
LGTM :+1:

PR #922, opened 2022-06-13
Extracts out of the SubiquityServer class path-related methods

CarlosNihelton:
This is the first step towards simplifying the SubiquityServer class.

Those paths are only relevant if we start Subiquity by directly executing Python on the same host. So far that's all we do. But that is about to change when we start Subiquity inside WSL from the OOBE running on the host (for example). A similar concept would be launching Subiquity inside a LXD container. That will make more sense when we extract the launching logic as well.

By the magic of the operator `??=` we are able to cache the results of finding the subiquity_client package directory and subiquity socket paths, so we only assign the static variables once (race conditions might be created with two subsequent invocations of _findSubiquityPath() happen before a future could complete, but even so that won't be a problem because it will assign the same value twice and never again).

The definitions below were also moved to a separate file, so the new files related to the server can easily import it. We could make them `part of` but I think the current approach makes the `subiquity_server.dart` file simpler than the alternative.

```dart
enum ServerMode { LIVE, DRY_RUN }

/// @internal
final log = Logger('subiquity_server');
```

### DIFF
--- a/packages/subiquity_client/lib/src/server.dart
+++ b/packages/subiquity_client/lib/src/server.dart
@@ -1,17 +1,12 @@
 import 'dart:io';
 
 import 'package:meta/meta.dart';
-import 'package:package_config/package_config.dart';
 import 'package:path/path.dart' as p;
-import 'package:ubuntu_logger/ubuntu_logger.dart';
 import 'package:xdg_directories/xdg_directories.dart' as xdg;
 
 import 'endpoint.dart';
-
-enum ServerMode { LIVE, DRY_RUN }
-
-/// @internal
-final log = Logger('subiquity_server');
+import 'server/common.dart';
+import 'server/paths.dart';
 
 abstract class SubiquityServer {
   Process? _serverProcess;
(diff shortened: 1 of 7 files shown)


jpnurmi (commented):
Looks good in principle, but I'm wondering if we should drop the unnecessary class to follow the Effective Dart guidelines to [avoid defining a class that contains only static members](https://dart.dev/guides/language/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members).

There are cases where it definitely makes sense to use such classes, but in this case, it would make the naming less cumbersome and repetitive: `SubiquityPaths.getSubiquityPath()` vs. `getSubiquityPath()`. Dart's top-level variables are implicitly static and lazy so there's no difference in terms of initialization of the `_subiquityPath` variable whether it's top-level.

CarlosNihelton:
> Looks good in principle, but I'm wondering if we should drop the unnecessary class to follow the Effective Dart guidelines to [avoid defining a class that contains only static members](https://dart.dev/guides/language/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members).
> 
> There are cases where it definitely makes sense to use such classes, but in this case, it would make the naming less cumbersome and repetitive: `SubiquityPaths.getSubiquityPath()` vs. `getSubiquityPath()`. Dart's top-level variables are implicitly static and lazy so there's no difference in terms of initialization of the `_subiquityPath` variable whether it's top-level.

That leads to making file global variables to cache the values that are currently held inside the class. Do you see a problem with that? Do you see another option for the globals?

jpnurmi:
Holding them inside or outside the class makes no technical difference. Both are static, initialized the same way, and the level of testability is the same too. The other option just doesn't force the caller to repeat "subiquity path" twice. :)

For reference, here's a couple of Flutter team's own packages with similar functionality that use top-level functions too:
- https://pub.dev/packages/path_provider
- https://pub.dev/packages/xdg_directories

PR #927, opened 2022-06-14
Update Subiquity & add new properties

jpnurmi:


CarlosNihelton:
I'm eager to get a WSL rootfs built after this is merged. We will get the TCP support in Subiquity to play with.

jpnurmi:
Cool!

PR #928, opened 2022-06-14
Use FormLayout from ubuntu_flutter_plugins

jpnurmi:


jpnurmi:
ubuntu_wizard is on a diet :)

PR #932, opened 2022-06-15
UDI: skip disabled routes instead of commenting them out

jpnurmi:
Some pages that are not fully functional were taken out (#373) but it's
still sometimes necessary to be able to test them.

With this change, it becomes more convenient because even though the
route is always skipped in production, it becomes possible to choose it
for testing purposes. Previously, the route did not exist at all so one
would need to first edit code and uncomment the route to be able to test
the respective page. Now, one can set e.g. Routes.tryOrInstall or
Routes.chooseYourLook as Routes.initialRoute (or pass --initial-route).


CarlosNihelton (approved):
LGTM

Issue #783, opened 2022-04-06
Respect new gnome dark mode settings in app-theme and eventually new yaru color versions


Feichtmeier:
With 22.04, GCC triggers a new value 

org.gnome.desktop.interface color-scheme

![grafik](https://user-images.githubusercontent.com/15329494/161951370-27cfb427-aed0-46c9-8783-d44122e8a583.png)

It could be added to the settings class beneath
https://github.com/canonical/ubuntu-desktop-installer/blob/ba5a9505ff7d936cb236ceafc35efd8446cda5bb/packages/ubuntu_wizard/lib/settings.dart#L31

Also, if the installer wants to support the accent colors, things get a little bit more complicated. There are probably other ways to do it, but in the settings project, I've used changenotifiers for the gtk-themes instead:
https://github.com/Feichtmeier/settings/blob/master/lib/view/pages/appearance/theme_section.dart

I need to set up a new VM for the installer, then you can hopefully assign this issue to me if you want to :)

jpnurmi:
Perhaps there could be an [inherited widget](https://api.flutter.dev/flutter/widgets/InheritedTheme-class.html) in `yaru_widgets` that provides the correct theme to its children?

jpnurmi:
The 99% case
```dart
// follow the system defaults
YaruTheme(child: MyApp())
```

For the rest (e.g. an app wants a specific accent, always dark, or such)
```dart
// override specific attributes
YaruTheme(
  data: YaruThemeData(
    foo: bar, // 
    baz: qux, // 
  ),
  child: MyApp(),
)
```

jpnurmi:
@seb128 @elioqoshi Do we want to offer accent color selection on the "Choose your look" page?

![Screenshot from 2022-04-06 14-16-17](https://user-images.githubusercontent.com/140617/161972627-e55a4c9c-cf95-44d6-91da-2c97077121ba.png)

elioqoshi:
Oh no, that would be way too granular, I'd recommend to skip that part in the installer.

seb128:
I've no strong opinion and would follow Elio's suggestion

jpnurmi:
@Feichtmeier I have reported https://github.com/flutter/flutter/issues/101438 and proposed a fix https://github.com/flutter/engine/pull/32456

Feichtmeier:
@jpnurmi nice!

are you sure this will be enough to get the dark theme no matter the yaru accent color?
[...]

PR #933, opened 2022-06-15
UDI: use YaruTheme & color-scheme in GNOME 42

jpnurmi:
https://user-images.githubusercontent.com/140617/173798209-d02b830e-9faa-45d3-bb37-fd25b8dbdea5.mp4

Close: #783

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   run:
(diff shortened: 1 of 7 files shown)


jpnurmi:
The GNOME 42 dependency is ok for now - we don't need to support older than 22.04. Flavors are going to need a way to customize/replace/remove the whole page in the future, but that's another task. :slightly_smiling_face:

Feichtmeier (approved):

PR #934, opened 2022-06-15
Extracts logic to start the server out of the SubiquityServer class

CarlosNihelton:
This enables further composition combining any sort of command line programs and arguments that we can think of to use to start the server. My idea is to expose context-aware factory constructors as we find need to start subiquity in such contexts. For instance, `SubiquityLauncher.python()` combines the parameters to enable the GUI to launch subiquity on the same host it's running by invoking `python3 -m [pythonModule]`. A `SubiquityLauncher.lxd()` would have different parameters to compose a command line that would allow start subiquity inside an LXD container while the GUI would be running outside of the container, on the host environment. (Just an idea, not implemented yet). 

There are some parameters apparently not used in the general constructors which I plan to use in my next PR which will implement a `SubiquityLauncher.wsl()` and make use of it. I'll remove the `ignore lint` clauses then.

Most of this PR is just moving code out of the SubiquityServer class. Conceptually it is not that big. I promise the next ones will be smaller. :D

### DIFF
--- a/packages/subiquity_client/lib/src/server.dart
+++ b/packages/subiquity_client/lib/src/server.dart
@@ -1,23 +1,17 @@
 import 'dart:io';
 
 import 'package:meta/meta.dart';
-import 'package:path/path.dart' as p;
 
 import 'endpoint.dart';
 import 'server/common.dart';
+import 'server/process.dart';
 import 'server/paths.dart';
-import 'server/pidfile.dart';
 
-abstract class SubiquityServer {
-  Process? _serverProcess;
+class SubiquityServer {
+  /// An optional server launcher, should we need to start the server.
+  SubiquityProcess? launcher;
 
-  SubiquityServer._();
-
-  /// Creates a new subiquity server.
-  factory SubiquityServer() => _SubiquityServerImpl();
-
-  /// Creates a WSL variant of the server.
-  factory SubiquityServer.wsl() => _WslSubiquityServerImpl();
+  SubiquityServer({this.launcher});
 
   /// A callback for integration testing purposes. The callback is called when
   /// the server has been started and thus, the application is ready for
@@ -25,86 +19,18 @@ abstract class SubiquityServer {
   @visibleForTesting
   static void Function(Endpoint)? startupCallback;
 
-  // Whether the server process should be started in the specified mode.
-  bool _shouldStart(ServerMode mode);
-
-  // The name of the server's Python module.
-  String get _pythonModule;
-
-  // Prefer local curtin and probert python modules that are pinned to the
-  // correct versions
-  Map<String, String> _pythonPath(String subiquityPath) {
-    final pythonPath = (Platform.environment['PYTHONPATH'] ?? '').split(':');
-    pythonPath.add(subiquityPath);
-    pythonPath.add(p.join(subiquityPath, 'curtin'));
-    pythonPath.add(p.join(subiquityPath, 'probert'));
-    return {'PYTHONPATH': pythonPath.join(':')};
-  }
-
   Future<Endpoint> start(ServerMode serverMode,
       {List<String>? args, Map<String, String>? environment}) async {
     final socketPath = await getSocketPath(serverMode);
     final endpoint = Endpoint.unix(socketPath);
-    if (_shouldStart(serverMode)) {
-      var subiquityCmd = <String>[
-        '-m',
-        _pythonModule,
-        if (serverMode == ServerMode.DRY_RUN) '--dry-run',
-        ...?args,
-      ];
-      await _startSubiquity(serverMode, subiquityCmd, environment);
+    if (launcher != null) {
+      await launcher!.start(additionalArgs: args, additionalEnv: environment);
     }
 
     return _waitSubiquity(endpoint).then((_) {
       startupCallback?.call(endpoint);
       return endpoint;
     });
-  }
-
-  Future<void> _startSubiquity(ServerMode serverMode, List<String> subiquityCmd,
-      Map<String, String>? environment) async {
-    final subiquityPath = await getSubiquityPath();
-    String? workingDirectory;
-    // try using local subiquity
-    if (Directory(subiquityPath).existsSync()) {
-      workingDirectory = subiquityPath;
-    }
-
-    // kill the existing test server if it's already running, so they don't pile
-    // up on hot restarts
-    final pid = await readPidFile();
-    if (pid != null) {
-      Process.killPid(pid);
-    }
-
-    // Use `/usr/bin/python3` over `/snap/flutter/current/usr/bin/python3` when
-    // developing with flutter-snap on the desktop. This ensures that subiquity
-    // has locally installed Python module dependencies available. (#364)
-    _serverProcess = await Process.start(
-      Platform.environment['SNAP_PYTHON'] ?? '/usr/bin/python3',
-      subiquityCmd,
-      workingDirectory: workingDirectory,
-      environment: {
-        ..._pythonPath(subiquityPath),
-        if (serverMode == ServerMode.DRY_RUN) ...{
-          // so subiquity doesn't think it's some other snap (e.g. flutter or vs code)
-          'SNAP': '.',
-          'SNAP_NAME': 'subiquity',
-          'SNAP_REVISION': '',
-          'SNAP_VERSION': '',
-        },
-        ...?environment,
-      },
-    ).then((process) {
-      stdout.addStream(process.stdout);
-      stderr.addStream(process.stderr);
-      return process;
-    });
-    log.info(
-      'Starting server (PID: ${_serverProcess!.pid}) with args: $subiquityCmd',
-    );
-
-    await writePidFile(_serverProcess!.pid);
   }
 
   static Future<void> _waitSubiquity(Endpoint endpoint) async {
@@ -129,35 +55,5 @@ abstract class SubiquityServer {
     client.close();
   }
 
-  Future<void> stop() async {
-    try {
-      await pidFile().delete();
-    } on FileSystemException catch (_) {}
-    _serverProcess?.kill();
-    await _serverProcess?.exitCode;
-  }
-}
-
-class _SubiquityServerImpl extends SubiquityServer {
-  _SubiquityServerImpl() : super._();
-
-  // Normally, the server is already running in live mode and thus, only
-  // started in dry-run mode.
-  @override
-  bool _shouldStart(ServerMode mode) => mode == ServerMode.DRY_RUN;
-
-  @override
-  String get _pythonModule => 'subiquity.cmd.server';
-}
-
-// A server that runs in a WSL environment.
-class _WslSubiquityServerImpl extends SubiquityServer {
-  _WslSubiquityServerImpl() : super._();
-
-  // The server should be always started in WSL because there's no systemd.
-  @override
-  bool _shouldStart(ServerMode mode) => true;
-
-  @override
-  String get _pythonModule => 'system_setup.cmd.server';
+  Future<void> stop() async => await launcher?.stop();
 }
--- a/packages/subiquity_client/lib/src/server.dart
+++ b/packages/subiquity_client/lib/src/server.dart
@@ -7,11 +7,17 @@ import 'server/common.dart';
 import 'server/process.dart';
 import 'server/paths.dart';
 
+Future<Endpoint> defaultEndpoint(ServerMode serverMode) async {
+  final socketPath = await getSocketPath(serverMode);
+  return Endpoint.unix(socketPath);
+}
+
 class SubiquityServer {
   /// An optional server launcher, should we need to start the server.
   SubiquityProcess? launcher;
+  final Endpoint endpoint;
 
-  SubiquityServer({this.launcher});
+  SubiquityServer({this.launcher, required this.endpoint});
 
   /// A callback for integration testing purposes. The callback is called when
   /// the server has been started and thus, the application is ready for
@@ -19,10 +25,10 @@ class SubiquityServer {
   @visibleForTesting
   static void Function(Endpoint)? startupCallback;
 
-  Future<Endpoint> start(ServerMode serverMode,
-      {List<String>? args, Map<String, String>? environment}) async {
-    final socketPath = await getSocketPath(serverMode);
-    final endpoint = Endpoint.unix(socketPath);
+  Future<Endpoint> start({
+    List<String>? args,
+    Map<String, String>? environment,
+  }) async {
     if (launcher != null) {
       await launcher!.start(additionalArgs: args, additionalEnv: environment);
     }
--- a/packages/subiquity_client/lib/src/server.dart
+++ b/packages/subiquity_client/lib/src/server.dart
@@ -14,10 +14,10 @@ Future<Endpoint> defaultEndpoint(ServerMode serverMode) async {
 
 class SubiquityServer {
   /// An optional server launcher, should we need to start the server.
-  SubiquityProcess? launcher;
+  SubiquityProcess? process;
   final Endpoint endpoint;
 
-  SubiquityServer({this.launcher, required this.endpoint});
+  SubiquityServer({this.process, required this.endpoint});
 
   /// A callback for integration testing purposes. The callback is called when
   /// the server has been started and thus, the application is ready for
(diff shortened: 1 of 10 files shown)


packages/subiquity_client/lib/src/server.dart
```diff
@@ -1,58 +1,49 @@
...
-  /// Creates a new subiquity server.
-  factory SubiquityServer() => _SubiquityServerImpl();
-
-  /// Creates a WSL variant of the server.
-  factory SubiquityServer.wsl() => _WslSubiquityServerImpl();
+class SubiquityServer {
+  /// An optional server launcher, should we need to start the server.
+  SubiquityLauncher? launcher;
+
+  SubiquityServer._({this.launcher});
```
jpnurmi, line 14: can we make this public?

packages/subiquity_client/lib/src/server.dart
```diff
@@ -1,58 +1,49 @@
...
+
+  /// A variant of the server for Ubuntu Desktop Installer. By default starts
+  /// subiquity in dry-run. Set [isLiveRun] for a live run.
+  factory SubiquityServer({bool isLiveRun = false}) => SubiquityServer._(
+        launcher: isLiveRun
+            ? null
+            : SubiquityLauncher.python(
```
jpnurmi, line 21: can we construct the launcher in `udi/installer.dart`?

packages/subiquity_client/lib/src/server.dart
```diff
@@ -1,58 +1,49 @@
...
+                'subiquity.cmd.server',
+                serverMode: ServerMode.DRY_RUN,
+              ),
+      );
+
+  /// A server that runs in a WSL environment.
+  factory SubiquityServer.wsl(ServerMode mode) => SubiquityServer._(
```
jpnurmi, line 28: ...and this in `wsl/main.dart`

packages/subiquity_client/lib/src/server/launcher.dart
```diff
@@ -0,0 +1,233 @@
...
+  } on FileSystemException catch (e) {
+    log.warning('Error writing ${file.path} (${e.message}). '
+        'Hot restarts may cause multiple Subiquity test servers to run.');
+  }
+}
+
+File _pidFile() {
+  final dir = xdg.runtimeDir?.path ?? Directory.current.path;
+  return File('$dir/subiquity-test-server.pid');
+}
```
jpnurmi, line 31: we could drop these to a separate `pidfile.dart` while moving them :)

packages/subiquity_client/lib/src/server/launcher.dart
```diff
@@ -0,0 +1,233 @@
...
+  /// list generated by calling [argsFactory].
+  final Future<List<String>> Function()? argsFactory;
+
+  /// Optional function that will be called to regenerate the server process's
+  /// working directory.
+  final Future<String?> Function()? workingDirectoryFactory;
+
+  /// Optional function that will be called to regenerate the server environment
+  /// variables set.
+  final Future<Map<String, String>> Function()? environmentFactory;
```
jpnurmi, line 108: i'm hoping we could make these ordinary properties if we call getSubiquityPath() first and pass it to the launcher

CarlosNihelton:
Forgot to mention that with this approach the SubiquityServer class becomes concrete, no need for split implementations.

jpnurmi (commented):
great work so far! :+1:

i'm wondering if we could change the flow a bit so that we call and await getSubiquityPath() first, then create the launcher and pass the path to it, and finally, pass the launcher to subiquity server that puts it all together? this should eliminate the need to use so asynchronous factories for the launcher properties.

CarlosNihelton:
> great work so far! +1
> 
> i'm wondering if we could change the flow a bit so that we call and await getSubiquityPath() first, then create the launcher and pass the path to it, and finally, pass the launcher to subiquity server that puts it all together? this should eliminate the need to use so asynchronous factories for the launcher properties.

I think the ideas you propose will make this nicer! I'm just wondering about having `main()` async or instead of passing a server to the runWizardApp we should pass a Future server or an async server factory instead. What do you think?

packages/subiquity_client/lib/src/server/launcher.dart
CarlosNihelton: I should have considered that before. It could be a small and self-contained PR. Anyways..

jpnurmi:
Async `main()` is fine. I'm surprised it's not already. :)

packages/subiquity_client/lib/src/server.dart
CarlosNihelton: Done

packages/subiquity_client/lib/src/server.dart
CarlosNihelton: Done. Async main helped a lot.

packages/subiquity_client/lib/src/server.dart
CarlosNihelton: Done.

packages/subiquity_client/lib/src/server/launcher.dart
CarlosNihelton: Factories are not necessary thanks to the async main.

packages/subiquity_client/lib/src/server/launcher.dart
```diff
@@ -0,0 +1,233 @@
...
+        '/usr/bin/python3';
+
+    return SubiquityLauncher(
+      cmd,
+      args: ['-m', pythonModule],
+      // will try using local subiquity
+      workingDirectoryFactory: () => getSubiquityPath().then((subiquityPath) =>
+          Directory(subiquityPath).existsSync() ? subiquityPath : null),
+      environmentFactory: () => getSubiquityPath().then(
+        (subiquityPath) => {
```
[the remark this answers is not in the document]
CarlosNihelton: Done.

packages/subiquity_client/lib/src/server/launcher.dart
```diff
@@ -0,0 +1,233 @@
...
+///     if (configFile != null) ...['-f', configFile],
+///     '--',
+///     command,
+///     ...?args
+///  ];
+///  return SubiquityLauncher('lxc-execute', lxcArgs);
+///}
+///```
+///
+class SubiquityLauncher {
```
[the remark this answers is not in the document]
CarlosNihelton: Indeed. Although it's focus is starting, there is the pid file handling and stopping, so I renamed it per your suggestion.

packages/subiquity_client/lib/src/server/launcher.dart
```diff
@@ -0,0 +1,233 @@
...
+            'SNAP_VERSION': '',
+          },
+        },
+      ),
+    );
+  }
+
+  /// Starts the server. Returns it's [Process].
+  Future<Process?> start(
+    ServerMode mode,
```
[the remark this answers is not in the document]
CarlosNihelton: Removed the server mode from this function call. Also it became unnecessary on the `runWizardApp()` function.

CarlosNihelton:
I'm wondering now if I can say SubiquityServer is also on a diet :D

packages/ubuntu_wsl_setup/lib/main.dart
```diff
@@ -22,6 +22,15 @@ screens, yet allowing user to overwrite any of those during setup.
...
   final variant = ValueNotifier<Variant?>(null);
+  final liveRun = isLiveRun(options);
+  final serverMode = liveRun ? ServerMode.LIVE : ServerMode.DRY_RUN;
+  final subiquityPath = await getSubiquityPath();
+  final endpoint = await defaultEndpoint(serverMode);
+  final launcher = SubiquityProcess.python(
+    'system_setup.cmd.server',
+    serverMode: serverMode,
+    subiquityPath: subiquityPath,
+  );
```
jpnurmi, line 33: this is pure gold :1st_place_medal:

packages/ubuntu_wsl_setup/lib/main.dart
```diff
@@ -22,6 +22,15 @@ screens, yet allowing user to overwrite any of those during setup.
     );
   })!;
   final variant = ValueNotifier<Variant?>(null);
+  final liveRun = isLiveRun(options);
+  final serverMode = liveRun ? ServerMode.LIVE : ServerMode.DRY_RUN;
+  final subiquityPath = await getSubiquityPath();
+  final endpoint = await defaultEndpoint(serverMode);
+  final launcher = SubiquityProcess.python(
```
jpnurmi, line 29: s/launcher/process/

packages/ubuntu_desktop_installer/lib/main.dart
```diff
@@ -1,3 +1,3 @@
 import 'installer.dart';
 
-void main(List<String> args) => runInstallerApp(args);
+void main(List<String> args) async => runInstallerApp(args);
```
jpnurmi, line 3:
```suggestion
Future<void> main(List<String> args) => runInstallerApp(args);
```

packages/ubuntu_wsl_setup/lib/main.dart
```diff
@@ -8,7 +8,7 @@ import 'package:ubuntu_wizard/utils.dart';
 import 'app.dart';
 import 'installing_state.dart';
 
-void main(List<String> args) {
+void main(List<String> args) async {
```
jpnurmi, line 11:
```suggestion
Future<void> main(List<String> args) async {
```

packages/subiquity_client/lib/src/server.dart
```diff
@@ -1,58 +1,36 @@
...
+Future<Endpoint> defaultEndpoint(ServerMode serverMode) async {
+  final socketPath = await getSocketPath(serverMode);
+  return Endpoint.unix(socketPath);
+}
 
-  /// Creates a new subiquity server.
-  factory SubiquityServer() => _SubiquityServerImpl();
+class SubiquityServer {
+  /// An optional server launcher, should we need to start the server.
+  SubiquityProcess? launcher;
```
jpnurmi, line 17: s/launcher/process/

packages/subiquity_client/lib/src/server.dart
```diff
@@ -129,35 +61,5 @@ abstract class SubiquityServer {
...
-class _WslSubiquityServerImpl extends SubiquityServer {
-  _WslSubiquityServerImpl() : super._();
-
-  // The server should be always started in WSL because there's no systemd.
-  @override
-  bool _shouldStart(ServerMode mode) => true;
-
-  @override
-  String get _pythonModule => 'system_setup.cmd.server';
+  Future<void> stop() async => await launcher?.stop();
```
jpnurmi, line 64:
```suggestion
  Future<void> stop() async => launcher?.stop();
```
the await is not necessary inside stop, we can return the future and let the caller decide (one could e.g. await multiple futures in parallel with Future.wait([...]))

packages/subiquity_client/lib/src/server.dart
```diff
@@ -1,58 +1,36 @@
...
-        if (serverMode == ServerMode.DRY_RUN) '--dry-run',
-        ...?args,
-      ];
-      await _startSubiquity(serverMode, subiquityCmd, environment);
+  Future<Endpoint> start({
+    List<String>? args,
+    Map<String, String>? environment,
+  }) async {
+    if (launcher != null) {
+      await launcher!.start(additionalArgs: args, additionalEnv: environment);
```
jpnurmi, line 33: s/launcher/process/

packages/subiquity_client/lib/src/server/process.dart
```diff
@@ -0,0 +1,183 @@
...
+/// The example below would allow starting subiquity inside a WSL2 distro
+/// while the GUI client is running on the Windows host.
+///
+/// ```dart
+/// /// Able to start the server inside [distroName] on WSL from the Windows host.
+/// factory SubiquityProcess.wsl(String distroName, String command,
+///     {List<String>? args,
+///     Future<void>? defer,
+///     Future<void> Function()? onProcessStart,
+///     Future<List<String>> Function()? argumentsFactory}) {
```
jpnurmi, line 33: was this meant to be a map of environment vars?

packages/subiquity_client/lib/src/server/process.dart
```diff
@@ -0,0 +1,183 @@
...
+  return {'PYTHONPATH': pythonPath.join(':')};
+}
+
+/// Manages the Subiquity process.
+/// This class knows the many ways with which we may want to start Subiquity.
+/// Starting the server builds on top of [Process.start], so this launcher
+/// can be customized by implementing factory constructors that adjust the
+/// [command] and [args] (and possibly some factories to enable enhancing the
+/// command line arguments, working directory and environment with the results
+/// of asynchronous computations).
```
jpnurmi, line 23: the content within parentheses is outdated and can be removed?

packages/subiquity_client/test/subiquity_client_test.dart
```diff
@@ -19,11 +19,57 @@ void main() {
...
+    test('set additional environment before starting the process', () async {
+      const foo = '42';
+      final env = {'TEST_VAR': foo};
+      final process = SubiquityProcess(
+        'bash',
+        ['-c', 'exit \$TEST_VAR'],
+      );
+      await process.start(additionalEnv: env);
+      expect(await process.exitCode, int.parse(foo));
+      await process.stop();
```
jpnurmi, line 32:
```suggestion
      await process.start(additionalEnv: env);
      addTearDown(process.stop);
      expect(await process.exitCode, int.parse(foo));
```
ensures clean up even if expect() may fail

jpnurmi (commented):
I love where this is going! Just a couple of nits left and we're ready to go

packages/ubuntu_wsl_setup/lib/main.dart
CarlosNihelton: I went to the opposite direction due the [effective Dart recommendation for async top-level main functions, where the Future annotation can (and generally should) be dropped in favor of void[(https://dart-lang.github.io/linter/lints/avoid_void_async.html).

Is there something in the runtime expecting to receive this Future<void> returned by main?

CarlosNihelton: This can be a little bit confusing:  (Here we are recommended to make even top level main Future<void>)[https://dart.dev/codelabs/async-await#working-with-futures-async-and-await:~:text=If%20the%20function,%7B%20%C2%B7%C2%B7%C2%B7%20%7D]

Would you have a decisive argument for that?

packages/subiquity_client/lib/src/server.dart
CarlosNihelton: Ditto

packages/subiquity_client/lib/src/server.dart
CarlosNihelton: Good point. Fixed.

packages/ubuntu_wsl_setup/lib/main.dart
jpnurmi: A hidden agenda. :) Later on, we could return a future that lets integration tests know when the app is fully up and running and that way get rid of the ugly static startup callback we're relying on today.

CarlosNihelton: Good point. I'll fix that.

packages/subiquity_client/lib/src/server/process.dart
CarlosNihelton: Right.

packages/subiquity_client/lib/src/server/process.dart
CarlosNihelton: Leftovers as well. Should have been just removed.

jpnurmi (approved):
Looks great, thank you! :+1: The manual .stop() calls can be removed now that we have addTearDown().

packages/subiquity_client/test/subiquity_client_test.dart
```diff
@@ -19,11 +19,60 @@ void main() {
...
+      const foo = '42';
+      final env = {'TEST_VAR': foo};
+      final process = SubiquityProcess.command(
+        'bash',
+        ['-c', 'exit \$TEST_VAR'],
+      );
+      await process.start(additionalEnv: env);
+      addTearDown(process.stop);
+      expect(await process.exitCode, int.parse(foo));
+      await process.stop();
```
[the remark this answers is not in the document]
CarlosNihelton: My bad. :(

CarlosNihelton:
Thank you for the guidance on this. It became a little big bigger than I anticipated, but we are going to benefit from this work!

jpnurmi:
This is a game-changer. Congrats, fantastic job! :tada:

PR #935, opened 2022-06-15
Extracts pid file related functions out of the SubiquityServer class

CarlosNihelton:


jpnurmi (approved):
Thanks for going through the trouble of splitting it into a separate PR. LGTM :+1:

PR #937, opened 2022-06-16
Subiquity process factory for WSL

CarlosNihelton:
Able to start a Subiquity server inside a WSL Linux distribution from the outside (host) environment.
With same portable and some non-portable tests. I'll figure out later how I want to setup a CI for Windows-specfic testing. That's more likely to live outside the UDI repository.
The added behavior is not used just yet.


jpnurmi (approved):
LGTM :+1:

PR #944, opened 2022-06-20
Adds TCP Helpers

CarlosNihelton:
Useful to safely find a port available on the system.

Subiquity will bind on that port later, on the Linux side.
This socket must be held binded prior to and closed just before the client attempts to connect to minimize chance of race condition. This way the localhost loopback interface forwarding will just work. Automagically.
It's required that the port range is agreed between Linux and Windows.

Testing suggestions are welcome.

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/services/tcp_helpers.dart
+++ b/packages/ubuntu_wsl_setup/lib/services/tcp_helpers.dart
@@ -1,19 +1,22 @@
 import 'dart:async';
 import 'dart:io';
+import 'dart:math';
 
 // This is inherently a race condition on its own, because in between finding an
 // available port and using it, the system may have binded other socket.
-FutureOr<RawServerSocket?> _tryBindUnsafe({
+// Thus, the solution is to hold the socket binded.
+FutureOr<ServerSocket?> _tryBindUnsafe({
   required int lower,
   required int higher,
 }) async {
-  final socket = await RawServerSocket.bind(InternetAddress.loopbackIPv4, 0);
-  final port = socket.port;
-  if (port > lower && port < higher) {
+  var rand = Random.secure();
+  var port = lower + rand.nextInt(higher - lower);
+  try {
+    final socket = await ServerSocket.bind(InternetAddress.loopbackIPv4, port);
     return socket;
-  }
-  socket.close();
+  } on SocketException {
   return null;
+  }
 }
 
 /// Returns a RawServerSocket already binded to a port in the open interval
--- a/packages/ubuntu_wsl_setup/lib/services/tcp_helpers.dart
+++ b/packages/ubuntu_wsl_setup/lib/services/tcp_helpers.dart
@@ -20,17 +20,34 @@ FutureOr<ServerSocket?> _tryBindUnsafe({
 }
 
 class TcpHelperService {
+  /// Ephemeral port ranges (* chart is not on scale)
+  ///
+  /// ```
+  /// <Linux 2.4+>           <Windows 8+>           </Linux 2.4+> </Windows 8+>
+  ///       |----------------------|--------------------|-------------|
+  ///     32768                  49152                60999         65535
+  ///                        [kMinPortNo]           [kMaxPortNo]
+  /// ```
+  static const kMinPortNo = 49152;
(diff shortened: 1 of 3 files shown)


packages/ubuntu_wsl_setup/lib/services/tcp_helpers.dart
```diff
@@ -0,0 +1,38 @@
...
+  socket.close();
+  return null;
+}
+
+/// Returns a RawServerSocket already binded to a port in the open interval
+/// [lower]-[higher] of the loopback interface where the referred range is
+/// a subset of the platform ephemeral port range.
+/// On failure returns null. [maxAttempts] limits the amount of asynchronous
+/// iterations performed.
+Future<RawServerSocket?> getSocketHolder({
```
jpnurmi, line 24: i like the "try bind" name of the internal helper function more than "get socket holder" because it's easier to understand what it does :smile:

jpnurmi (commented):
I remember we recently discussed that having global functions is ok in Dart in general but for this, it might make sense to make it a service object. Registering a mock service would help to avoid creating actual sockets in tests that will use the service. For testing the service itself, [IOOverrides](https://api.flutter.dev/flutter/dart-io/IOOverrides-class.html) should help. :slightly_smiling_face:

packages/ubuntu_wsl_setup/lib/services/tcp_helpers.dart
```diff
@@ -0,0 +1,38 @@
...
+
+// This is inherently a race condition on its own, because in between finding an
+// available port and using it, the system may have binded other socket.
+FutureOr<RawServerSocket?> _tryBindUnsafe({
+  required int lower,
+  required int higher,
+}) async {
+  final socket = await RawServerSocket.bind(InternetAddress.loopbackIPv4, 0);
+  final port = socket.port;
+  if (port > lower && port < higher) {
```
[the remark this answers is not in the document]
CarlosNihelton: I want to have some randomness in the port selection, thus I deferred to the OS to give me a port in the ephemeral range, but since the ranges are not equal between Windows and Linux I made those values as customization points for callers. My intended usage will be using Windows lower bound and Linux upper bound (the intersection of both ranges). In fact, the stricter the range, more prone to fail.
Another approach would be using random number generation to always generate ports inside that range and them attempt to bind. Actually after some experimentation I'm observing a monotonically growing trend on the Windows side, always returning higher and higher ports. Since Windows may generate 4536 ports beyond Linux's upper limit the current algorithm could easily experience a premature (and even not real) port exaustion. Even because every port attempted stays blocked after close for a certain amount of time (OS security feature).
I'll go with the random number generation.

packages/ubuntu_wsl_setup/lib/services/tcp_helpers.dart
```diff
@@ -0,0 +1,60 @@
...
+  /// ```
+  static const kMinPortNo = 49152;
+  static const kMaxPortNo = 60999;
+
+  /// Returns a ServerSocket already binded to a randomly-selected port in the
+  /// open interval [lower]-[higher] of the loopback interface where the range
+  /// is a subset of the platform ephemeral port range, limited by [kMinPortNo]
+  /// and [kMaxPortNo]. [maxAttempts] limits the amount of asynchronous
+  /// iterations performed. On failure returns null.
+  static Future<ServerSocket?> getRandomPortSocket({
```
jpnurmi, line 39: this needs to be non-static to be mockable. for example, we could eventually have something like this:
```dart
void main() {
  ...
  registerService<TcpSocketService>(TcpSocketService.new);
  ...
}

void somewhereElse() {
  final service = getService<TcpSocketService>();
  final socket = service.getRandomPortSocket();
  ...
}
```

and in tests:
```dart
final socket = MockServerSocket();
when(socket.port).thenReturn(54321);

final service = MockTcpSocketService();
when(service.getRandomPortSocket()).thenAnswer((_) async => socket);

registerMockService(service);
...
```
this way the tests would be 100% deterministic because they won't try to bind real sockets that could fail for unrelated reasons.

jpnurmi (approved):
LGTM :+1:

PR #945, opened 2022-06-22
Refactoring wsl wizard

CarlosNihelton:
At the first glance this seems redundant, but consider that we are about to add a different startup behavior that will delay considerably the moment the app is informed about the server variant (due WSL distribution registration). That behavior will require creating a new route (the slide show), but other than that the rest of the wizard would be duplication of what we currently have. The current design would require creating at least one new wizard class to accomodate that scenario. And with a new dimmension to decide upon (instead of just the variant): which Platform are we running on.
So, since the wizard_router started supporting per route onNext customisation we can build a single wizard able to start at different routes preserving the current behavior while making it easy to add new routes and conditions to jump between those routes.
This PR starts exploring this feature, for now just joining the wizards into one. Later, on top of this, I'll place a different `initalRoute` that will explore the `routeForVariant()` method as its `onNext` to select where it goes to.

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/app.dart
+++ b/packages/ubuntu_wsl_setup/lib/app.dart
@@ -6,6 +6,7 @@ import 'package:yaru/yaru.dart';
 import 'l10n.dart';
 import 'locale.dart';
 import 'wizard.dart';
+import 'routes.dart';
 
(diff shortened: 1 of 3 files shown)


jpnurmi (approved):
LGTM :+1:

PR #952, opened 2022-06-27
Check if subiquity path exists before passing it to the process class

CarlosNihelton:
I noticed a `File not found crash` in WSL preview, which seems to be due the working directory (I'm sure that Python and the required modules exist inside the snap). I'd expect a more specific exception message, but that's the way it is. Since the subiquityPath is passed at the highest level, I thought checking whether it exists or not and what to do with that information also belongs to that level. Right now it looks code duplication. That might change in the future. It's worthy mentioning that the check existed until recently, thus a regression.

_I also took the opportunity to fix a typo nearby_ :D

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/installer.dart
+++ b/packages/ubuntu_desktop_installer/lib/installer.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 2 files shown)


CarlosNihelton:
Reviewing older logs it should be expected for WSL not to find `subiquityPath` in live run and it is very likely that a subiquity subdirectory in the current working directory will not exist, because the current working directory might be anywhere.

jpnurmi (approved):
Thanks, LGTM :+1:

PR #953, opened 2022-06-28
Use handy_window

jpnurmi:
The [handy_window](https://pub.dev/packages/handy_window) plugin makes it straightforward to use `HdyWindow` from [libhandy](https://gitlab.gnome.org/GNOME/libhandy). We can remove all libhandy references - the plugin will take care of building and setting it up.


jpnurmi:
A proof that the snap works too :)

Screenshot from 2022-06-28 13-48-53

CarlosNihelton (approved):
That's really great!

Just out of curiosity: is there a technical reason for the upstream not calling `fl_register-plugins()` before `gtk_widget_show()` as the `handy_window` plugin requires?

PR #958, opened 2022-06-30
Remove Subiquity init and startup callbacks

jpnurmi:
This approach makes sure that all involved futures are awaited, and that integration tests can simply await main() to reach the ready-for-testing state.


CarlosNihelton (approved):
The hidden agenda became true! That looks cleaner and more natural! Thanks for doing that!

PR #960, opened 2022-07-01
Cross platform server/paths

CarlosNihelton:
`package.root.path` returns Unix style paths even when running on Windows.
This could cause crashes on test code running on Windows. No real issue is anticipated in production.

Even though, let's use `toFilePath()` to ensure a cross platform path representation.


jpnurmi (approved):
LGTM :+1:

PR #961, opened 2022-07-01
Platform capabilities

CarlosNihelton:
Groups all the results of decisions that would otherwise be a clause of type `if(Plaform.isXyz) {` spread everywhere in the code base so that WSL Setup wizard can run either on Linux (inside the WSL distro) or on the Windows host.

With this we are very close to enable the Windows platform for the `ubuntu_wsl_setup` package.


jpnurmi:
A common way to implement app flavors/variants in Flutter is to implement multiple entry-points, such as `main.dart` for production and `main_dev.dart` for development. It's a great place to create the preferred service variants, determine the initial route, and set up other things that may differ between the variants. In our case, we'd be even able to offer a slightly different set of command-line arguments (#962).

jpnurmi:
Yeah, for `flutter run|build -t` and during development one can just conveniently select the entry-point in your IDE if you use one. :)  We could make it possible to run manual tests for either variant on either platform if we don't "hard-wire" it to a specific platform.

CarlosNihelton:
That's pure gold! I'll implement it!

PR #964, opened 2022-07-04
Move UDI's mark_configured from the common wizard to the installer

jpnurmi:
All these controllers are UDI-specific and marking them configured is
irrelevant to WSL. Thus, moving them from the shared ubuntu_wizard to
UDI makes most sense. At the same time, this fixes the regression
introduced by #958 that markConfigured() must be called after
setVariant().


CarlosNihelton (approved):
Thanks for doing that!

PR #968, opened 2022-07-04
InstallationState.proceedToSetup is terminal

CarlosNihelton:
Noticed during live run testing with an actual WSL image. Installation slides view model should not allow further state transitions after the `proceedToSetup` state. This prevents executing code when the slide show is not the current page.


jpnurmi (approved):
LGTM. Would be a low effort to add a test, though? :)

jpnurmi (approved):
Thanks for adding tests! :+1:

PR #969, opened 2022-07-04
Entry point file for the Windows implementation

CarlosNihelton:
As suggested in #961 an entry point for the OOBE on Windows is being rolled out in this pull request. Testing this entry point is subject to a new PR (and spec) since it requires an actual WSL instance.

Also since the installation slides page is meant to be the default route for this app, it was required to integrate it.

Some common pieces of both `main()` functions were extracted to avoid duplication.


packages/ubuntu_wsl_setup/lib/main_common.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:args/args.dart';
```
jpnurmi, line 1: nit-pick: files called "main" are expected to have a "main" entry-point. :wink: could we give this file a name without "main" in it, please? perhaps something related to cli or args?

packages/ubuntu_wsl_setup/lib/app.dart
```diff
@@ -41,7 +43,7 @@ class UbuntuWslSetupApp extends StatelessWidget {
   }
 
   Widget buildWizard(BuildContext context) {
-    if (variant == null) {
+    if (variant == null && withSlides == false) {
```
jpnurmi, line 46: do we need a separate `withSlides` or would the existing `initialRoute` already carry this information?

packages/ubuntu_wsl_setup/lib/main_win.dart
```diff
@@ -0,0 +1,97 @@
...
+import 'services/journal.dart';
+
+Future<void> main(List<String> args) async {
+  final options = parseCommandLine(args, onPopulateOptions: (parser) {
+    addCommonCliOptions(parser);
+    parser.addOption(
+      'distro-name',
+      valueHelp: 'WSL Distribution name',
+      help: 'The name under which the distribution is being registered on WSL.',
+      defaultsTo: const String.fromEnvironment(
```
jpnurmi, line 24: is it intentionally a Dart define and not using Platform.environment?

packages/ubuntu_wsl_setup/lib/main_win.dart
```diff
@@ -0,0 +1,97 @@
...
+      ),
+    );
+  })!;
+  final variant = ValueNotifier<Variant?>(null);
+  final liveRun = isLiveRun(options);
+  final serverMode = liveRun ? ServerMode.LIVE : ServerMode.DRY_RUN;
+  final tcpService = TcpSocketService();
+  final socketHolder = await tcpService.getRandomPortSocket();
+  // TODO: Handle the null case.
+  // That would mean port exaustion and the simplest way to solve it is by
```
jpnurmi, line 36: unimportant typo: exhaustion :)

jpnurmi (approved):
LGTM :+1:

packages/ubuntu_wsl_setup/lib/main_win.dart
CarlosNihelton: Yes, it's intentional. That piece of information is known at compile time. I want to have that value defined by the build system instead of the execution runtime to reduce the chance of mistakes due user defining a variable with that name and that influencing the behavior of the OOBE. It might be a theoretical risk at this point, but for now this seems a stronger approach.

PR #970, opened 2022-07-05
Await window_manager initialization

CarlosNihelton:
`window_manager` basically does nothing without that call on Windows OS, it's during that call that the plugin retrieves the top-level HWND, which is later target of all the plugin methods.

While the plugin could retrieve that information from the `view()->GetNativeWindow()` at each method call, that could be less performant, I guess.

Anyway, it's easier to fix it here than on the plugin and upgrading it.

### DIFF
--- a/packages/ubuntu_wizard/lib/src/utils/window.dart
+++ b/packages/ubuntu_wizard/lib/src/utils/window.dart
@@ -1,6 +1,6 @@
 import 'dart:async';
 
-import 'package:flutter/cupertino.dart';
+import 'package:flutter/widgets.dart';
 import 'package:ubuntu_logger/ubuntu_logger.dart';
 import 'package:window_manager/window_manager.dart';
 
--- a/packages/ubuntu_wizard/lib/src/utils/window.dart
+++ b/packages/ubuntu_wizard/lib/src/utils/window.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 
+import 'package:flutter/cupertino.dart';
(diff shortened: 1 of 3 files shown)


packages/ubuntu_wizard/lib/src/utils/window.dart
```diff
@@ -43,3 +43,6 @@ Future<void> setWindowTitle(String title) {
 Future<void> setWindowClosable(bool closable) {
   return windowManager.setClosable(closable);
 }
+
+/// Ensures initialization of the plugin(s) related to managing windows.
+Future<void> ensureInitialized() => windowManager.ensureInitialized();
```
jpnurmi, line 48: can we have both ensure initialized calls here?
```suggestion
Future<void> ensureInitialized() {
  WidgetsFlutterBinding.ensureInitialized();
  return windowManager.ensureInitialized();
}
```

CarlosNihelton: I swear I considered that idea and even calling the function something with "framework", but I got unsure whether I could be mixing responsibilities. I agree it looks nicer.

jpnurmi (approved):
Thanks, LGTM :+1:

jpnurmi:
what if we would bury the await inside onWindowClosed()?

CarlosNihelton:
> 

Not sure if it's necessary. I'll get this one merged as is, since it's fixed and we reconsider the idea later, when our minds are restored.

PR #971, opened 2022-07-06
`flutter create --platform=windows .`

CarlosNihelton:
With the exception of the default `README`, `analysis_options.yaml` and `widgets_test.dart` that were excluded and the `app_icon.ico` that was replaced by the same icon we use for the Ubuntu Preview app, pretty much everything else is auto generated by the flutter tool after running the command above on the `ubuntu_wsl_setup` package root directory. 

The icon will be replaceable by the app build system according to the version of the WSL app being built anyway, so the current icon is just a default.

I was unsure about the `.metadata` file since it mentions iOS related files in its contents, but even though it seems to be required, so I pushed the updates herein as well.

With this and an instance of Ubuntu WSL it's possible to run the OOBE on Windows. The last missing piece feature-wise, besides integration tests, is the communication that will let the OOBE know that an Ubuntu WSL instance is available (registered) so it can start the Subiquity server (assuming we are talking about the setup workflow).


packages/ubuntu_wsl_setup/.metadata
```diff
@@ -1,10 +1,30 @@
 # This file tracks properties of this Flutter project.
 # Used by Flutter tool to assess capabilities and perform upgrades etc.
 #
-# This file should be version controlled and should not be manually edited.
+# This file should be version controlled.
 
 version:
-  revision: dd24cf037b8443011816c7eadc0ac567cf54cb5d
-  channel: autocomplete-rebuild-options
```
jpnurmi, line 8: haha, that'd be me https://github.com/flutter/flutter/pull/88193

packages/ubuntu_wsl_setup/windows/runner/main.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+  std::vector<std::string> command_line_arguments =
+      GetCommandLineArguments();
+
+  project.set_dart_entrypoint_arguments(std::move(command_line_arguments));
+
+  FlutterWindow window(project);
+  Win32Window::Point origin(10, 10);
+  Win32Window::Size size(1280, 720);
+  if (!window.CreateAndShow(L"ubuntu_wsl_setup", origin, size)) {
```
jpnurmi, line 30: should it be 960x680 and what about the title? these can be addressed in a separate PR if you prefer to import the template without modifications, first

jpnurmi (approved):

packages/ubuntu_wsl_setup/windows/runner/main.cpp
CarlosNihelton: Planned for another PR

PR #972, opened 2022-07-06
Set coding style for native code

CarlosNihelton:
Set the standard window size and adds a custom `.clang-format` file for consistency when editing C++ source files, imported from [flutter/plugins](https://github.com/flutter/plugins/blob/main/.clang-format) and [flutter/packages](https://github.com/flutter/packages/blob/main/.clang-format):
```
BasedOnStyle: Google
---
Language: Cpp
DerivePointerAlignment: false
PointerAlignment: Left
```

The files inside the `**/linux` subdirectories won't change at all when formatted using clang-format. The Windows app template, on the other hand, doesn't seem to follow a specific coding style, thus besides submitting a `.clang-format` configuration file, this PR also formats the non-compliant files.

### DIFF
--- a/packages/ubuntu_wsl_setup/windows/runner/main.cpp
+++ b/packages/ubuntu_wsl_setup/windows/runner/main.cpp
@@ -6,7 +6,7 @@
 #include "utils.h"
 
 int APIENTRY wWinMain(_In_ HINSTANCE instance, _In_opt_ HINSTANCE prev,
-                      _In_ wchar_t *command_line, _In_ int show_command) {
+                      _In_ wchar_t* command_line, _In_ int show_command) {
   // Attach to console when present (e.g., 'flutter run') or create a
   // new console when running with a debugger.
   if (!::AttachConsole(ATTACH_PARENT_PROCESS) && ::IsDebuggerPresent()) {
@@ -19,8 +19,7 @@ int APIENTRY wWinMain(_In_ HINSTANCE instance, _In_opt_ HINSTANCE prev,
 
   flutter::DartProject project(L"data");
 
-  std::vector<std::string> command_line_arguments =
-      GetCommandLineArguments();
+  std::vector<std::string> command_line_arguments = GetCommandLineArguments();
 
   project.set_dart_entrypoint_arguments(std::move(command_line_arguments));
 
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,5 @@
+BasedOnStyle: Google
+---
+Language: Cpp
+DerivePointerAlignment: false
+PointerAlignment: Left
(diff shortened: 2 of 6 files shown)


packages/ubuntu_wsl_setup/windows/runner/.clang-format
```diff
@@ -0,0 +1,8 @@
+# Format Style Options - Created with Clang Power Tools
```
jpnurmi, line 1: [flutter/plugins](https://github.com/flutter/plugins/blob/main/.clang-format) and [flutter/packages](https://github.com/flutter/packages/blob/main/.clang-format) both have:
```
BasedOnStyle: Google
---
Language: Cpp
DerivePointerAlignment: false
PointerAlignment: Left
```
could you also apply the format on all files so later changes won't include unrelated formatting changes?

packages/ubuntu_wsl_setup/windows/runner/main.cpp
```diff
@@ -26,7 +26,7 @@ int APIENTRY wWinMain(_In_ HINSTANCE instance, _In_opt_ HINSTANCE prev,
 
   FlutterWindow window(project);
   Win32Window::Point origin(10, 10);
-  Win32Window::Size size(1280, 720);
+  Win32Window::Size size(960, 680);
```
jpnurmi, line 29: - does this only change the default size? should there be any min/max/fixed size constraints because our slides won't scale to arbitrary sizes?
- could this be moved to a separate PR because we might want to include the respective formatting changes with the .clang-format config?

packages/ubuntu_wsl_setup/windows/runner/.clang-format
jpnurmi: i'd propose that we put the `.clang-format` all the way to the repo root. later on, we could even make the CI format C++ code too i.e. run `flutter-plugin-tools format` instead of just `flutter format`

CarlosNihelton: I really like the idea!

CarlosNihelton: > [flutter/plugins](https://github.com/flutter/plugins/blob/main/.clang-format) and [flutter/packages](https://github.com/flutter/packages/blob/main/.clang-format) both have:
> 
> ```
> BasedOnStyle: Google
> ---
> Language: Cpp
> DerivePointerAlignment: false
> PointerAlignment: Left
> ```
> 
> could you also apply the format on all files so later changes won't include unrelated formatting changes?

I started with that and in between rebases I forgot to edit the commit message. The app template code doesn't seem to follow any specific standard so adopting one that favors one set of files causes format changes to others. The solution really is choosing one thing and reformatting the whole source set at once. That makes more sense on its own PR.

packages/ubuntu_wsl_setup/windows/runner/main.cpp
CarlosNihelton: Good point. That makes the size-related changes more interesting to put on a separate PR.

jpnurmi (approved):
Thanks, LGTM! :+1:

PR #973, opened 2022-07-06
[WSL] Named event API

CarlosNihelton:
Win32 offers a concept of a named event, which can be used for one-shot IPC, where a process can `CreateEvent` and other processes can register themselves to get notified when the event is set. It cannot carry arbitrary payloads, instead is a kind synchronisation object similar to a mutex or condition variable that has a name. There are also unnamed events as part of the API, but those are out of scope of this PR. [See the MS docs about event objects](https://docs.microsoft.com/en-us/windows/win32/sync/event-objects) for more information.

Built on top of that concept, `NamedEvent` Dart class monitors a platform event selected by name and offers a `Future<void>` getter than completes when the event is set somewhere else. That's useful when IPC'ing with low level code like the WSL distro launcher.

Listening to one platform named event is done in the native side by registering a callback that will be triggered at maximum once. The operating system thread pool provides a worker thread that will execute the callback once the event is set. To comply with the desired API for `NamedEvent` the callback's job is to `invokeMethod` with the event name as the payload. The handler lives in the Dart side and just completes the appropriate future.

Lifetime of the native `NamedEvent` objects is handled automatically by the FlutterWindow (by storing them inside a std::vector field) and the lifetime of the internal resources of a native `NamedEvent` (HANDLE for the event and for the waiting thread subscription) is handled by the the construction and destruction of the referred object (RAII, as we like to say in C++ coffee shops). It means that whether they are ever triggered or not they will last until the FlutterWindow object is destroyed.

A singleton Dart class (alla plugin style) manages the link between which event names are being listened to and which callbacks must be triggered once one event is set.

For this application only one event is planned to be listened to, but I wanted to make the implementation flexible so we can find other uses for that in the future and maybe extract a plugin out of it. Maybe in the future we can build a similar API on top of  `eventfd` for Linux (exposing names instead of file descriptors to stay compliant with the current API).

This removes one `TODO` item from the WSL setup 😅


packages/ubuntu_wsl_setup/windows/runner/flutter_window.cpp
```diff
@@ -59,3 +131,63 @@ FlutterWindow::MessageHandler(HWND hwnd, UINT const message,
 
   return Win32Window::MessageHandler(hwnd, message, wparam, lparam);
 }
+
+VOID NamedEvent::CallbackForWait(PVOID context, BOOLEAN) {
```
jpnurmi, line 135: could it be in a separate `named_event.cpp`?

packages/ubuntu_wsl_setup/test/named_event_test.dart
```diff
@@ -0,0 +1,45 @@
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:ubuntu_wsl_setup/services/named_event.dart';
+
+void main() {
+  const methodChannel = MethodChannel(NamedEventConstants.channel);
+  TestWidgetsFlutterBinding.ensureInitialized();
+  final binding =
```
jpnurmi, line 8: s/binding/messenger/

packages/ubuntu_wsl_setup/windows/runner/flutter_window.cpp
```diff
@@ -59,3 +131,63 @@ FlutterWindow::MessageHandler(HWND hwnd, UINT const message,
...
+  if (waitHandle != nullptr) {
+    [[maybe_unused]] auto res = UnregisterWait(waitHandle);
+    // Win32 Docs: Note that a wait handle cannot be used in functions that
+    // require an object handle, such as CloseHandle.
+  }
+
+  if (event != nullptr) {
+    CloseHandle(event);
+  }
+}
```
jpnurmi, line 193: :no_entry: i have agree with github :grin: any chance to have files ending with a newline?

packages/ubuntu_wsl_setup/windows/runner/flutter_window.cpp
```diff
@@ -59,3 +131,63 @@ FlutterWindow::MessageHandler(HWND hwnd, UINT const message,
...
+
+  if (event == nullptr || event == INVALID_HANDLE_VALUE) {
+    // the most likely error
+    if (GetLastError() == ERROR_FILE_NOT_FOUND) {
+      throw std::invalid_argument(name);
+    } else {
+      throw std::runtime_error("Cannot open event");
+    }
+  }
+  if (FALSE == RegisterWaitForSingleObject(
```
jpnurmi, line 173:
```suggestion
  if (!RegisterWaitForSingleObject(

```

packages/ubuntu_wsl_setup/windows/runner/flutter_window.cpp
```diff
@@ -59,3 +131,63 @@ FlutterWindow::MessageHandler(HWND hwnd, UINT const message,
...
+                   &waitHandle, event, &CallbackForWait,
+                   reinterpret_cast<void*>(this), INFINITE,
```
jpnurmi, line 175:
```suggestion
                   this, INFINITE,

```

packages/ubuntu_wsl_setup/lib/services/named_event.dart
```diff
@@ -0,0 +1,62 @@
...
+    );
+  }
+}
+
+@visibleForTesting
+class NamedEventConstants {
+  static const channel = 'ubuntuWslSetupChannel';
+  static const onEventSet = 'onEventSet';
+  static const addListenerFor = 'addListenerFor';
+  static const argKey = 'eventName';
```
jpnurmi, line 27: we could get rid of this if we pass the event name directly

packages/ubuntu_wsl_setup/lib/services/named_event.dart
```diff
@@ -0,0 +1,62 @@
...
+  final MethodChannel _channel =
+      const MethodChannel(NamedEventConstants.channel);
+
+  Future<void> _methodCallHandler(MethodCall call) async {
+    if (call.method == NamedEventConstants.onEventSet) {
+      String eventName = call.arguments[NamedEventConstants.argKey];
+      _listeners[eventName]?.call();
+    } else {
+      throw UnimplementedError();
+    }
```
jpnurmi, line 47:
```suggestion
    assert(call.method == NamedEventConstants.onEventSet);
    _listeners[call.arguments]?.call();
```
unless there are plans for more methods and/or arguments?

packages/ubuntu_wsl_setup/lib/services/named_event.dart
```diff
@@ -0,0 +1,62 @@
...
+    required void Function() listener,
+  }) {
+    if (_listeners.containsKey(name)) {
+      throw ArgumentError('Cannot listen to the same event "$name" twice');
+    }
+    _channel.invokeMethod(NamedEventConstants.addListenerFor, name);
+    _listeners[name] = listener;
+  }
+
+  final _listeners = HashMap<String, void Function()>();
```
jpnurmi, line 61:
```suggestion
  final _listeners = <String, void Function()>{};
```

packages/ubuntu_wsl_setup/lib/services/named_event.dart
```diff
@@ -0,0 +1,62 @@
...
+
+/// Exposes a [future] that completes when a platform event named [name] is set.
+class NamedEvent {
+  /// The event name as known by the platform, UTF-8 encoded.
+  String name;
+  final _completer = Completer<void>();
+
+  /// Completes when the platform event is set.
+  Future<void> get future => _completer.future;
+  NamedEvent(this.name) {
```
jpnurmi, line 14: - could be const after making name final
- add empty line before the constructor

packages/ubuntu_wsl_setup/windows/runner/flutter_window.h
```diff
@@ -28,6 +31,39 @@ class FlutterWindow : public Win32Window {
...
+// only once which waits on another thread and executes the supplied [callback]
+// once the event is first set.
+class NamedEvent {
+ private:
+  std::function<void()> callback;
+  HANDLE event = nullptr;
+  HANDLE waitHandle = nullptr;
+  static VOID CALLBACK CallbackForWait(PVOID context, BOOLEAN);
+
+ public:
```
jpnurmi, line 62: Flutter follows the Google style: public first, then protected, then private

packages/ubuntu_wsl_setup/windows/runner/utils.cpp
```diff
@@ -61,3 +61,24 @@ std::string Utf8FromUtf16(const wchar_t* utf16_string) {
...
+  }
+  utf16_string.resize(target_length);
+  int converted_length =
+      ::MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, utf8_string.data(),
+                            cBytes, utf16_string.data(), target_length);
+  if (converted_length == 0) {
+    return std::wstring();
+  }
+  return utf16_string;
+}
```
jpnurmi, line 84: newline

jpnurmi (commented):
sorry, i spammed it full of comments but they are all minor things :smile:

jpnurmi:
hmm, we should add a CI job that runs `flutter build windows` for ubuntu_wsl_setup on windows :)

packages/ubuntu_wsl_setup/lib/services/named_event.dart
jpnurmi: brainfart, cannot be const with a body :D

packages/ubuntu_wsl_setup/windows/runner/named_event.cpp
```diff
@@ -0,0 +1,66 @@
+#include "named_event.h"
+
+#include <flutter/standard_method_codec.h>
```
jpnurmi, line 3: flutter includes not needed in named_event.cpp?

jpnurmi (approved):
There may be some unnecessary Flutter includes in `named_event.*` but otherwise LGTM! :+1:

packages/ubuntu_wsl_setup/lib/services/named_event.dart
CarlosNihelton: No plans indeed.

packages/ubuntu_wsl_setup/windows/runner/flutter_window.cpp
CarlosNihelton: :D

packages/ubuntu_wsl_setup/lib/services/named_event.dart
CarlosNihelton: Indeed

packages/ubuntu_wsl_setup/lib/services/named_event.dart
CarlosNihelton: Still the String name being final is useful. 😅

packages/ubuntu_wsl_setup/windows/runner/flutter_window.cpp
```diff
@@ -1,8 +1,18 @@
 #include "flutter_window.h"
 
 #include <optional>
+#include <stdexcept>
 
 #include "flutter/generated_plugin_registrant.h"
+#include "flutter/standard_method_codec.h"
```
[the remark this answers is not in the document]
CarlosNihelton: Cool

CarlosNihelton:
> hmm, we should add a CI job that runs `flutter build windows` for ubuntu_wsl_setup on windows :)

It's coming...

packages/ubuntu_wsl_setup/test/named_event_test.dart
jpnurmi: > I'm struggling with that one.

Sorry for the confusion. I meant just the variable name because it's not the binding but a messenger. :)

> `TestWidgetsFlutterMessenger` doesn't exist. sweat_smile

Hmm, it's defined in flutter_test. Did you have the words in the correct order? :smile: 

> `TestDefaultBinaryMessengerBinding.instance.initInstances();` is not what I'm looking for because instance can be null. `initInstances()` is not static.
> 
> What is the correct call, then?

Not sure what you prefer but we could even combine it into one line since we're specifically interested in the messenger:

```dart
final messenger =
      TestWidgetsFlutterBinding.ensureInitialized().defaultBinaryMessenger;
```

jpnurmi: I never remember how to spell TestWidgetsFlutterBinding, it's always try-n-error with the auto-completer... :smile:

packages/ubuntu_wsl_setup/windows/runner/named_event.cpp
CarlosNihelton: Thanks for catching those. :D

packages/ubuntu_wsl_setup/test/named_event_test.dart
jpnurmi: Oops, sorry! Sed :smile:

CarlosNihelton:
> > It's coming...
> 
> Leave it for later, this is good to go? :wink: 

Yes, it is!

Issue #819, opened 2022-05-04
Screen 8.5 - Install Alongside


jpnurmi:
![image](https://user-images.githubusercontent.com/140617/166716193-2961bc49-745e-4a03-a762-5a420a972b2a.png)

jpnurmi:
![Screenshot from 2022-05-06 09-41-12](https://user-images.githubusercontent.com/140617/167090057-1ac51117-0914-4e1d-8bb7-c0e1bc53c4eb.png)

jpnurmi:
Notes to self. In Ubiquity
- if there are multiple OSes detected, there's no way to install alongside either of them even if they are on separate disks
- Ubiquity favors reinstalling or replacing an existing Ubuntu if there's already Windows + Ubuntu installed

![VirtualBox_Ubuntu 21 10_06_05_2022_13_54_55](https://user-images.githubusercontent.com/140617/167126964-feef6b12-80e0-4537-b2d6-c28584951c14.png)

corradoventu:
On my desktop I have 5 different linux installations on 3 different disk, the current way (like gparted) to present the partitions used by ubiquity seems much better.

jpnurmi:
@corradoventu All other screenshots above except the first one are from Ubiquity. We haven't implemented guided resizing for the new installer yet. Did you mix this with the [manual partitioning](https://github.com/canonical/ubuntu-desktop-installer/issues/36) page?

jpnurmi:
There are some issues with the implementation in Ubiquity that we should aim to fix in the new installer:
- The splitter type of UI control for resizing partitions supports only mouse interaction. There's no way to interact with keys because it's not focusable at all ==> not accessible.
- The option to install alongside existing OS is only available when a single existing OS is detected. If there are multiple OSes detected, the option is not shown but one can only replace or reinstall => where/why does this limitation come from?

corradoventu:
Sorry I got confused with the manual partitioning page. Thanks.

PR #976, opened 2022-07-07
Promote OS probe results to DiskStorageService

jpnurmi:
In addition to the "Installation type" page, the same information about existing OS installations is also needed on the upcoming "Install X alongside Y" page. We can share the logic via DiskStorageService.

Ref: #819


packages/ubuntu_desktop_installer/lib/pages/installation_type/installation_type_model.dart
```diff
@@ -74,22 +73,10 @@ class InstallationTypeModel extends SafeChangeNotifier {
...
-                .whereType<OsProber>(),
-          )
-          .toList();
-      notifyListeners();
-    });
-  }
+  List<OsProber>? get existingOS => _diskService.existingOS;
+
+  /// Initializes the model.
+  Future<void> init() async {}
```
jpnurmi, line 79: i'm leaving an empty method here because soon there will be initialization steps again :)

CarlosNihelton (approved):
It's quite hard to follow the dance of the new partitioning API, but the end result is becoming really nice!

PR #977, opened 2022-07-08
Write changes to disk: present all non-preserved entries

jpnurmi:
Collect the list of non-preserved partitions from all disks, not only from non-preserved disks.

Before:
- collect non-preserved disks, and then non-preserved partitions from them

After:
- collect non-preserved disks
- collect non-preserved partitions from all disks

This is necessary for the upcoming guided installation alongside an existing OS, where the entire disk won't be reformatted and thus, is considered preserved.

Ref: #819


CarlosNihelton (approved):
LGTM

PR #978, opened 2022-07-08
[UDI] port to the new target-based guided storage API

jpnurmi:
The options for guided storage are no longer presented as disks but as "storage targets" that can a) reformat a disk, b) resize an existing partition, or c) use a free gap. This PR applies the necessary modifications to move over to the new target-based API, but does not yet add handling for the new target types.

Ref: #819


jpnurmi:
Manual live installation with guided partitioning passed :+1: 

Screenshot from 2022-07-08 13-38-13
Screenshot from 2022-07-08 13-41-18

CarlosNihelton (approved):
Thanks for splitting the a), b) and c) cases in different pull requests. It would be very hard for me to follow on what's going on.

jpnurmi:
Thanks so much for helping out and reviewing these storage changes! Very much appreciated! I know they're hard to make sense of without having been involved in the work. I'm trying to split the changes into small chunks to make them reasonable to glance through.

PR #980, opened 2022-07-08
Build workflow for the WSL OOBE target

CarlosNihelton:
We can discuss later about the other pieces of a good CI workflow, but at least since we are touching native files which may introduce silent failures, attempting to build the Windows entry point app when relevant pull requests are opened should be a good starting point.

This filters out PR's that don't touch either the `.github/workflows` directory as well as the packages `ubuntu_wsl_setup`, `ubuntu_wizard` and `subiquity_client`.

### DIFF
--- a/.github/workflows/flutter-build-win.yml
+++ b/.github/workflows/flutter-build-win.yml
@@ -1,0 +1,43 @@
+name: Windows WSL Setup CI
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - packages/subiquity_client/**
+      - packages/ubuntu_wizard/**
+      - packages/ubuntu_wsl_setup/**
+      - .github/workflows/**
+  pull_request:
+    branches:
+      - main
+    paths:
+      - packages/subiquity_client/**
+      - packages/ubuntu_wizard/**
+      - packages/ubuntu_wsl_setup/**
+      - .github/workflows/**      
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: windows-latest
+    steps:
+    
+    - uses: actions/checkout@v2
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/flutter-build-win.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - uses: subosito/flutter-action@v1
+      with:
+        channel: 'stable'
+
+    - name: Install Melos
+      run: flutter pub global activate melos
+
+    - name: Bootstrap workspace
+      run: melos bootstrap
```
jpnurmi, line 39: we could skip melos for this job because we only need to build one package. melos would unnecessarily slow it down by running flutter pub get for all packages.

.github/workflows/flutter-build-win.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: Build Windows Target
+      working-directory: packages\ubuntu_wsl_setup\
+      run: flutter build -t .\lib\main_win.dart
```
jpnurmi, line 43: flutter build \_windows\_?

.github/workflows/flutter-build-win.yml
CarlosNihelton: Indeed! I thought I'd need melos to bring ubuntu_wizard and subiquity_client, but `flutter pub get` is enough. Thanks for the advice.

.github/workflows/flutter-build-win.yml
CarlosNihelton: Thanks for noticing! :D

jpnurmi (approved):
Thanks, LGTM :+1:

PR #986, opened 2022-07-12
Allocate disk space: remove a hard-coded height to fix overflow

jpnurmi:
A hard-coded 48px value was used because a horizontal ListView doesn't
calculate intrinsic height. Use SingleChildScrollView + Row to ensure
that the height is calculated as appropriate. This is fine for a list of
partition labels because the amount is never going to be awfully high.


jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2022-07-12 16-46-56 | Screenshot from 2022-07-12 16-45-57 |

CarlosNihelton (approved):
LGTM

PR #991, opened 2022-07-13
Generalize and promote StorageSizeBox

jpnurmi:
The widget was previously only used in _Allocate disk space_ but it would be useful in _Install X alongside Y_ too. This PR promotes it up in the project tree and adds a few more properties to make it more general purpose.


jpnurmi:
image

CarlosNihelton (approved):
Just one minor nitpickI: 'd have named it `StorageSizeField` in tandem with `TextField`, for instance. But I can see we don't support stuff like an editing controller, so feel free to disregard this comment.

jpnurmi:
Hmm, it was named in tandem with `SpinBox`, which does internally contain a `TextField`, though :smile:

PR #992, opened 2022-07-13
Guided install alongside other OSes

jpnurmi:
https://user-images.githubusercontent.com/140617/178792980-8f4d7b9f-7487-4e95-893b-40ecc8e83e76.mp4

Close: #819

PR #998, opened 2022-07-14
Add StorageButton & StorageIcon widgets

jpnurmi:
These are small standalone helper widgets extracted from #992 for a more pleasant review experience. :smile: 

image


CarlosNihelton (approved):
Thanks for splitting those out. :D

PR #1009, opened 2022-07-18
StorageSplitView: apply new weights and limits on rebuild

jpnurmi:
When the user selects another guided storage i.e. the split view gets
rebuilt with a different set of storage size constraints, make sure to apply
the new weights and limits by rebuilding the split view controller.


CarlosNihelton (approved):
LGTM! The test is even more interesting than the main source code. :D

PR #1014, opened 2022-07-19
Fix SelectGuidedStorageModel.getDisk(index)

jpnurmi:
This PR fixes a typo that slipped in in #978. Instead of using the specified index, it would always use the currently selected one, which resulted in the currently selected drive being repeated in the dropdown:

| Before | After |
|---|---|
| Screenshot from 2022-07-19 13-46-30 | Screenshot from 2022-07-19 13-46-15 |


CarlosNihelton (approved):
LGTM

PR #1015, opened 2022-07-19
Write changes to disk: summarize all partition changes

jpnurmi:
This PR makes the partition change summary a bit more elaborate to give the user a better understanding of the changes being written. Furthermore, the changes are formatted with bullet points to make the summary easier to glance through.

Previously, we would list all partitions with `preserved=false` and phrase it as being "formatted". This proposal takes all the following partition attributes into account:
- `bool resize`
- `String wipe`
- `String mount`
- `bool preserve`

The proposal is to format the summary as such:

> "The following partition changes are going to be applied:"
>
> - "partition #{disk}{partition} resized from {oldsize} to ${newsize}" (`resize` is true)
> - "partition #{disk}{partition} formatted as {format} used for {mount}" (`wipe` and `mount` are non-empty)
> - "partition #{disk}{partition} formatted as {format}" (`wipe` is non-empty)
> - "partition #{disk}{partition} used for {mount}" (`mount` is non-empty)
> - "partition #{disk}{partition} created" (`preserve` is false)

Screenshot from 2022-07-19 13-59-00

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
+++ b/packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
@@ -347,10 +347,20 @@
       "path": {}
     }
   },
-  "writeChangesPartitionsHeader": "The following partitions are going to be formatted:",
-  "writeChangesPartitionEntryMounted": "partition #{disk}{partition} as {format} used for {mount}",
-  "@writeChangesPartitionEntryMounted": {
-    "description": "A mounted partition entry",
+  "writeChangesPartitionsHeader": "The following partition changes are going to be applied:",
+  "writeChangesPartitionResized": "partition #{disk}{partition} resized from {oldsize} to {newsize}",
+  "@writeChangesPartitionResized": {
+    "description": "A resized partition entry",
+    "placeholders": {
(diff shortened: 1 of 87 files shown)


jpnurmi:
Design acked by @elioqoshi:
> I think I mentioned this to JP before but it's a low hanging fruit improvement and quite better already!
> It allows for easier scanning as body text and volume names are easier to distinguish now

CarlosNihelton (approved):
Well done! It's a nice design improvement.

corradoventu:
Installing from canary ISO Ubuntu 22.10 "Kinetic Kudu" - Alpha amd64 (20220816) subiquity 588 on the preallocated partition sdb2 screen 'Write changes to disk' says partitions nvme0n11 and sdb2 are used (which is correct) but sdb2 should be formatted because contains some old data, and is NOT formatted.
Screen  'Allocate disk space' and 'Edit partition' have a check box to select 'format' but the check box is inactive.
Installation log can be found in https://bugs.launchpad.net/subiquity/+bug/1986676

Screenshot from 2022-08-16 14-06-55
Screenshot from 2022-08-16 14-06-00

PR #1023, opened 2022-07-21
Installation slides: clean up unused installation "steps"

jpnurmi:
The temporary installation slides page added in #186 used to show a text
label e.g. "3/5" to indicate installation progress. The label has not
been there for a good while anymore so the logic is not needed either.
Some of the application statuses used for implementing the unused logic
are being removed from Subiquity.


CarlosNihelton (approved):
LGTM

PR #1025, opened 2022-07-21
Not enough disk space

jpnurmi:
image

Design: https://people.canonical.com/~platform/desktop/installer/12.%20Write%20Changes%20to%20Disk/#artboard1
Specs: https://docs.google.com/document/d/1bZ4yQIVgGaUGSYu3qiUHnQt3ieBZoqunP_DcleHCr3I/preview#heading=h.77dcshyn3tzm
Ref: #440

Issue #27, opened 2021-03-08
Screen 2 - installation type


seb128:
The design for screen n°2, installation type

![try-or-install](https://user-images.githubusercontent.com/10974330/110313110-47eff780-8006-11eb-93fd-3d1379939d36.png)

Design details https://people.canonical.com/~platform/desktop/installer/2.%20Installation%20Type/

fitojb:
I’m afraid those icons look a bit too wireframe-y. I would expect filled shapes with the Suru color palette and folds.

fitojb:
Also, the Ubuntu logo is used excessively. I remember @matthewpaulthomas advising against that many years ago…

fitojb:
I’m afraid those icons look a bit too wireframe-y. I would expect filled shapes with the Suru color palette and folds.
Also, the Ubuntu logo is used excessively. I remember @matthewpaulthomas advising against that many years ago…

BillDietrich:
Is some "preseed" capability going to be supported ?  I would like it to be supported.  And there should be an option to read the preseed config file from anywhere, including another USB drive (not same as installer USB).  I don't know if it should affect the UI of this page, or some other.  Thanks.

Mannshoch:
I hope so -> #16

Feichtmeier:
@oSoMoN 

could you eventually close the issues that have been implemented? This could help a bit to keep track for external contributors 😊

oSoMoN:
You're right. Implementation isn't 100% complete, but it's good enough to consider closing this issue I suppose.

Mannshoch:
Could you please give me a short feedback about your thinking for #28

jpnurmi:
- Do we have any specs for _Repair installation_?
- Does the _Try Ubuntu_ option do anything else besides closing the installer?

PR #1027, opened 2022-07-21
[WSL] Request reboot on Applying changes page

CarlosNihelton:
Instead of just closing the window. `SystemShutdown.reboot` itself closes the window after talking to the server.

When running the OOBE on windows this prevents the server to stay alive after the GUI exits.
When running on Linux it might help with the intermittent CI failure due a future that never completes (could the application exit and kill the server at some improper moment?).

Consider that the server doesn't have the power to reboot the distro under WSL, instead it leaves some breadcrumbs guiding the launcher to "reboot" the distro, action taken on the Windows side.

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_model.dart
+++ b/packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_model.dart
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'package:flutter/widgets.dart';
 import 'package:safe_change_notifier/safe_change_notifier.dart';
 import 'package:subiquity_client/subiquity_client.dart';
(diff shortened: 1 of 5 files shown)


jpnurmi:
I'm not sure how we want

- the "Try" button on the [Try or install](https://github.com/canonical/ubuntu-desktop-installer/issues/27) page, and
- the "Quit" button on the [Not enough disk space](https://github.com/canonical/ubuntu-desktop-installer/pull/1025) page

to behave but a clean Subiquity exit without reboot or shutdown could potentially be useful.

CarlosNihelton:
I propose we move forward with this one while I find some time to make that option available in Subiquity, shall the other maintainers not manifest themselves against the idea. I like Dan's idea.

jpnurmi (approved):
I owe you a beer if this fixes the integration test :)

PR #1030, opened 2022-07-26
Revert "Upgrade to split_view 3.2.0"

CarlosNihelton:
Reverts canonical/ubuntu-desktop-installer#1016.

Version 3.2.0 of the `split_view` package has been temporarily removed from pub.dev due [a bug report](https://github.com/toshiaki-h/split_view/issues/19), and since there is no such version CI keeps failing.

This temporarily reverts back to [JP's branch of split_view](https://github.com/jpnurmi/split_view/tree/controller-weights) until the upstream is settled.


robert-ancell (approved):
Makes sense. It might be useful to open an issue for this and add that as a comment into the pubspec.yaml and/or the commit message so this can be more easily fixed in the future.

PR #1032, opened 2022-07-28
[WSL] Let the launcher request graceful window close

CarlosNihelton:
Since the WSL launcher program is parent of the OOBE, it can terminate the process at any moment, but that is too dramatic.
There are situations where the launcher may request the OOBE to gracefully exit.
Listening to that `NamedEvent ('Local\\${options["distro-name"]}-close-oobe')` and closing the Window when set is enough for that purpose, since `closeWindow()` takes care of the server.


jpnurmi (approved):
LGTM :+1:

PR #1034, opened 2022-07-28
Nullable instead of late Endpoint.

CarlosNihelton:
Server may fail to start or the app be closed before the server starts (dryrun or WSL cases).
Without the server, endpoint will never be set, causing LateInitializationError to be thrown.
Making the fields nullable avoid such class of error.

### DIFF
--- a/packages/subiquity_client/lib/src/client.dart
+++ b/packages/subiquity_client/lib/src/client.dart
@@ -42,18 +42,16 @@ class SubiquityException implements Exception {
 class SubiquityClient {
   final _client = HttpClient();
   Endpoint? _endpoint;
-  String? _host;
   final _isOpen = Completer<bool>();
 
   Future<bool> get isOpen => _isOpen.future;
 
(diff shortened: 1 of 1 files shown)


packages/subiquity_client/lib/src/client.dart
```diff
@@ -41,14 +41,14 @@ class SubiquityException implements Exception {
 
 class SubiquityClient {
   final _client = HttpClient();
-  late Endpoint _endpoint;
-  late String _host;
```
jpnurmi, line 45: we could possibly remove the `_host` member and use `_endpoint.authority` instead?

jpnurmi (approved):
LGTM :+1:

packages/subiquity_client/lib/src/client.dart
CarlosNihelton: Yes, we can. It's only used in one place.

jpnurmi (approved):
LGTM :+1:

PR #1035, opened 2022-07-29
[WSL] Only defer server initialization in setup

CarlosNihelton:
If the command line '--reconfigure' is explicitely set, then we should not defer subiquity initialization.
This makes the WSL launcher side simpler.
Deferring is meant only for setup, when we need to wait for registration completion before attempting to run any Linux process. For reconfiguration we don't need to wait for anything, since the GUI is responsible for starting the server.

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/main_win.dart
+++ b/packages/ubuntu_wsl_setup/lib/main_win.dart
@@ -56,7 +56,7 @@ Future<void> main(List<String> args) async {
       '--server-only',
       '--tcp-port=${socketHolder.port}'
     ],
-    defer: options['reconfigure'] ? null : registrationEvent.future,
+    defer: options['reconfigure'] == true ? null : registrationEvent.future,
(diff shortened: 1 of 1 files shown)


packages/ubuntu_wsl_setup/lib/main_win.dart
```diff
@@ -56,7 +56,7 @@ Future<void> main(List<String> args) async {
       '--server-only',
       '--tcp-port=${socketHolder.port}'
     ],
-    defer: registrationEvent.future,
+    defer: options['reconfigure'] ? null : registrationEvent.future,
```
jpnurmi, line 59: should this compare to true because `ArgResults.operator[]` returns a dynamic and Dart doesn't have an implicit conversion from null to boolean?

CarlosNihelton: You're absolutely right.

jpnurmi (approved):
LGTM :+1:

PR #1036, opened 2022-08-02
Enables generating full PDB for the WSL app.

CarlosNihelton:
By default debug symbols are not exported for the Flutter application runner. Per instructions found in [this comment](https://github.com/flutter/flutter/issues/98863#issuecomment-1047178402) on the Flutter framework repository, this one-liner enables exporting the debug symbols in a program debug database (.pdb) file.

The WSL launcher will be prepared to collect such artifact as part of the release process, providing us with a better crash analysis capabilities, shall we need it.

Ideally we should also collect those files for the plugins as well (those that generate `dll`s). [Per this comment](https://github.com/flutter/flutter/issues/105997#issuecomment-1179230032) it seems we lack a way to propagate that option from the app to the required plugins. So, for now, let's focus on the app.


jpnurmi (approved):
LGTM :+1:

PR #1041, opened 2022-08-08
Use canonical/udev.dart

jpnurmi:
There are more use cases for udev outside the installer so a standalone
(and more complete) udev package was born:

- https://github.com/canonical/udev.dart
- https://pub.dev/packages/udev


CarlosNihelton (approved):
That's really cool!

PR #1051, opened 2022-08-10
Disable unusable partition gaps 

jpnurmi:
This change greys out and annotates unusable gaps with tooltips in the storage table, and disables the "+" button to prevent the user from trying to create partitions into unusable gaps, which would result in `Exception: Exceeded number of available partitions` being thrown under the hood.

| Before | After |
|---|---|
| Screenshot from 2022-08-10 10-07-34 | Screenshot from 2022-08-10 10-06-59 |

EDIT: There's an updated screenshot of the final version below: https://github.com/canonical/ubuntu-desktop-installer/pull/1051#issuecomment-1422480453


elioqoshi:
I get where you are coming from but disabling the row like that has some accessibility issues, mostly the lack of contrast of labels.

Maybe consider this alternative:

grafik

- strikethrough icon
- grayed out text, but still enough contrast
- label on the last column messaging it's disabled

The SVG icon for the disabled harddrive is attached here (but we might want to add it to Yaru properly)


harddisk-disabled

Jupi007:
I created a PR at `yaru_icons.dart`: https://github.com/ubuntu/yaru_icons.dart/pull/66

I personally recommend to use the **warning** or the **error** variation :)
Because the strikethrough mean _disabled_ in our icons language.

Jupi007:
New icons are available in `yaru_icons: 0.2.5`.

jpnurmi:
@Jupi007 Thank you so much for the new icons! :heart_eyes: 

@elioqoshi, inserting a "Limit reached" label into the last column makes the whole (previously narrow) column grow in width. The text may grow significantly longer when translated into other languages.

| Before | After |
|---|---|
| Screenshot from 2022-10-12 09-46-07 | Screenshot from 2022-10-12 09-47-59 |

There are three existing columns that may already contain more or less arbitrary strings:
- Device: "/foo/bar123", "free space"
- Mount point: "/path/to/somewhere"
- System: "Windows Boot Manager"

Would it be possible to reuse one of these columns to avoid stealing space with the Format column that currently only has checkboxes? Perhaps we could simply change the "free space" text?

image

P.S. On a related note, I'm not sure if "limit reached" alone is descriptive enough to explain the situation to the user.

jpnurmi:
Screenshot from 2023-02-08 12-56-05

d-loose (approved):
LGTM :+1:

PR #1052, opened 2022-08-11
Fix telemetry report location

jpnurmi:
Write the report to `/var/log` instead of `/target/var/log` because the GUI
has no permission to write to the latter. The report will be copied to
its final place on the target system by the upcoming post-install script.
Furthermore, in debug/dry-run mode, write to the app dir similarly to
the logs.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/services.dart
+++ b/packages/ubuntu_desktop_installer/lib/services.dart
@@ -4,5 +4,5 @@ export 'services/disk_storage_service.dart' hide log;
 export 'services/journal_service.dart';
 export 'services/network_service.dart';
 export 'services/power_service.dart';
-export 'services/telemetry_service.dart';
+export 'services/telemetry_service.dart' hide log;
 export 'services/udev_service.dart';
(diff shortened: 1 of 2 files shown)


oSoMoN (commented):
The changes make sense and look good to me.

There's one thing I'm unsure about though: if we write the telemetry report when the installation slides begin, then the "Stages" section will be missing an entry for "/installationcomplete", right? It might be an acceptable trade-off (considering that the current implementation doesn't actually work anyway), but it would be a regression compared to the current installer that records in its telemetry data when the installation is complete, and when it reboots.

jpnurmi:
Good point, I wonder if we should integrate this closer with Subiquity because we won't be able to write anything from the GUI when Subiquity reports "done" status. Also, the "done" timestamp is wrong with the current proposal. I was too focused on getting the file copied on the target. :smile:

jpnurmi:
@oSoMoN Let's take a step back and have this PR only fix the report location. Getting it written in time with all Subiquity states logged is a challenge of its own.

oSoMoN (approved):
LGTM, thanks!
Can you file a follow-up issue to track improved integration with Subiquity states?

jpnurmi:
> Can you file a follow-up issue to track improved integration with Subiquity states?

Thanks, I filed #1057

PR #1053, opened 2022-08-11
Sort translated keyboard layout names

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2022-08-11 16-34-49 | Screenshot from 2022-08-11 16-33-48 |


CarlosNihelton (approved):
LGTM 👍

PR #1058, opened 2022-08-12
Post-install scripts

jpnurmi:
- Network: `/etc/NetworkManager/system-connections/*`
- Bluetooth: `/var/lib/bluetooth`
- Desktop (dconf) settings for the `ubuntu` user:
  - `/org/gnome/desktop/a11y`
  - `/org/gnome/desktop/interface`
  - `/org/gnome/desktop/peripherals`
  - `/org/gnome/desktop/wm`

### DIFF
--- a/snap/local/postinst.d/10_copy_bluetooth_config
+++ b/snap/local/postinst.d/10_copy_bluetooth_config
@@ -1,0 +1,11 @@
+#!/bin/sh
+set -e
+
+target=/target
+source_bluetooth=/var/lib/bluetooth
+target_bluetooth=$target/$source_bluetooth
+
+if [ -d $source_bluetooth ]; then
+    rm -rf $target_bluetooth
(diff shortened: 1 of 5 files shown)


jpnurmi:
@seb128 I bundled post-install scripts in the snap for easy testing, but would it be better to have them on the image so that flavors can easily extend?

seb128 (approved):
Thanks, that sound reasonable. I didn't try to think about the details of which gsettings keys we need exactly but it is a start point and we can iterate from there. 

It would make sense to move the script to the ISO but the snap makes it easier to iterate and we don't expect any flavor to start using the new installer this cycle so let's include it in the snap for now and once it's confirmed to do what we want we can move it next cycle

jpnurmi:
Thanks! I tried to make the settings export generous and dump any explicit set a11y, keyboard, and theme-related values but I'm sure there's more.

PR #1069, opened 2022-08-17
Installation slides: show Subiquity status & revise journal logs 

jpnurmi:
Instead of showing "Copying files..." from the beginning to the end, show a more descriptive status based on subiquity's syslog events:

- "Copying files..."
- "Installing system..."
- "Configuring system..."

For logs, use the provided syslog ID instead of showing anything from the snap which would include e.g. previous runs if the installer had been restarted.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/services/journal_service.dart
+++ b/packages/ubuntu_desktop_installer/lib/services/journal_service.dart
@@ -2,45 +2,47 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:async/async.dart';
+
+enum JournalOutput {
+  /// The default and generates an output that is mostly identical to the
+  /// formatting of classic syslog files, showing one line per journal entry.
+  short,
+
+  /// A very terse output, only showing the actual message of each journal entry
+  /// with no metadata, not even a timestamp
+  cat,
+}
+
 /// Provides `journalctl` output.
 class JournalService {
-  /// Creates the service. Optionally, the output can be limited to a specific
-  /// systemd unit.
-  JournalService(this._unit);
-
-  Process? _process;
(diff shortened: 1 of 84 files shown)


jpnurmi:
https://user-images.githubusercontent.com/140617/185122580-24c9de87-8cab-4184-8379-d6794c230159.mp4

jpnurmi:
Screenshot from 2022-08-17 15-18-32

packages/ubuntu_desktop_installer/lib/services/journal_service.dart
```diff
@@ -2,45 +2,47 @@ import 'dart:async';
...
-  }
-
-  /// Kills the `journalctl` process.
-  void stop() {
-    _controller.close();
-    _process?.kill();
-    _process = null;
+    try {
+      yield* StreamGroup.mergeBroadcast([
+        process.stdout
```
CarlosNihelton, line 37: I had a suggestion about considering [systemd-journal-gatewayd](https://www.freedesktop.org/software/systemd/man/systemd-journal-gatewayd.service.html), but Dart & Flutter Async API is so ellegant that continously parsing a process console outputs shines.

CarlosNihelton (approved):
LGTM :+1:

Cannot think of anything better to escape from the coverage complains.

packages/ubuntu_desktop_installer/lib/services/journal_service.dart
jpnurmi: Are we going to need something like that on Windows?

CarlosNihelton: Not in my plans. Reading from stdin should be enough for our OOBE.

PR #1070, opened 2022-08-17
Partition edit: remove incorrect fallback format

jpnurmi:
Don't claim that an unformatted existing partition has ext4 which is the default for new partitions.


jpnurmi:
Given an unformatted partition, the edit dialog would confusingly have ext4 selected by default.

Screenshot from 2022-08-17 17-58-55

| Before | After |
|---|---|
| Screenshot from 2022-08-17 18-00-29 | Screenshot from 2022-08-17 17-59-03 |

CarlosNihelton (approved):
Makes sense to avoid users thinking that it was already formatted with ext4.

PR #1076, opened 2022-08-18
Per entry point language blocklist

CarlosNihelton:
In #829 we added a blocklist for languages we needed to avoid translating the UI into due lack of fonts in the root filesystem.
By moving the OOBE into Windows that restriction is now dependent on the entry point.

So, this PR moves out of the `SelectLanguageModel` the code that enforces that restriction into a service class that can be instantiated with different restrictions from `main()`.

Shall other restrictions related to language support arise, that class would be extended to accomodate those.


CarlosNihelton:
Both entry points running simultaneously to demonstrate the resulting behavior: 

image

jpnurmi (approved):
LGTM :+1:

PR #1082, opened 2022-08-23
Fix keyboard layout selection

jpnurmi:
Make use of Subiquity's new input-source API (https://github.com/canonical/subiquity/pull/1388) to sync the system's input-source setting with the selection in the installer GUI.

image


CarlosNihelton (approved):
Great! 👍

PR #1085, opened 2022-08-26
Replace RadioIconTile with simple padding

jpnurmi:
The only reason for using `RadioIconTile` on this page was to get an indentation that matches with `RadioListTile` and `RadioButton`. Given that `RadioIconTile` does not play nicely with `AnimatedExpanded` and `Flexible`, the easiest solution is to define the indentation that matches with `kRadioSize` + 8px spacing so that it can be easily used with a plain simple `Padding` widget.


jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2022-08-26 11-43-54 | Screenshot from 2022-08-26 11-43-37 |

CarlosNihelton (approved):
Simple and effective! LGTM 👍

PR #1086, opened 2022-08-26
[WSL] Splash screen for the reconfiguration workflow running on Windows

### LINKED ISSUE #1066: [WSL] Splash screen for the reconfiguration workflow
One issue that appears when we move the WSL OOBE to Windows is that the window appears much earlier than the server can respond to it in the reconfiguration workflow. That can lead to some seconds without any feedback, just a black screen.

I propose to display a short animation (6 seconds long) while waiting for the server to be ready.

## Current:

https://user-images.githubusercontent.com/11138291/185009593-9cf5f94b-ec61-4bc6-8a17-dabf613b32f5.mp4

## Proposed:

https://user-images.githubusercontent.com/11138291/185009613-afa79103-1c29-45ea-ad1c-fd6a42f43a5e.mp4

The animation choosen is a cut of a longer animation intended to welcome the users during the first install. The fade transition between the animation and the configuration screen was set to 1s in the clip, for reference.

CarlosNihelton:
As reported in #1066 the Windows entry point is able to show the window seconds earlier than the server can start running.

This PR fills that gap with a nice animation provided by the design team. The window size is fixed to avoid the risk of distorting the animation. If the animation finishes before the server is responsive the last frame stays put until the server responds, triggering a fade transition to the reconfiguration wizard.

See below how it looks like. It took longer than usual to start the server when recording this, I guess due processing load on my computer at that time (compiling a program, recording screen, launching a WSL VM etc). It usually stays under 10 s even on cold boot.

https://user-images.githubusercontent.com/11138291/186886305-6dd405f2-5911-466a-8d59-cb361b6b9c24.mp4

This closes #1066


jpnurmi:
Looks good! Did you debug the black flash? Would pre-caching the GIF help?

CarlosNihelton:
> Looks good! Did you debug the black flash? Would pre-caching the GIF help?

Not really, but making the top-level widget child of a `ColoredBox` with color property set to `Colors.white` does the magic, although it seems cheating 🙈

packages/ubuntu_wsl_setup/lib/app.dart
```diff
@@ -13,36 +14,46 @@ class UbuntuWslSetupApp extends StatelessWidget {
...
-          theme: yaruLight,
-          darkTheme: yaruDark,
-          debugShowCheckedModeBanner: false,
-          localizationsDelegates: localizationsDelegates,
-          supportedLocales: supportedLocales,
-          home: buildWizard(context),
-        );
-      }),
+    return ColoredBox(
+      color: Colors.white,
```
jpnurmi, line 27: should it respect the theme? perhaps this could be moved inside SplashScreen and use `Theme.of(context).backgroundColor`?

jpnurmi: oh wait, but we don't have a dark variant of the gif :rofl:

jpnurmi: i think i'd have this in SplashScreen regardless because it's specifically for the splash screen. does it cause any issues that way?

CarlosNihelton: Being on the splash screen doesn't change anything 😅

CarlosNihelton: Maybe I should make the splash the top level one. I avoided that because it's a conditional thing.

CarlosNihelton: Ha! Figured it out! I only need the AnimatedSwitcher to be child of the ColoredBox, that's enough. Let me fix it.

CarlosNihelton: This is the result. 

https://user-images.githubusercontent.com/11138291/186925883-93ae0bb5-c354-4ec7-9d49-1aa287887739.mp4

jpnurmi (approved):
LGTM :+1:

packages/ubuntu_wsl_setup/lib/app.dart
jpnurmi: Smooth!

CarlosNihelton: Thank you for pointing that out.

PR #1087, opened 2022-08-26
UDI: mark network configured after it actually has been configured

jpnurmi:
Subiquity does a connectivity check when network is marked as configured. The result of the connectivity check affects e.g. how `/etc/apt/sources.list` is setup. Therefore, it's important to mark network configured when pressing continue on the "Connect to internet" page instead of on startup.

- https://github.com/canonical/subiquity/blob/main/subiquity/server/controllers/network.py#L284
- https://github.com/canonical/subiquity/blob/main/subiquity/server/apt.py#L210


CarlosNihelton (approved):
LGTM 👍

PR #1089, opened 2022-08-26
UDI: restore the "Try Ubuntu" option

jpnurmi:
It simply closes the window which now does the right thing (restores the dock etc.) in the live session. We don't have repair functionality available at this point, so the "Repair Ubuntu" option was commented out. Also, there's no such route as "/tryubuntu" so it was cleaned up. :)

Screenshot from 2022-08-26 17-11-22


CarlosNihelton (approved):
LGTM

PR #1092, opened 2022-08-29
[WSL] remove mark_configured(filesystem)

jpnurmi:
The WSL OOBE does not use the filesystem endpoint. The call was added in #644 but is no longer necessary.


CarlosNihelton (approved):
I tested it in a live run and I'm confident it works.

PR #1093, opened 2022-08-29
Partition edit dialog: fix the "format" checkbox 

jpnurmi:
Whether the "Format" checkbox is enabled should not be bound to `Partition.canWipe` which is the original partition object without modifications. Bind it to `PartitionFormat.canWipe` instead and use the currently selected value from the edit dialog.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/storage_types.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/storage_types.dart
@@ -9,7 +9,7 @@ extension DiskExtension on Disk {
 }
 
 extension PartitionExtension on Partition {
-  bool get canWipe => mount != null;
+  bool get canWipe => PartitionFormat.fromPartition(this)?.canWipe == true;
   bool get isWiped => wipe == 'superblock';
   String get prettySize => filesize(size ?? 0);
 }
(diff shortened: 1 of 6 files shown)


packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/storage_types.dart
```diff
@@ -9,7 +9,67 @@ extension DiskExtension on Disk {
...
+
+/// Available partition formats.
+class PartitionFormat {
+  const PartitionFormat._(this.type);
+
+  /// The type of the partition format (e.g. 'ext4').
+  final String type;
+
+  /// Whether a partition with this format can be wiped.
+  bool get canWipe => type != 'swap';
```
CarlosNihelton, line 25: Could this ever be inquired about a partition containing an existing OS the user won't want to wipe out?

CarlosNihelton (approved):
Overall, it looks good. Just one minor question, but we are good to go!

packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/storage_types.dart
jpnurmi: https://github.com/canonical/ubuntu-desktop-installer/issues/1101

PR #1094, opened 2022-08-29
subiquity_client: wait until subiquity is really ready

jpnurmi:
A `/meta/status` response alone doesn't mean that subiquity is ready. Wait until subiquity is ready according to the returned status. This fixes random startup issues such as [LP#1988042](https://bugs.launchpad.net/subiquity/+bug/1988042).

https://github.com/canonical/subiquity/blob/1bf562c85a66d13ea2be229829d6b36cd0d1763d/subiquity/tests/api/test_api.py#L104-L117

### DIFF
--- a/packages/subiquity_client/lib/src/server.dart
+++ b/packages/subiquity_client/lib/src/server.dart
@@ -1,9 +1,11 @@
+import 'dart:convert';
 import 'dart:io';
 
 import 'endpoint.dart';
 import 'server/common.dart';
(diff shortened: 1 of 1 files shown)


CarlosNihelton (approved):
LGTM 👍

mwhudson:
Inching closer to thinking about how autoinstall will work!! We should talk about that in Prague if not before, I guess.

PR #1095, opened 2022-08-29
[WSL] Adds an opt-out option for installing language packs

### LINKED ISSUE #1073: [WSL] Option for deferring language pack installation
A considerable amount of time of the WSL setup workflow in spent running apt to download and install language packs and fonts to ensure a good language support. The amount of time depends on which language is selected.

To make things worse, WSL 2 is known to have some internet speed limitation issue, such as what's described in https://github.com/microsoft/WSL/issues/4901 .

In line with the suggestion in https://github.com/ubuntu/WSL/issues/174 I plan to let the user opt-out for that language pack installation feature, trading setup speed for better language support. In such case Subiquity will only mark those packages for later installation, requiring from the user to run `sudo apt-get dselect-upgrade` at some point, whenever she feels like it's more appropriate.

To reflect that possibility in the UI, I propose changes to the `Select your language` page:

## Current

image

## Proposal

image 
image

Notice that the checkbox subtitle changes from a disclaimer about the setup duration to an information about the command required to install the packages marked for later.

CarlosNihelton:
The motivation is outlined in #1073.

This closes #1073 by:

- Updating Subiquity to 68b92f10229560cc34f780500da5a6f8955909f5 to explore the new endpoint;
- Adding a new type for the new entry point;
- Adding a check box below the list of languages in the "Select your language" page and
- Updating the relevant tests.

Since the new endpoint is optional even for the setup workflow, it's still possible to start the wizard at a different page, as demonstrated in the second integration test case.
Also this was made somewhat flexible, so in the future it will be easy to move this flag to a new page if the need for more options with related semantics raise.

### DIFF
--- a/packages/subiquity_client/lib/src/client.dart
+++ b/packages/subiquity_client/lib/src/client.dart
@@ -448,6 +448,22 @@ class SubiquityClient {
     } on HttpException catch (_) {}
   }
 
+  Future<WSLSetupOptions> wslSetupOptions() async {
+    final request = await _openUrl('GET', url('wslsetupoptions'));
+    final response = await request.close();
+
+    final json = await _receiveJson('wslsetupoptions()', response);
+    return WSLSetupOptions.fromJson(json);
+  }
+
+  Future<void> setWslSetupOptions(WSLSetupOptions options) async {
+    final request = await _openUrl('POST', url('wslsetupoptions'));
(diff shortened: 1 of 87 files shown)


jpnurmi:
Sorry for the long delay. :) The change looks good! Do you think we could first upgrade Subiquity separately, though? It probably brings in several unrelated changes and fixes.

PR #1121, opened 2022-09-13
[WSL] Consistent page transitions across platforms

CarlosNihelton:
Yaru is made for Linux only (see [yaru/common_themes.dart](https://github.com/ubuntu/yaru.dart/blob/502783632b853aca1d774c17ee472f1ff7050777/lib/src/themes/common_themes.dart#L244-L248) ), thus running the Windows entry point applies Flutter default page transitions.

This ensures consistency between both platforms by using the first builder (the only one Yaru offers, actually. It happens to be `CupertinoPageTransitionsBuilder` now but could be something else in the future).

We want the look and feel to be the closest possible to the Desktop. Even when running on Windows. It's still an Ubuntu installer.

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/app.dart
+++ b/packages/ubuntu_wsl_setup/lib/app.dart
@@ -32,8 +32,8 @@ class UbuntuWslSetupApp extends StatelessWidget {
             setWindowTitle(lang.windowTitle);
             return lang.appTitle;
           },
-          theme: yaruLight,
-          darkTheme: yaruDark,
+          theme: yaruLight.withConsistentPageTransitions(),
+          darkTheme: yaruDark.withConsistentPageTransitions(),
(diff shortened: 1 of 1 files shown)


jpnurmi:
We could propose a yaru.dart patch that adds consistent page transitions for all desktop platforms?

CarlosNihelton:
> We could propose a yaru.dart patch that adds consistent page transitions for all desktop platforms?

We definitely can, but is this a feature pertinent to Yaru?

jpnurmi:
At least we've made sure the theme runs on any platform. :)

CarlosNihelton:
> At least we've made sure the theme runs on any platform. :)

There we go! https://github.com/ubuntu/yaru.dart/pull/203

CarlosNihelton:
This is not necessary since Yaru 0.4.1.

PR #1124, opened 2022-09-16
CI: unpin Flutter version.

CarlosNihelton:
[Flutter 3.3.2](https://groups.google.com/g/flutter-announce/c/YGf90mszrAw/m/7q1a3UmrBgAJ) reportedly fixes the raster cache rendering issues that motivated pinning the version in the CI workflows.

But since https://github.com/flutter/flutter/commit/bbdf617034171ab1128a594fb24e1c72a09e072e `Uint8List` is exported by `flutter/services.dart` library, thus one small patch in one test file was necessary to avoid linter complains.

Yet, recent changes in [`code_builder`](https://github.com/dart-lang/code_builder/pull/376) caused `mockito` to add trailing commas all around the generated mocks. I considered finding a way to disable it, but it appears to improve the code generation run time, so I opted in for pushing the regenerated files. The same stands for l10n code generation.

Unfortunately I didn't see a way to break those changes in smaller pieces because they are motivated by the Flutter version upgrades.

### DIFF
--- a/.github/workflows/flutter-build-win.yml
+++ b/.github/workflows/flutter-build-win.yml
@@ -31,7 +31,6 @@ jobs:
     - uses: subosito/flutter-action@v2
       with:
         channel: 'stable'
-        flutter-version: '3.0.x'
 
     - name: Build Windows Target
       working-directory: packages\ubuntu_wsl_setup\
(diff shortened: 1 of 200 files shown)


jpnurmi (approved):
Thanks! LGTM :+1:

PR #1128, opened 2022-09-20
Use new simplified API in yaru.dart 0.4

jpnurmi:
The YaruThemeData object passed to builder has gained new theme and darkTheme getters that automatically fallback to the appropriate light and dark themes if a variant could not be detected from the system.


d-loose (approved):
👍

PR #1129, opened 2022-09-20
Allow resize in manual partitioning

jpnurmi:
image


jpnurmi:
There's a great opportunity to refactor and merge the two separate "Add partition" and "Edit partition" dialogs into one because they are now becoming so similar to each other but I'd leave that to a separate refactoring step.

CarlosNihelton (approved):
LGTM

PR #1130, opened 2022-09-21
DiskStorageService: rename hasEncryption to useEncryption

jpnurmi:
This change was extracted from #951.

The "hasEncryption" property was somewhat confusingly named because it's supposed to describe whether the installation should use encryption, not whether the system has encryption.


CarlosNihelton (approved):
Agreed! 👍

PR #1132, opened 2022-09-22
Clear the security key after installation begins

jpnurmi:
At that point, the GUI no longer needs the security key. This does not erase the key from memory but allows the area to be reallocated for something else.

See https://github.com/canonical/ubuntu-desktop-installer/pull/1115#discussion_r976446414 for related discussion.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/services/disk_storage_service.dart
+++ b/packages/ubuntu_desktop_installer/lib/services/disk_storage_service.dart
@@ -71,8 +71,10 @@ class DiskStorageService {
   /// A security key for full disk encryption.
   String? get securityKey => _securityKey;
   set securityKey(String? securityKey) {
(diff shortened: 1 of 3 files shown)


packages/ubuntu_desktop_installer/lib/services/disk_storage_service.dart
```diff
@@ -71,8 +71,10 @@ class DiskStorageService {
   /// A security key for full disk encryption.
   String? get securityKey => _securityKey;
   set securityKey(String? securityKey) {
-    final hiddenKey = securityKey == null ? null : '*' * securityKey.length;
-    log.debug('set security key: $hiddenKey');
+    if (securityKey != null) {
```
jpnurmi, line 74: "set security key: null" felt confusing :)

CarlosNihelton (approved):
LGTM

Issue #697, opened 2022-02-22
Wrong window size


jpnurmi:
The window has the wrong size after switching to [HdyWindow](https://github.com/canonical/ubuntu-desktop-installer/pull/618).

HdyWindow doesn't use GtkWindow.titlebar. The requested window size should take the header bar into account so that 960x680 is left for the content.

| Before | After |
|---|---|
| ![image](https://user-images.githubusercontent.com/140617/155127204-d5b1e65d-a2fd-4ffc-bcb2-ba21ea304665.png) | ![image](https://user-images.githubusercontent.com/140617/155127234-b57cb24d-7065-40fc-bf80-2fc8e7dcc3e0.png) |

bittin:
The Text placement is sometime cut off (Live system or install screen), or behind pictures and such for the Slideshow, would be nice if this can be fixed before the final release

PR #1135, opened 2022-09-23
Fix window size

jpnurmi:
Ever since we switched to [`HdyWindow`](https://github.com/ubuntu-flutter-community/handy_window), the window title bar has been inside the client/content area. We need 960x680 for the actual content i.e. the Flutter view, so add the height of `GtkHeaderBar` to the requested size.

| Before | After |
|---|---|
| image | image |

Close: #697


CarlosNihelton (approved):
LGTM 👍

PR #1144, opened 2022-09-29
[WSL] Fix error reporting on NamedEvent native code

CarlosNihelton:
The OOBE uses an abstraction created on top of Win32 named events to receive notifications from the distro launcher of certain milestones achieved during the setup workflow. It presumes that the event object will have been previously created by the distro launcher, thus an exception is thrown if the call to `OpenEvent()` fails. Some gibberish is being transported by `std::invalid_argument` in such cases, causing the Dart UTF-8 parser to crash.

Using a literal instead of a variable from the scope where the exception is thrown prevents that error.

Besides, the call to `result->Success()` was misplaced, allowing it to happen on exceptional cases as well.


jpnurmi (approved):
LGTM :+1:

PR #1145, opened 2022-09-29
[WSL] Ensures named events are not created in reconfig

CarlosNihelton:
They only make sense in the setup workflow.


jpnurmi (approved):
LGTM :+1:

PR #1147, opened 2022-09-30
snapcraft.yaml: clean up flutter commands

jpnurmi:
flutter-git clones the latest stable branch/channel straight from git so there's no need to switch channels and upgrade. Also, running flutter doctor once after installing flutter is enough.


CarlosNihelton (approved):
LGTM

Issue #1146, opened 2022-09-30
Snap build fails on Launchpad


jpnurmi:
[amd64 builds](https://launchpad.net/~ubuntu-desktop/ubuntu-desktop-installer/+snap/ubuntu-desktop-installer) have been failing since 2022-09-15.

Timeline:
- Flutter 3.3.2 + Dart 2.18.1 were released 2022-09-14: https://docs.flutter.dev/development/tools/sdk/releases
- Builds started failing 2022-09-15: https://launchpad.net/~ubuntu-desktop/ubuntu-desktop-installer/+snap/ubuntu-desktop-installer
- Flutter 3.0.5 was unpinned for Snap builds 2022-09-16: https://github.com/canonical/ubuntu-desktop-installer/pull/1126

Steps to reproduce:
- comment out `source-type: git` in `snap/snapcraft.yaml` to use local source
- clean up `git clean -xdf && git submodule foreach git clean -xdf`
- run `snapcraft remote-build`

Details:

<details><summary>flutter pub get</summary>

```
+ flutter pub get
   Woah! You appear to be trying to run flutter as root.
   We strongly recommend running the flutter tool without superuser privileges.
  /
📎
Running "flutter pub get" in ubuntu_desktop_installer...        [29/Sep/2022:08:30:11 +0000] "CONNECT github.com:443 HTTP/1.1" 200 5683866 "-" "git/2.25.1"
[29/Sep/2022:08:30:21 +0000] "CONNECT pub.dartlang.org:443 HTTP/1.1" 200 10660 "-" "Dart/2.18 (dart:io)"
[29/Sep/2022:08:30:21 +0000] "CONNECT pub.dartlang.org:443 HTTP/1.1" 200 15252 "-" "Dart/2.18 (dart:io)"
[29/Sep/2022:08:30:21 +0000] "CONNECT pub.dartlang.org:443 HTTP/1.1" 200 14277 "-" "Dart/2.18 (dart:io)"
[29/Sep/2022:08:30:21 +0000] "CONNECT pub.dartlang.org:443 HTTP/1.1" 200 5142 "-" "Dart/2.18 (dart:io)"
[29/Sep/2022:08:30:21 +0000] "CONNECT pub.dartlang.org:443 HTTP/1.1" 200 5924 "-" "Dart/2.18 (dart:io)"
[...]

PR #1150, opened 2022-09-30
Alternative workaround for failing `pub get` on Launchpad

jpnurmi:
Fixes: #1146


seb128 (approved):
Thanks, let's merge that one to unblock the builds while we try to figure out the real issue

PR #1169, opened 2022-10-05
melos: add `upgrade` script for running `pub upgrade` in all packages

jpnurmi:


CarlosNihelton (approved):
Great! Many times I ran this manually and though "I should push this eventually :D" ...

PR #1173, opened 2022-10-05
[WSL] Enabling integration tests as an automation tool

CarlosNihelton:
This reuses most of the existing code for the `ubuntu_wsl_setup` package integration tests to create new scenarios that will run as part of the end-to-end test I demonstrated last Friday. Since those test cases depend on WSL, the files were tagged with `@TestOn('windows')` to avoid disturbing this repository CI workflows.

The aim of those additional test scenarios is drive the GUI until the end of a certain workflow. We can assert specific UI elements, but goldens should not be checked in here, because some effects are only possible after the parent process of the test drive exits.

There is a subtle change in the `expectPage()` function introduced due the desire to test the slideshow, which places more than one widget with the same appbar title, thus we match against `findsWidgets` instead of `findsOneWidget` (662e9d8f8103cc73b0d116097a66c4bbe149d2a3).

```diff
void expectPage(
  WidgetTester tester,
  Type page,
  String Function(AppLocalizations lang) title,
) {
  LangTester.type = page;
  expect(find.byType(page), findsOneWidget);
-  expect(find.widgetWithText(AppBar, title(tester.lang)), findsOneWidget);
+  expect(find.widgetWithText(AppBar, title(tester.lang)), findsWidgets);
}
```

### DIFF
--- a/packages/ubuntu_wsl_setup/integration_test/e2e/reconfigure_test.dart
+++ b/packages/ubuntu_wsl_setup/integration_test/e2e/reconfigure_test.dart
@@ -1,0 +1,36 @@
+@TestOn('windows')
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+import 'package:subiquity_client/subiquity_client.dart';
+import 'package:ubuntu_wsl_setup/main_win.dart' as app;
+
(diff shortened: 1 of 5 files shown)


jpnurmi (approved):
Yay, good stuff! LGTM :+1:

PR #1178, opened 2022-10-10
[WSL] Skip listening to server status change if it's not installing

CarlosNihelton:
During integration testing it's possible that `ApplyingChangesModel` start listening after the server is already DONE. If so, the next state change would be to null (shutdown) which only happens if the GUI ends. But GUI won't exit because it's expecting a status change. Thus, infinite loop.
We've seen this consistently with the end-to-end workflow on Windows with a broken rootfs image, causing the server to exit sooner than anticipated. It's possible that a similar situation is also happening on the Linux integration testing.

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_model.dart
+++ b/packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_model.dart
@@ -23,6 +23,11 @@ class ApplyingChangesModel extends SafeChangeNotifier with SystemShutdown {
   bool _previousState = true;
 
   void init({required VoidCallback onDoneTransition}) {
+    if (null != _monitor.status && !_monitor.status!.state.isInstalling) {
+      _previousState = false;
+      onDoneTransition();
+      return;
+    }
     _sub = _monitor.onStatusChanged.listen((status) {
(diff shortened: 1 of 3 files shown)


packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_model.dart
```diff
@@ -23,6 +23,11 @@ class ApplyingChangesModel extends SafeChangeNotifier with SystemShutdown {
   bool _previousState = true;
 
   void init({required VoidCallback onDoneTransition}) {
+    if (null != _monitor.status && !_monitor.status!.state.isInstalling) {
+      _previousState = false;
+      onDoneTransition();
```
jpnurmi, line 28: perhaps it was a mistake to use a callback for this... it could be cleaner to return a Future and simply act upon when the call returns :)

jpnurmi (approved):
LGTM :+1:

packages/ubuntu_wsl_setup/lib/pages/applying_changes/applying_changes_model.dart
CarlosNihelton: Indeed. I'll put this into my refactoring wish list.

PR #1186, opened 2022-10-17
Skip the CLA check for pull requests opened by Renovate (Automated Dependency Updates).

oSoMoN:

PR #1187, opened 2022-10-17
cla-check.yaml: attempt to fix syntax

jpnurmi:
See https://github.com/canonical/ubuntu-desktop-installer/pull/1186#issuecomment-1280669581

image

https://github.com/canonical/ubuntu-desktop-installer/actions/runs/3264412612/workflow

### DIFF
--- a/.github/workflows/cla-check.yaml
+++ b/.github/workflows/cla-check.yaml
@@ -3,7 +3,7 @@ on: [pull_request_target]
 
 jobs:
   cla-check:
-    if: !contains(fromJson('["weblate", "renovate[bot]"]'), github.event.pull_request.user.login)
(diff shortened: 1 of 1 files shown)


oSoMoN (approved):
LGTM. It would be useful if we could do dry runs of workflows from any branch to avoid this sort of situation.

jpnurmi:
Hmm aha, `pull_request_target` used by the CLA check is special. It makes sense now why #1186 was all green. :)

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the `pull_request` event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.

Issue #1189, opened 2022-10-17
Locations popdown is cut at the bottom edge of the dialog


seb128:
On the 'Where are you' page, if you type a location name there is a popdown list displaying, it's cut at the bottom of the UI though which has basically only space for 1.5 entry. Is that a flutter issue? If so could we maybe make it pop up where there is space?

jpnurmi:
- https://github.com/flutter/flutter/issues/101620#issuecomment-1286576417

PR #1191, opened 2022-10-17
Where are you: swap timezone map & dropdowns

jpnurmi:
Leaving a 180px vertical gap below the dropdowns makes the timezone map look squeezed (the slightly wrong window height (#697) makes it even worse). Therefore, I'd propose swapping the order to let the popups open over the timezone map.

| Before | Design | Proposal |
|---|---|---|
| image | image | image |
| image | image | image |
| image | image | image |

Design: https://people.canonical.com/~platform/desktop/installer/13.%20Where%20Are%20You/#artboard0
Resolves: #1189

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/where_are_you/where_are_you_page.dart
@@ -73,20 +73,9 @@ class WhereAreYouPageState extends State<WhereAreYouPage> {
       title: YaruWindowTitleBar(
         title: Text(lang.whereAreYouPageTitle),
       ),
-      headerPadding: EdgeInsets.zero,
       contentPadding: EdgeInsets.zero,
       content: Column(
         children: <Widget>[
-          Expanded(
-            child: TimezoneMap(
-              offset: controller.selectedLocation?.offset,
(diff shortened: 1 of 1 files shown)


jpnurmi:
@elioqoshi There's an issue in Flutter that dropdowns always open downwards even if there's no space. Would it be a reasonable workaround to change the page layout so that the dropdowns are at the top until Flutter might fix them to open upwards when there's no space below?

jpnurmi:
Flutter 3.7 introduced a new M3 DropdownMenu that seemed to open upwards when at the bottom of the screen. We could give it a try for this.

elioqoshi:
@jpnurmi Sorry I missed this before. Let me know if I should review it, the original proposal makes sense and I think the dropdown should be a primary way of selecting your timezone anyway

jpnurmi:
@elioqoshi The looks have evolved a little since the proposal was originally created but here's how it would look like today. Does it still look ok to you?

image
image
image

elioqoshi:
This is fully accessible via keyboard, right? Then it looks good to me!

jpnurmi:
Great, thanks! Yes, these are fully accessible via keyboard :+1:

d-loose (approved):
LGTM :+1:

PR #1192, opened 2022-10-17
[WSL] Skip listening to server status change if it's not installing - PART II

CarlosNihelton:
Since #1178 it is now possible that the page causes the router to call setState during build. The solution was already present on tests, but ideally that should have been moved into the page itself as part of that PR.
This completes that job.


jpnurmi (approved):
LGTM :+1:

PR #1194, opened 2022-10-18
WizardPage: allow specifying content spacing

jpnurmi:
Setting the spacing to 0 allows removing a gap that is currently unconditionally inserted between the content and the footer.

Notice that the attached image is just an illustration of the problem we're facing - not the final design of that page. :)

image


CarlosNihelton (approved):
LGTM 👍

PR #1195, opened 2022-10-18
Desktop installer: add generic service for config values

jpnurmi:
The first and foremost use-case is to write the auto-login user name, which is then read and used in subiquity postinst scripts.

/tmp/ubuntu_desktop_installer.conf:
```
AutoLoginUser=ubuntu
```

P.S. I did consider using the existing `ini` or `ini_file` packages but neither of them allows .ini files without sections. I ended up rolling my own overly simple `key=value` file helper that can be sourced straight into subiquity postinst scripts.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/services/config_service.dart
+++ b/packages/ubuntu_desktop_installer/lib/services/config_service.dart
@@ -1,0 +1,82 @@
+import 'package:file/file.dart';
+import 'package:file/local.dart';
+import 'package:meta/meta.dart';
+import 'package:ubuntu_logger/ubuntu_logger.dart';
+
+/// @internal
+final log = Logger('config');
+
+class ConfigService {
+  ConfigService(this._path, {@visibleForTesting FileSystem? fs})
+      : _fs = fs ?? const LocalFileSystem();
+
+  final String _path;
+  final FileSystem _fs;
+
+  Future<Map<String, String?>> load() async {
+    final file = _fs.file(_path);
+    try {
+      final config =
+          await file.readAsLines().then((lines) => lines.toKeyValueMap());
+      log.debug('loaded ${config.length} entries from $_path');
+      return config;
+    } on FileSystemException catch (e) {
+      if (file.existsSync()) {
+        log.error('failed to load $_path', e);
+      }
+      return {};
+    }
+  }
+
+  Future<void> save(Map<String, String?> config) async {
+    try {
+      final file = _fs.file(_path);
+      final data = config.toKeyValueList().join('\n');
+      await file.create(recursive: true);
+      await file.writeAsString('$data\n');
+      log.debug('saved ${config.length} entries to $_path');
+    } on FileSystemException catch (e) {
+      log.error('failed to save $_path', e);
+    }
+  }
+}
+
+extension _KeyValueList on List<String> {
+  Map<String, String?> toKeyValueMap() {
+    return Map.fromEntries(
+      where((line) => !line.startsWith('#'))
+          .map((line) => line.split('='))
+          .where((parts) => parts.length == 2)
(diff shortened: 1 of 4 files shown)


jpnurmi:
@seb128 do you prefer `/tmp/ubuntu_desktop_installer.conf` over `$XDG_RUNTIME_DIR/ubuntu_desktop_installer.conf`?

seb128:
> @seb128 do you prefer `/tmp/ubuntu_desktop_installer.conf` over `$XDG_RUNTIME_DIR/ubuntu_desktop_installer.conf`?

Using /tmp seems fine to me, it's a temporary filesystem from the installer media and doesn't require to compute the path dynamically

packages/ubuntu_desktop_installer/lib/services/config_service.dart
```diff
@@ -0,0 +1,82 @@
...
+    } on FileSystemException catch (e) {
+      log.error('failed to save $_path', e);
+    }
+  }
+}
+
+extension _KeyValueList on List<String> {
+  Map<String, String?> toKeyValueMap() {
+    return Map.fromEntries(
+      where((line) => !line.startsWith('#'))
```
CarlosNihelton, line 47: I think we need to trim initial whitespaces first to reliably detect comments.

```
# Example usage:          // that's ok
  # AutoLoginUser=ubuntu // that breaks the parser. It will consider `  # AutoLoginUsername` a key.
```

CarlosNihelton: Also, `ini` supports semicolon (`;`) comments. I'm not used to see this being used in Linux config files, though, so I'm not sure whether we should consider it in or out of scope.

packages/ubuntu_desktop_installer/lib/services/config_service.dart
```diff
@@ -0,0 +1,82 @@
...
+          .map((line) => line.split('='))
```
CarlosNihelton, line 48: `ini` additionally allows assignment with a colon (`:`). Is supporting that out of scope?

packages/ubuntu_desktop_installer/lib/services/config_service.dart
```diff
@@ -0,0 +1,82 @@
...
+          .where((parts) => parts.length == 2)
+          .map((parts) => MapEntry(parts.first, parts.last.unquote())),
```
CarlosNihelton, line 50: You may consider also trimming whitespaces from the parts as well.

```ini
AutoLoginUser = ubuntu
```

CarlosNihelton: Also, what happens if we have a trailing inline comment?

```ini
AutoLoginUser = ubuntu # because I'm creative!
```

jpnurmi:
It's an internal file written by the GUI and read by our own Subiquity post-install scripts. It's not worth it trying to build a full-fledged INI parser out of this. :)

packages/ubuntu_desktop_installer/lib/services/config_service.dart
jpnurmi: I've dropped the artificial support for comments. It's just not worth it for an internal machine-generated file. We need to keep it bash-compatible and full INI support is definitely out of scope. :)

packages/ubuntu_desktop_installer/lib/services/config_service.dart
jpnurmi: I don't see any reason to support anything Bash-incompatible. The purpose is to transfer variables from the GUI to our own Bash scripts. :)

CarlosNihelton (approved):
LGTM 👍 

Thanks for the clarification!

PR #1193, opened 2022-10-18
Test postinst scripts

jpnurmi:
Based on the Bach unit testing framework for Bash - https://bach.sh/

PR #1196, opened 2022-10-18
Adapt auto-login setup from Ubiquity

jpnurmi:
A direct copy and paste of https://git.launchpad.net/ubiquity/tree/d-i/source/user-setup/user-setup-apply#n265

Ref: #362

### DIFF
--- a/snap/local/postinst.d/10_configure_auto_login
+++ b/snap/local/postinst.d/10_configure_auto_login
@@ -37,7 +37,7 @@ fi
 
 if $chroot $ROOT [ -f /etc/kde4/kdm/kdmrc ]; then
     # Configure KDM autologin
-    $log $chroot $ROOT sed -i$BACKUP -r \
+    $chroot $ROOT sed -i$BACKUP -r \
         -e "s/^#?AutoLoginEnable=.*\$/AutoLoginEnable=true/" \
         -e "s/^#?AutoLoginUser=.*\$/AutoLoginUser=$USER/" \
         -e "s/^#?AutoReLogin=.*\$/AutoReLogin=true/" \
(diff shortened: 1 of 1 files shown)


jpnurmi:
Some sort of tests would be great... (#1193).

snap/local/postinst.d/10_configure_auto_login
```diff
@@ -0,0 +1,95 @@
...
+        if ! `grep -qs '\[daemon\]' $GDMCustomFile` ; then
+            echo '[daemon]' >> $GDMCustomFile
+        fi
+        sed -i "s/\[daemon\]/\[daemon\]\n$AutologinParameters/" $GDMCustomFile
+    fi
+fi
+
+if $chroot $ROOT [ -f /etc/kde4/kdm/kdmrc ]; then
+    # Configure KDM autologin
+    $log $chroot $ROOT sed -i$BACKUP -r \
```
jpnurmi, line 38: we don't have ubiquity's log-output tool but I left "$log" here regardless. it seems harmless to let it evaluate to empty?

seb128: it isn't a functional issue but it's a bit crufty and at some point in the futur someone might wonder why it's there so I would be in favor of having them removed

Do you want to keep them to make easier to bring updated versions from that file from Debian? If so it seems they should be easier to filter out with a sed, maybe we can just document that at the top of the file?

Also please add a reference to the origin of the file, I don't expect the original to really see changes but it might help to ensure we keep them in sync if ever needed

seb128 (changes requested):
Seems fine to me but let's remove the $log unless there is a reason, see inline comment

seb128 (approved):
Thanks for doing the changes, it is fine like that I think!

jpnurmi:
Ready to go. GUI was merged:
- https://github.com/canonical/ubuntu-desktop-installer/pull/1198

PR #1198, opened 2022-10-19
Restore auto-login GUI

jpnurmi:
The original design had two radio buttons:
- _Log in automatically_
- _Require my password to log in_

The problem was that they wouldn't fit on the screen together with error labels below input fields and the _Show password_ checkbox. @elioqoshi proposed using a single switch toggle instead. :+1: 

| Light | Dark |
|---|---|
| image | image |

Ref: #362


jpnurmi:
The original design had two radio buttons:
- _Log in automatically_
- _Require my password to log in_

The problem was that they wouldn't fit on the screen together with error labels below input fields and the _Show password_ checkbox. @elioqoshi proposed using a single switch toggle instead. :+1:

d-loose (approved):
LGTM :+1:

PR #1201, opened 2022-10-19
Add postinst script for telemetry

jpnurmi:
The script inserts the "done" stage into the telemetry JSON. The duration is calculated based on the creation time of the telemetry report.

The telemetry report is edited in place at `/var/log/installer/telemetry`. Subiquity copies the whole directory to `/target` right after the post-install step.

```json
{
  "Language": "en",
  "Minimal": false,
  "RestrictedAddons": false,
  "PartitionMethod": "use_device",
  "Type": "Flutter",
  "OEM": false,
  "Stages": {
    "0": "/welcome",
    "35": "/tryorinstall",
    "38": "/keyboardlayout",
    "39": "/updateothersoftware",
    "40": "/installationtype",
    "42": "/whereareyou",
    "44": "/whoareyou",
    "48": "/chooseyourlook",
    "49": "/installationslides",
    "80": "/installationcomplete",
    "1234": "done"
  }
}
```
Ref: #1057

Issue #1057, opened 2022-08-12
Telemetry report is written too late


jpnurmi:
The GUI writes `/var/log/installer/telemetry` when the installation is finished and the user presses _Restart_ or _Shutdown_, which is too late for either:
- `subiquity/Late/run` that executes post-install scripts in `/etc/subiquity/postinst.d`, or
- `subiquity/Shutdown/copy_logs_to_target` that copies `/var/log/installer` to `/target`.

What makes this complicated is that the telemetry report is supposed to contain a timestamp for when the installation was finished. At this point, it's too late for the GUI to write anything. Therefore, it may be necessary to have some help from Subiquity.

jpnurmi:
We have a plan. We'll write all steps up until the installation start from the GUI and finish up the report from a Subiquity post-install script.

GUI:
```json
{
  "Language": "en",
  "Minimal": false,
  "RestrictedAddons": false,
  "PartitionMethod": "use_device",
  "Type": "Flutter",
  "OEM": false,
  "Stages": {
    "0": "/welcome",
    "35": "/tryorinstall",
    "38": "/keyboardlayout",
    "39": "/updateothersoftware",
    "40": "/installationtype",
    "42": "/whereareyou",
    "44": "/whoareyou",
    "48": "/chooseyourlook",
    "49": "/installationslides",
    "80": "/installationcomplete"
  }
}
```

post-install script:
```json
{
  "Language": "en",
  "Minimal": false,
  "RestrictedAddons": false,
  "PartitionMethod": "use_device",
  "Type": "Flutter",
  "OEM": false,
  "Stages": {
    "0": "/welcome",
    "35": "/tryorinstall",
    "38": "/keyboardlayout",
    "39": "/updateothersoftware",
    "40": "/installationtype",
    "42": "/whereareyou",
    "44": "/whoareyou",
    "48": "/chooseyourlook",
    "49": "/installationslides",
    "80": "/installationcomplete",
    "123": "done"
  }
}
```

jpnurmi:
Ubiquity's telemetry report for comparison:
```json
{
  "Media": "Ubuntu 22.04 LTS \"Jammy Jellyfish\" - Release amd64 (20220419)",
  "Type": "GTK",
  "OEM": false,
  "PartitionMethod": "use_crypto",
  "DownloadUpdates": true,
  "Language": "en",
  "Minimal": false,
  "RestrictedAddons": true,
  "Stages": {
    "0": "language",
[...]

PR #1202, opened 2022-10-20
TelemetryService: write the report on the disk as we go

jpnurmi:
Instead of waiting until the GUI is closed, write stages and metrics right away on the disk. This ensures that the report is available for the telemetry post-install script (#1201) when Subiquity is just about to finish the installation and copy the telemetry report on the target.

Ref: #1057


CarlosNihelton (approved):
Very good!

Issue #1158, opened 2022-10-05
Pin or publish git dependencies


jpnurmi:
Using our "own" dependencies (e.g. ubuntu-flutter-plugins) straight from git is convenient but we should either start pinning them at certain revisions or preferably publish (*) those that make sense to publish, to avoid build breaks when incompatible changes are pushed.

https://github.com/canonical/ubuntu-flutter-plugins/pull/87
```
ERR : Because every version of timezone_map from git depends on safe_change_notifier ^0.2.0 and ubuntu_desktop_installer depends on safe_change_notifier ^0.1.0, timezone_map from git is forbidden.
```

(*) With published versions, we'd get automatic updates offered by [Renovate](https://github.com/canonical/ubuntu-desktop-installer/pull/1155).

jpnurmi:
The last Git-dependency was pinned in:
- #1822

PR #1209, opened 2022-10-24
Use ubuntu_logger 0.0.1

jpnurmi:
This takes us one step closer to deterministic builds. Now that we have the Renovate bot automatically proposing updates, convenience is no longer an excuse to use the latest main straight from Git. :)

Ref: #1158


jpnurmi:
No code changes. Using the exact same package but hosted instead of git.

PR #1236, opened 2022-11-16
Add contact information

murraybd:
This will allow users to use ubuntu-bug ubuntu-desktop-installer from the "try ubuntu" environment and have bugs created at launchpad which fixes https://bugs.launchpad.net/ubuntu-desktop-installer/+bug/1979845.


kenvandine (approved):
Thanks!

PR #1221, opened 2022-11-11
Update subiquity

jpnurmi:
Generated (99.8%) changes:
- `types.dart` was [generated](https://github.com/canonical/ubuntu-desktop-installer/pull/965) from [`subiquity/common/types.py`](https://github.com/canonical/subiquity/blob/main/subiquity/common/types.py)
- `types.g.dart` and `types.freezed.dart` were generated by `build_runner`
- `snapcraft.yaml` was updated by the [`update_subiquity`](https://github.com/canonical/ubuntu-desktop-installer/blob/main/scripts/update-subiquity) script

Manual (0.2%) changes:
- added the new [`use_tpm`](https://github.com/canonical/subiquity/blob/main/subiquity/common/types.py#L323) field to guided storage choice unit tests
- marked the new [`codecs` endpoint](https://github.com/canonical/subiquity/pull/1473) configured until it gets implemented

PR #1238, opened 2022-11-18
Installation type: prevent pressing "Continue" too early

jpnurmi:
Querying guided storage started taking slightly longer after the latest [Subiquity update](https://github.com/canonical/ubuntu-desktop-installer/pull/1221). There's now a chance that pressing _Continue_ too early takes the user incorrectly to the _Choose guided storage_ page which lists no disks at all until the guided storage query completes. Prevent this by disabling the _Continue_ button until the guided storage information has been received.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/installation_type/installation_type_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/installation_type/installation_type_model.dart
@@ -74,6 +74,9 @@ class InstallationTypeModel extends SafeChangeNotifier {
   /// A list of existing OS installations or null if not detected.
   List<OsProber>? get existingOS => _diskService.existingOS;
 
+  /// Whether storage information has been queried and installation can proceed.
+  bool get hasStorage => _storages != null;
+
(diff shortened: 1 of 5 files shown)


jpnurmi:
## Before

[Screencast from 2022-11-18 09-59-54.webm](https://user-images.githubusercontent.com/140617/202662655-c984822f-9c07-4f3a-a0da-d01b692ea7ba.webm)

## After

[Screencast from 2022-11-18 09-55-32.webm](https://user-images.githubusercontent.com/140617/202662296-7932126a-b30b-4cf1-8f30-11b60a51489b.webm)

mwhudson:
Clearly this is an improvement and I'm not going to try to insist you change anything but in general the fact that you're addressing this by changing the preceding screen is a bit counter to how subiquity expects the client to operate, because in a partially interactive autoinstall scenario, the installation type screen might not even be shown. Instead, the TUI's storage screen shows a "probing for block device" message in this scenario.

CarlosNihelton (approved):
LGTM! 👍

PR #1237, opened 2022-11-17
Updates and other software: codecs

jpnurmi:
This PR is a continuation of #733 and adds a similar option for codecs.

image

Design: https://people.canonical.com/~platform/desktop/installer/6.%20Updates%20and%20Other%20Software/

PR #1239, opened 2022-11-18
Don't mark 'drivers' and 'codecs' configured on startup

jpnurmi:
Drivers & codecs have been implemented in #733 and #1237, respectively, and both endpoints are now configured from the _Updates and other software_ screen.


d-loose (approved):
:+1:

PR #1250, opened 2022-12-05
Downgrade freezed

jpnurmi:
We have issues with version [2.3.0](https://pub.dev/packages/freezed/changelog#230) that added support for covariant unions:
```
lib/src/types.freezed.dart:4142:18: Error: The parameter 'offset' of the method '_$$GapCopyWith.call' has type 'int', which does not match the corresponding type, 'int?', in the overridden method, '$PartitionOrGapCopyWith.call'.
```


jpnurmi:
The issue with WSL `named_event.h` will be fixed by #1251.

CarlosNihelton (approved):
LGTM :+1:

Issue #1243, opened 2022-11-23
Shell theme


jpnurmi:
I selected the dark theme during the installation. Apps are dark, the shell is not.

![image](https://user-images.githubusercontent.com/140617/203551694-299e19a4-f2bb-4010-83f3-97292a20d56e.png)

jpnurmi:
Something overrides gtk-theme:
```sh
$ cat /usr/share/glib-2.0/schemas/20_ubuntu-desktop-installer-interface.gschema.override 
[org.gnome.desktop.interface]
color-scheme='prefer-dark'
gtk-theme='Yaru-dark'
jpnurmi@jpnurmi-Standard-PC-i440FX-PIIX-1996:~$ gsettings get org.gnome.desktop.interface gtk-theme
'Yaru'
jpnurmi@jpnurmi-Standard-PC-i440FX-PIIX-1996:~$ gsettings get org.gnome.desktop.interface color-scheme
'prefer-dark'
```

jpnurmi:
Not sure why the override wasn't picked up.
```
/usr/share/glib-2.0/schemas/20_ubuntu-desktop-installer-interface.gschema.override:gtk-theme='Yaru-dark'
/usr/share/glib-2.0/schemas/10_ubuntu-settings.gschema.override:theme-name = 'Yaru'
/usr/share/glib-2.0/schemas/10_ubuntu-settings.gschema.override:gtk-theme = "Yaru"
/usr/share/glib-2.0/schemas/10_ubuntu-settings.gschema.override:icon-theme = "Yaru"
/usr/share/glib-2.0/schemas/10_ubuntu-settings.gschema.override:cursor-theme = "Yaru"
/usr/share/glib-2.0/schemas/10_ubuntu-settings.gschema.override:gtk-theme = "Yaru"
/usr/share/glib-2.0/schemas/10_ubuntu-settings.gschema.override:icon-theme = "Yaru"
/usr/share/glib-2.0/schemas/10_ubuntu-settings.gschema.override:cursor-theme = "Yaru"
```
> By convention, vendor override files begin with nn_ where nn is a number from 00 to 99. Higher numbered files have higher priority (eg: if the same override is made in a file numbered 10 and then again in a file numbered 20, the override from 20 will take precedence).

jpnurmi:
The one with the ":ubuntu" selector wins even if it's in a file with a lower priority.

```ini
[...]

PR #1255, opened 2022-12-06
postinst: fix desktop schema overrides

jpnurmi:
The ":ubuntu" selector is necessary to override defaults with the same selector.

```ini
# /usr/share/glib-2.0/schemas/10_ubuntu-settings.gschema.override
[org.gnome.desktop.interface:ubuntu]
gtk-theme = "Yaru"
```
vs.

```ini
# /usr/share/glib-2.0/schemas/20_ubuntu-desktop-installer-interface.gschema.override
[org.gnome.desktop.interface] # <== missing ":ubuntu"
gtk-theme='Yaru-dark'
```

Fixes: #1243


seb128 (approved):
Ah indeed, thanks!

PR #1261, opened 2022-12-09
Allows statically linking VC runtime libs

CarlosNihelton:
Back in the time we only planned to support Windows Desktop it was possible to rely on Microsoft Store to deliver a compatible VCLibs package to ensure the required runtime libraries would always be present in the target system.

We cannot rely on that mechanism on Windows Server. Thus we should avoid linking against them dynamically, since CMake 3.15 onwards offers way to do it statically.

The recipe can be applied to any recent Flutter app targeting Windows to avoid the need for shipping msvcp140.dll, vcruntime140.dll and vcruntime140_1.dll libraries. That can be handy for deployment under certain situations, but increases the overall app size since every native plugin (and the runner binary) will be affected by this option (a lot of binary code will be duplicated).

The size increase is irrelevant for the WSL case, because our main size constraint is orders of magnitude bigger than the Flutter-produced binaries.

Notice the usage of generator expressions in the snippet `"MultiThreaded$<$<CONFIG:Debug>:DebugDLL>"` to ensure we still link against the debug runtime DLL's in debug builds, otherwise those builds would be much bigger and likely slower than they currently are.

Relevant documentation about this topic can be found in MS [FAQ about Runtime libs](https://learn.microsoft.com/en-US/troubleshoot/developer/visualstudio/cpp/libraries/faq-standard-cpp-library#what-is-difference-between-crt-and-standard-c---library--what-libraries-will-runtime-library-compiler-options-include), CMake [CMP0091 policy](https://cmake.org/cmake/help/latest/policy/CMP0091.html#:~:text=Runtime%20library%20selection%20typically%20varies%20with%20build%20configuration,edit%20their%20cache%20entries%20to%20adjust%20the%20flags.) and [MSVC_RUNTIME_LIBRARY property](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html#prop_tgt:MSVC_RUNTIME_LIBRARY)

### DIFF
--- a/packages/ubuntu_wsl_setup/windows/CMakeLists.txt
+++ b/packages/ubuntu_wsl_setup/windows/CMakeLists.txt
@@ -1,5 +1,7 @@
 # Project-level configuration.
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.15)
+# Enable statically linking the VC Runtime
+cmake_policy(SET CMP0091 NEW)
 project(ubuntu_wsl_setup LANGUAGES CXX)
 
 # The name of the executable created for the application. Change this to change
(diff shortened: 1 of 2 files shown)


jpnurmi (approved):
LGTM :+1:

PR #1266, opened 2022-12-13
Use YaruBorderContainer in favor of RoundedContainer & RoundedListView

jpnurmi:
`YaruBorderContainer` is very similar to our homemade `RoundedContainer` but with a tiny bit larger border-radius. Let Yaru define the default border-radius for consistency with other Flutter-based Ubuntu apps.

| Before | After |
|---|---|
| Screenshot from 2022-12-13 20-09-28 | image |
| image | image |

NOTE: For reusability and performance reasons, the generic `YaruBorderContainer` doesn't force clipping by default. Any container that can have scrollable content, a press/ripple effect, or anything else that needs to clip along the rounded border should therefore explicitly request clipping.


CarlosNihelton (approved):
LGTM :+1:

PR #1267, opened 2022-12-14
Allows unfinished stdout/err on integration tests.

CarlosNihelton:
When playing with removing screens from WSL setup I observed that it was possible during integration tests to have a new Subiquity process started without the standard streams of the previous one having finished, thus causing the `addStream` calls to throw exceptions.


jpnurmi (approved):
Thanks, LGTM :+1:

PR #1268, opened 2022-12-14
SubiquityMonitor to stop itself on status DONE or EXITED.

CarlosNihelton:
This helps preventing socket exceptions being thrown due server down.
DONE should be enough, but there are cases in which the transition goes too fast and we may only get notified of the EXITED status.


jpnurmi (approved):
LGTM :+1:

PR #1269, opened 2022-12-14
Refactor: AppModel for the WSL app

CarlosNihelton:
To simplify the decision of the default route. That decision was being taken quite scattered between the `main` functions and the `UbuntuWslSetupApp` class. When playing with conditionally skipping the `SelectLanguagePage` it became evident that the complexity of that decision was going too far.
So this refactor implements an POD `AppModel` class to encapsulate that decision, accepting the parameters that affect it. It's intended to be used behind a `ValueListentable<AppModel>` so the app can react to the server response to calls such as `GET /meta/variant' and `GET /locale`.
The following changes should not yet affect any behavior, but allow for simple changes to make that happen.


jpnurmi (approved):
LGTM :+1:

PR #1289, opened 2023-01-05
Use padding & spacing constants from Yaru

jpnurmi:
The padding and spacing values were somewhat inconsistent. The values used were mostly 20px or 24px, sometimes using app-defined constants, and sometimes values hardcoded inline. Use Yaru's 20px padding constant to have consistent padding throughout the installer GUI and with other Flutter-based Ubuntu apps.

| Before (24px) | After (20px) |
|---|---|
| image | image |


d-loose (approved):
LGTM :+1:

PR #1290, opened 2023-01-05
Installation slides: use YaruIconButton

jpnurmi:
The Yaru icon button has a bit nicer highlight and stays circular, unlike the M3 icon button.

| Before | After |
|---|---|
| image | image |
| image | image |


d-loose (approved):
LGTM :+1:

Issue #1245, opened 2022-11-28
Key search in list views


jpnurmi:
_Originally posted by @madsrh in https://github.com/canonical/ubuntu-desktop-installer/issues/343#issuecomment-1328030411_

> hitting a key in the list, doesn't jump to that letter. Eg. if English keyboard is selected, pressing D should jump to Danish

![image](https://user-images.githubusercontent.com/3986894/204086767-b21ca6a3-d105-486b-9749-ec749f390a62.png)

jpnurmi:
Thanks again for the suggestion, @madsrh! Key search has been added to language and keyboard layout lists.

PR #1291, opened 2023-01-05
Welcome page: implement key search for languages

jpnurmi:
[Screencast from 2023-01-05 16-15-33.webm](https://user-images.githubusercontent.com/140617/210814165-5623f407-57fb-4f82-8206-3641c7bff929.webm)

Ref: #1245

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/welcome/welcome_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/welcome/welcome_model.dart
@@ -3,6 +3,7 @@ import 'dart:ui';
 import 'package:safe_change_notifier/safe_change_notifier.dart';
 import 'package:subiquity_client/subiquity_client.dart';
 import 'package:ubuntu_logger/ubuntu_logger.dart';
+import 'package:ubuntu_widgets/ubuntu_widgets.dart' show KeySearchX;
 
 import '../../l10n.dart';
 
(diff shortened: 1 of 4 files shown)


packages/ubuntu_desktop_installer/lib/pages/welcome/welcome_model.dart
```diff
@@ -55,6 +56,22 @@ class WelcomeModel extends SafeChangeNotifier {
   /// Returns the name of the language at the given [index].
   String language(int index) => _languageList[index].name;
 
+  /// Searches for a language matching the given [query]. The search is case-
+  /// insensitive, ignores diacritics, starts from the selected language, and
+  /// wraps around if not found.
+  int searchLanguage(String query) {
```
jpnurmi, line 62: hmm, we're going to need the same wrapping search logic for all lists. this could be shared somewhere

packages/ubuntu_desktop_installer/test/welcome/welcome_model_test.dart
```diff
@@ -79,4 +79,32 @@ void main() {
...
+    // case-insensitive
+    final french = model.searchLanguage('FRA');
+    expect(model.language(french), equals('Français'));
+    model.selectedLanguageIndex = french;
+
+    // wrap around
+    final danish = model.searchLanguage('d');
+    expect(model.language(danish), equals('Dansk'));
+    model.selectedLanguageIndex = danish;
+
```
CarlosNihelton, line 105: Would it be an overkill to add one test case for when the string doesn't match from start and other when it fails to match after a few characters?

CarlosNihelton (approved):
LGTM :+1: 

Great job on this searchability implementation.

packages/ubuntu_desktop_installer/lib/pages/welcome/welcome_model.dart
jpnurmi: Hmm, this would probably fit the collection of "non-visual" ubuntu_widgets better.

PR #1298, opened 2023-01-09
Removes Advanced Setup Route

CarlosNihelton:
Addressing feedbacks we receive that the options exposed via this route are not relevant for most users during setup we put together this PR to remove that, while still preserving that page for the reconfiguration workflow.

Latest subqiuity (since https://github.com/canonical/subiquity/pull/1522 ) is required.

Apparently updating subiquity in a separate PR broke the integration tests (ref.: #1296 ), so I had to put them all together.

### DIFF
--- a/packages/ubuntu_wsl_setup/lib/wizard.dart
+++ b/packages/ubuntu_wsl_setup/lib/wizard.dart
@@ -31,12 +31,8 @@ class UbuntuWslSetupWizard extends StatelessWidget {
         ),
         Routes.profileSetup: WizardRoute(
           builder: ProfileSetupPage.create,
-          onNext: (settings) {
-            if ((settings.arguments as bool?) == true) {
-              return Routes.advancedSetup;
-            }
-            return Routes.applyingChanges;
-          },
+          // Unconditionally skip the AdvancedSetupPage due MS feedback.
+          onNext: (settings) => Routes.applyingChanges,
         ),
         Routes.advancedSetup: const WizardRoute(
           builder: AdvancedSetupPage.create,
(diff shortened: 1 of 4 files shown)


jpnurmi:
It wouldn't hurt to leave the route for development and testing purposes because it's hidden from help anyway?

CarlosNihelton:
> It wouldn't hurt to leave the route for development and testing purposes because it's hidden from help anyway?

That's right as long as we don't refer to that route in the setup part of the wizard.

packages/ubuntu_wsl_setup/lib/wizard.dart
```diff
@@ -29,17 +29,8 @@ class UbuntuWslSetupWizard extends StatelessWidget {
         Routes.selectLanguage: const WizardRoute(
           builder: SelectLanguagePage.create,
         ),
-        Routes.profileSetup: WizardRoute(
+        Routes.profileSetup: const WizardRoute(
           builder: ProfileSetupPage.create,
-          onNext: (settings) {
-            if ((settings.arguments as bool?) == true) {
-              return Routes.advancedSetup;
-            }
```
jpnurmi, line 37: would it be enough to remove this part and let it unconditionally jump to `Routes.applyingChanges`? could we that way keep `Routes.advancedSetup` in the list of known routes?

jpnurmi (commented):
> That's right as long as we don't refer to that route in the setup part of the wizard. But doing so would allow for passing a route that the wizard doesn't have a builder for.
>
> Isn't that worse?

I'm not sure I follow but the route should have a builder or else it will be useless. :) I find it really handy to be able to specify the initial route in general. When working on a specific feature, you can get straight to the respective page without navigating through other pages on startup or after a hot restart.

packages/ubuntu_wsl_setup/lib/wizard.dart
CarlosNihelton: Yeah, that's doable.

CarlosNihelton: Done

jpnurmi (approved):
LGTM :+1:

PR #1299, opened 2023-01-10
Keyboard layout: fix disposal

jpnurmi:
Close streams, cancel subscriptions, and dispose in correct order.


CarlosNihelton (approved):
LGTM 👍🏽 

Funny to see how many reviews the PR which introduced those stream controllers had... :D

jpnurmi:
I was surprised to find ... streams :open_mouth:

Issue #1309, opened 2023-01-13
Widgets in "Advanced features" are not aligned properly


jibel:
ubuntu-desktop-installer   0+git.7b30974e    734

![installer_2023-01-13_09-56](https://user-images.githubusercontent.com/1928546/212282469-fc181a42-8a90-4ea9-9aee-330326f4a9cd.png)

PR #1311, opened 2023-01-13
Fix advanced features dialog alignment

jpnurmi:
Yaru check and radio buttons were [fixed to not stretch](https://github.com/ubuntu/yaru_widgets.dart/pull/415), which screwed up the alignment here because columns are center-aligned by default.

| Before | After |
|---|---|
| Screenshot from 2023-01-13 10-17-48 | Screenshot from 2023-01-13 10-17-37 |

Fixes: #1309

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/routes.dart
+++ b/packages/ubuntu_desktop_installer/lib/routes.dart
@@ -1,5 +1,5 @@
 abstract class Routes {
-  static const initial = installationType;
+  static const initial = welcome;
   static const welcome = '/welcome';
   static const tryOrInstall = '/try-or-install';
   static const turnOffRST = '/turn-off-rst';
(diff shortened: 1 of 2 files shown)


packages/ubuntu_desktop_installer/lib/routes.dart
```diff
@@ -1,5 +1,5 @@
 abstract class Routes {
-  static const initial = welcome;
+  static const initial = installationType;
```
d-loose, line 2: Did you intend to change the initial route in this PR as well?

jpnurmi: Hah, and not the first time because it's so convenient for testing. :D I'm glad there's a test that catches this.

d-loose (approved):
:+1:

PR #1316, opened 2023-01-16
NetworkTile: use YaruTile

jpnurmi:
Replace the home-made tile layout with `YaruTile` that offers a convenient way to achieve the common leading-title-subtitle-trailing "tile" layout.

The `kYaruCheckradio*` constants were considered internal implementation details and moved out of the public API to avoid people using them in creative ways, like myself here...
- https://github.com/ubuntu/yaru_widgets.dart/pull/495

No visual difference:

| Before | After |
|---|---|
| image | image |


d-loose (approved):
LGTM :+1:

Issue #670, opened 2022-02-09
The decoration title is cut when selecting another language


seb128:
Using the current canary daily with rev288, if you select another language on the first screen the win title is cut, for example in french it displays 'Installer Ub...'. That's in a gnome boxes instance but the resolution choice doesn't seem to change the result

jpnurmi:
https://user-images.githubusercontent.com/140617/153219723-4617c132-14c7-42dc-aa81-0af38b4123b1.mp4

jpnurmi:
I'll double check after lunch but I think https://bugs.launchpad.net/ubuntu/+source/gtk+3.0/+bug/1960947 solves this too!

jpnurmi:
Even though this is clearly more like a layout calculation issue, I can't seem to be able to reproduce this after [LP#1960947](https://bugs.launchpad.net/ubuntu/+source/gtk+3.0/+bug/1960947) has landed. It's a bit strange that a miscalculated OpenGL update area would have a side-effect like this, but perhaps it leaves an intermediate layout visible due to a missing redraw.

jpnurmi:
https://user-images.githubusercontent.com/140617/156997390-1b1c495b-a0d4-4171-ad86-71087240efde.mp4

jpnurmi:
The title is elided even shorter when running [integration tests](https://github.com/canonical/ubuntu-desktop-installer/pull/283)

![Screenshot from 2022-03-07 16-42-11](https://user-images.githubusercontent.com/140617/157067373-11a827d2-0c71-4e17-866c-74e3f41bfb49.png)

https://user-images.githubusercontent.com/140617/131746320-9b5c6049-74bd-42bf-99d8-8bb4626cf674.mp4

jpnurmi:
A minimal Flutter app + integration test that reproduces the issue: [flutter_window_title_issue](flutter_window_title_issue). I've added a long list of notes there too.

jpnurmi:
daily-live 20230103
![image](https://user-images.githubusercontent.com/140617/210332947-92096f2b-b26b-4cd6-9c88-2f4a9939b21e.png)

jpnurmi:
We could replace the GTK header bar with a Flutter widget when I'm done with:
- https://github.com/ubuntu/yaru_widgets.dart/pull/455

jpnurmi:
A quick and dirty proof of concept:

[...]

PR #1319, opened 2023-01-17
Move AppBar outside of WizardPage

jpnurmi:
This PR does not introduce visual changes but only prepares `WizardPage` to allow passing any `PreferredSizeWidget` to make it possible to replace it with `YaruWindowTitleBar` in the desktop installer. The WSL setup may choose to continue using `AppBar`.

Ref: #670


d-loose (approved):
LGTM :+1:

PR #1320, opened 2023-01-17
Installation slides: move AppBar outside SlideShow

jpnurmi:
This PR does not introduce visual changes but only prepares `InstallationSlidePage` so that we can easily replace `AppBar` with `YaruWindowTitleBar`.

Previously, each slide was wrapped with a `Scaffold` bundled with an `AppBar` for the slide title. This made slide titles nicely animate together with the slides. This approach won't work with `YaruWindowTitleBar` because we don't want the window controls sliding around. :)

This PR creates a separate `AppBar` that can be easily replaced with a `YaruWindowTitleBar` when the time comes. It contains an embedded `PageView` for the slide titles and keeps it in sync with the slide show. The animations use the same curve and duration so it looks the same as before i.e. as if the app bar was inside the slide show.

[Screencast from 2023-01-17 16-14-15.webm](https://user-images.githubusercontent.com/140617/212938071-d72113c7-e483-43c2-bb18-8317719a5938.webm)

Ref: #670

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_page.dart
@@ -34,6 +34,8 @@ class InstallationSlidesPage extends StatefulWidget {
 }
 
 class _InstallationSlidesPageState extends State<InstallationSlidesPage> {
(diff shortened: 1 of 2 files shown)


d-loose (approved):
LGTM :+1:

PR #1323, opened 2023-01-19
Use YaruDialogTitleBar

jpnurmi:
This not only makes the dialogs look nicer but also adds such behavior that could be considered a pre-requisite for replacing the native `HdyHeaderBar` with `YaruWindowTitleBar` (outside vs. inside the Flutter view, respectively).

When a modal dialog is open, the entire Flutter view is covered by a modal barrier. This means that `YaruWindowTitleBar` is not accessible underneath whenever a modal dialog is open. `YaruDialogTitleBar` solves this by offering the usual top-level window actions (move by drag, right-click menu) via the dialog title bar. It feels a lot like a modal top-level dialog attached to and blocking its parent window.

| Before | After |
|---|---|
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

Ref: #670

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
@@ -21,7 +21,8 @@ Future<bool> showConfirmationDialog(
     context: context,
     builder: (context) {
       return AlertDialog(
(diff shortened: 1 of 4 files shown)


jpnurmi:
The test failure was fixed by https://github.com/ubuntu/yaru_widgets.dart/pull/535 - not yet released.

d-loose (approved):
LGTM :+1:

PR #1324, opened 2023-01-19
Fix "Resize partition" -dialog's button spacing

jpnurmi:
It was spotted in #1323 that the spacing between the ok and cancel buttons is different than in all other dialogs. I wish this was themable in Flutter...

| Before | After |
|---|---|
| image | image |


d-loose (approved):
:+1:

Issue #1321, opened 2023-01-18
Installing: use yaru icons


Feichtmeier:
From the recent screenshots I saw that the material terminal icon is still being used.
![grafik](https://user-images.githubusercontent.com/15329494/213319203-47d5968e-57dd-4578-bcf4-0686a8d7e00c.png)

Would be cool if you could use the yaru icon instead
![grafik](https://user-images.githubusercontent.com/15329494/213319119-ac1f5c2c-4a1c-4dc6-942a-19f83d8e73bf.png)

Edit: also the arrow seems wrong, it should be this one
![grafik](https://user-images.githubusercontent.com/15329494/213319492-033de8b0-b9ae-4eae-9ae7-779f411b6d67.png)

jpnurmi:
Somehow the go_next/previous icons look very thin compared to Ubiquity:
| Flutterstaller | Ubiquity |
|---|---|
| ![image](https://user-images.githubusercontent.com/140617/213477821-2ccb444a-0c1b-4409-97b3-4973f193690f.png) | ![Screenshot from 2023-01-19 15-50-29](https://user-images.githubusercontent.com/140617/213478001-f1ed683b-c690-45d3-87ad-edf993f16a00.png) |

Feichtmeier:
I can't say what's the design specification is for this page but those navigation buttons are normally covered with the pan_xxx buttons on the yaru theme

jpnurmi:
Fixed by:
- https://github.com/canonical/ubuntu-desktop-installer/pull/1598

PR #1326, opened 2023-01-19
Installation slides: use terminal icon from Yaru

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2023-01-19 15-09-51 | Screenshot from 2023-01-19 15-09-42 |

Ref: #1321


bittin:
@jpnurmi the terminal is in the way of the slideshow, does it work to get it smaller so you can both see parts of the terminal output and the slide show?

jpnurmi:
@bittin I like the idea. It could be some kind of draggable bottom sheet. 🙂

Issue #1307, opened 2023-01-13
Try/Install page is displayed again during installation from live session


jibel:
Lunar / ubuntu-desktop-installer 0+git.518d1ff6 712

Test Case:
1. Boot the installer
2. When the page Try/Install is displayed, select "Try" to boot to the live session
3. Start the installer from the live session
4. Select a language and proceed to next page

Actual result
The page Try/Install is displayed again, while it shouldn't (we're in the live session already)

![installer_2023-01-13_09-42](https://user-images.githubusercontent.com/1928546/212278338-fcbb0940-b28b-49cc-8fde-3ae665ebf625.png)

jpnurmi:
@seb128 Any thoughts on how to handle this? The installer is always running in the live session but can we detect the initial startup mode? GSettings?

seb128:
> @seb128 Any thoughts on how to handle this? The installer is always running in the live session but can we detect the initial startup mode? GSettings?

Unsure what automatic logic we could use, but an option is to add a command line parameter --try-or-install and update the systemd unit to include the option (or the other way around, make it default and have a flag to disable the page and update the .desktop to use it)?

jpnurmi:
@seb128 Could you help adding `--try-or-install` for the systemd unit, please? The command-line argument is available in the latest stable.

seb128:
I've added it now, the change should be included in tomorrow's daily image

jpnurmi:
Thanks! I've confirmed the correct behavior with 20230206.

PR #1328, opened 2023-01-20
Add --try-or-install command-line option

jpnurmi:
Skip the _Try or install_ page unless requested by passing `--try-or-install`.

Ref: #1307

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/installer.dart
+++ b/packages/ubuntu_desktop_installer/lib/installer.dart
@@ -49,6 +49,7 @@ Future<void> runInstallerApp(
         defaultsTo: 'examples/simple.json',
         help: 'Path of the machine config (dry-run only)');
     parser.addOption('bootloader', hide: true);
+    parser.addFlag('try-or-install', hide: true);
   })!;
 
   final subiquityClient = SubiquityClient();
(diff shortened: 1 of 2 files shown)


jpnurmi:
I guess we can live with this property drilling for now but I think soon it's time to introduce a top-level app model as you did in the WSL setup. :)

packages/ubuntu_desktop_installer/lib/installer.dart
```diff
@@ -256,8 +264,17 @@ class _UbuntuDesktopInstallerWizard extends StatelessWidget {
     return Wizard(
       initialRoute: initialRoute ?? Routes.initial,
       routes: <String, WizardRoute>{
-        Routes.welcome: const WizardRoute(
+        Routes.welcome: WizardRoute(
           builder: WelcomePage.create,
+          onNext: (_) {
+            if (tryOrInstall == true) {
+              return Routes.tryOrInstall;
```
jpnurmi, line 271: these lines are covered by the integration test even though we don't collect coverage info for them because by running the entire app it'd be too easy to cover 100% :P

CarlosNihelton (approved):
I see something like a global constant app config easily fits in here, but let's keep the idea for the future. 

LGTM 👍🏽

PR #1333, opened 2023-01-24
Move MockSubiquityStatusMonitor together with other Subiquity mocks

jpnurmi:
The ongoing work with window title bars introduces yet another use case for `MockSubiquityStatusMonitor`. Share it via `ubuntu_test/mocks` the same way as `MockSubiquityClient` and `MockSubiquityServer` to avoid generating the same mock again and again.


d-loose (approved):
LGTM :+1:

Issue #1332, opened 2023-01-24
Unsupported format options offered in custom partitioning


xnox:
curtin supports consuming a large variety of filesystems. These are not supported by all Operating Systems.

For example the large variety of FAT and NTFS filesystems is to support installting Windows from MAAS. Installing Ubuntu on NTFS is not supported.

Majority of these are not supported by Ubuntu or Canonical for new installations. And are not offered in neither MAAS or subiquity.

See screenshots below:

![subiquity-options](https://user-images.githubusercontent.com/19779/214284883-37a1ed30-057f-45ac-be16-d5f3ef2c9f54.png)
![instaler-options-2](https://user-images.githubusercontent.com/19779/214284886-297bcc60-7f9f-47e0-aadd-b6e85a167079.png)
![installer-options-1](https://user-images.githubusercontent.com/19779/214284888-4a210eac-90f2-4990-999d-a39963971edd.png)

Please do not allow creating legacy/obsolete filesystems, or those with limited kernel support:
* ext2
* ext3
* ISO 9660 (read-only FS)
* JFS
* ReiserFS

No new installs should contain that.

Please do not offer creating FAT:
* FAT
* FAT12
* FAT16
* FAT32
* VFAT
It must not be used for any mountpoints, apart from EFI partition which is separate.

Please do not offer creating NTFS
* R/W and ACL support is limited, thus it must not be offered as a mountpoint

Unformatted option is not offered
* installer should offer leaving partitions unformatted

The offering of zfsroot option is confusing - it is not offered by subiquity; it is considered experimental in curtin; it doesn't setup things compatible with encryption; zsys; snapshotting; grub integration; or even a supported featureset. Meaning it is easy to bork such installations. Note that subiquity does not offer zfsroot option.

Compared with subiquity installer manual partitioning lacks LVM and RAID options. I don't know if this is intentional, or simply missing feature parity for now.

cc @mwhudson @dbungert @ogayot @vorlonofportland

jpnurmi:
@dbungert `add_partition` throws when trying to create an unformatted partition

```
Traceback (most recent call last):
  File "/home/jpnurmi/Projects/canonical/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/common/api/server.py", line 134, in handler
    result = await implementation(**args)
[...]

PR #1335, opened 2023-01-24
Hide unsupported/legacy/obsolete partition format options

jpnurmi:
When creating new partitions:
- leave out unsupported/legacy/obsolete partition formats,
- offer a "Leave unformatted" option, and
- reorder the options to match Subiquity.

| Before | After |
|---|---|
| image | image |

Close: #1332

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
+++ b/packages/ubuntu_desktop_installer/lib/l10n/app_en.arb
@@ -457,6 +457,8 @@
   "@partitionFormatReiserFS": {},
   "partitionFormatZfsroot": "ZFS root file system",
   "@partitionFormatZfsroot": {},
+  "partitionFormatNone": "Leave unformatted",
+  "@partitionFormatNone": {},
   "partitionErase": "Format the partition",
   "@partitionErase": {},
   "partitionMountPointLabel": "Mount point:",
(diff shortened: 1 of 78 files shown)


xnox (approved):
No idea what dart is, but this looks good to me.

kenvandine:
I see these are alphabetical, but perhaps we should put ext4 at the top of the list as the *preferred*?  @xnox thoughts on that?

dbungert:
Subiquity dialog for reference:

image

ext4 is the default.

xnox:
> I see these are alphabetical, but perhaps we should put ext4 at the top of the list as the _preferred_? @xnox thoughts on that?

i don't have an opinion, i think ext4 is pre-selected as the default option in the drop down already? it doesn't have to be the first one in the drop down. But maybe using the same custom order as subiquity is nice - for the three people who will use both installers side by side (paride, xnox and sil2100).

jpnurmi:
We could eventually replace the Material style dropdown with a popup menu from Yaru widgets that allows inserting a separator:

image

xnox:
Has this been merged?

jpnurmi:
Sorry for the delay. I've added "Leave unformatted" and reordered the options to match Subiquity.

I'm not satisfied with the looks but let's leave it to a separate PR. :)

d-loose (approved):
LGTM :+1:

PR #1329, opened 2023-01-20
Use YaruWindowTitleBar

jpnurmi:
[Screencast from 2023-01-23 08-47-10.webm](https://user-images.githubusercontent.com/140617/213988916-fa096166-6364-43b1-bbb3-013ddccc5718.webm)

Close: #670

PR #1337, opened 2023-01-24
Control window closability from the widget tree

jpnurmi:
Instead of subscribing to Subiquity status changes in runInstallerApp() (and leaking the subscription), do it from within the widget tree to be able to update the upcoming Flutter-based window title bar in #1329.


d-loose (approved):
LGTM :+1:

PR #1345, opened 2023-01-27
WSL: re-enable automatic l10n generation

jpnurmi:
Unlike UDI which is used like a package to build installer flavors and therefore must have l10n checked in, the WSL setup can generate them on the fly just like a normal Flutter app.

There were times when `flutter: generate` didn't play well with IDEs because they weren't able to resolve synthetic packages generated into `.dart_tool` so you'd be stuck with red errors all over the tree no matter what you did. Those Flutter tooling issues have been solved a long ago, though. :)


CarlosNihelton (approved):
LGTM 👍🏽 

In line with the conversation we had about the snapcraft plugin.

PR #1351, opened 2023-01-30
Installation slides: remove log view background & border

jpnurmi:
yaru.dart 0.5.0 added fill color and a default border in different states that we don't want in the log view.

| Before Yaru update | After Yaru update | After this PR |
|---|---|---|
| image | image | image |


d-loose (approved):
LGTM :+1:

Issue #1354, opened 2023-01-31
Flavours: Guidance sought how best to implement per flavour slides


fossfreedom:
Hi,

  Is there any guidance available how per flavour slides are to be submitted?

https://github.com/canonical/ubuntu-desktop-installer/blob/main/packages/ubuntu_desktop_installer/lib/slides/default_slides.dart

Looking at the above the list of slides is hard-coded.   So would an alternative list for each flavour need some-sort of (pseudo code alert) if then else statement?

```
"if budgie flavour do

  this slides

elseif mate flavour do ....

else
  do the default slides
endif
```

I'm presuming there is only one snap - so the snap somehow needs to work out if its running on a specific flavour to provide the customisation? Or is there suppose to be a flavour specific snap installer?

jpnurmi:
The idea is that the theme, slides, etc. can be passed to `runInstallerApp()` but we have a known issue that the current project structure is not good for creating flavors outside of the main repo, unfortunately. The `subiquity_client` package is not a normal Dart package because it needs to be coupled with the Subiquity submodule. We should move it to a separate repo to make it possible to build flavors with the same structure we have in the main installer repo.

jpnurmi:
I'm preparing a dummy example with a custom theme, image assets, and installation slides:

[Screencast from 2023-02-01 17-47-37.webm](https://user-images.githubusercontent.com/140617/216107934-cbd1372f-4b94-4bd0-906f-9e0f28ff8a19.webm)

It will be a separate standalone repo that you can take as a starting point. It uses `ubuntu_desktop_installer` and `subiquity_client` as Git dependencies. I will push it as soon as I have all the necessary changes merged. I have already moved [subiquity_client](https://github.com/canonical/subiquity_client.dart) and the next step is to take it into use here: https://github.com/canonical/ubuntu-desktop-installer/pull/1355. After that, we'll need to tweak the local dependencies a bit and then we should be ready to start trying it out. :+1:

ItzSwirlz:
Cool! When should it be implemented per flavor?

jpnurmi:
There will be an example to get everyone started soon. :slightly_smiling_face: We will support you with any changes that may be needed for `ubuntu_desktop_installer` and `subiquity`. It's quite likely that there are some desktop-specific adjustments necessary here and there.

jpnurmi:
[...]

PR #1355, opened 2023-02-01
Switch to subiquity_client submodule

jpnurmi:
`subiquity_client` is not a fully standalone Dart package as it needs to be coupled with Subiquity, which in turn needs its own dependencies, to be able to function. Therefore, `subiquity_client` cannot be hosted on pub.dev nor used directly as a Git dependency because neither approach supports Git submodules. The only possibility is to use `subiquity_client` as a local dependency from path, where it can have the Subiquity submodule checked out and Subiquity’s own dependencies installed.

The current project structure where the client is placed into a `packages/subiquity_client` subdirectory is problematic for flavors because that way `subiquity_client` cannot be added as a submodule to installer flavors to achieve the same structure as the vanilla installer uses. The correct structure is important for being able to correctly locate the client and set up a suitable environment for dry-running purposes. It also helps to avoid hitting the classic `AF_UNIX` path length limit for the UNIX domain socket used for communicating with Subiquity.

Moving `subiquity_client` into its own repository not only allows flavors to customize the installer without forking the whole project but also makes it easier to manage multiple Subiquity branches in the future as `subiquity_client` can follow Subiquity’s branching model and different installers and flavors can easily choose the exact version they want.

https://github.com/canonical/subiquity_client.dart

Ref: #1354


seb128 (approved):
Alright, those seems convinced reasons for the change!

PR #1370, opened 2023-02-06
Reduce the window size to match the design

jpnurmi:
We should not reserve the extra 48px for a window title bar anymore because we merged the app bar and the window header bar in #1329. Without the extra 48px we still have effectively the same content area as it was before the title bar changes.

Thanks to the app & title bar merger, we ended up having too much empty space on some pages making them look a bit unbalanced.

| Before | After |
|---|---|
| Screenshot from 2023-02-06 14-09-41 | Screenshot from 2023-02-06 14-08-44 |


d-loose (approved):
:+1:

Issue #1372, opened 2023-02-07
Label of partition is incomplete


jibel:
Lunar 20230207

I selected an LVM installation (no encryption) and the label of the 3rd partition is incomplete. Cf screenshot below

![image](https://user-images.githubusercontent.com/1928546/217264580-7cf77be9-c87a-485b-935b-5e36e2baf88d.png)

PR #1381, opened 2023-02-09
Write changes to disk: add missing partition format check

jpnurmi:
Don't assume that a wiped partition has a format because it's not always the case.

| Before | After |
|---|---|
| image | Screenshot from 2023-02-09 10-49-04 |

Fixes: #1372


d-loose (approved):
:+1:

PR #1383, opened 2023-02-09
ValidatedFormField: select all when gaining input focus

jpnurmi:
> For example the hostname is pre-filled but it's unlikely the name you want to give to the machine. So you have to do CTRL+A or select with the mouse or delete the entire text to replace the content

[Screencast from 2023-02-09 12-06-58.webm](https://user-images.githubusercontent.com/140617/217797144-cf9ba165-eab4-4da6-a1bc-e074c9f450b4.webm)


d-loose (approved):
LGTM :+1:

Issue #1385, opened 2023-02-09
Codecs installation fail when offline


seb128:
The ubuntu-restricted-addons packages isn't on the ISO and shouldn't be so we should probably disable the option for offline installations

PR #1387, opened 2023-02-09
Updates and other software: disable codecs when offline

jpnurmi:
| Online | Offline |
|---|---|
| Screenshot from 2023-02-09 17-41-50 | Screenshot from 2023-02-09 17-41-56 |

Close: #1385


seb128:
The change makes sense since we can't provide the required packages on the image so we need internet access to do the job. 
@elioqoshi should we also include some clues in the UI about why the option can't be selected?

d-loose (approved):
LGTM :+1:

seb128:
I've moved the design question in https://github.com/canonical/ubuntu-desktop-installer/issues/1390 so we don't block merging the current improvement

Issue #1373, opened 2023-02-07
Message is truncated in slideshow


jibel:
Lunar 20230207

The message on the last slide of the slideshow is incomplete and there is no scrollbar to read then end.

![image](https://user-images.githubusercontent.com/1928546/217273681-5b19910a-442c-4853-ba98-4911a7025328.png)

jpnurmi:
Thanks for bringing it up! I'll hold off with this one a little because @local-optimum and @elioqoshi are updating the slides.

elioqoshi:
Would it be possible to scale the content if needed for languages which have more than x characters?

jpnurmi:
Scaling the font dynamically is an option but I'm a bit worried that it results in arbitrarily changing font sizes between slides.

Alternatively, we could tweak the font size and implement a test that ensures there's no overflow in any language we have translations for. We'd be at the mercy of our translators but at least we'd catch the issue if any translated string suddenly would no longer fit after a translation update...

seb128:
Not directly related to fixing it but what's the layout difference with the old slideshow which didn't have the issue, afaik we didn't change the content?

jpnurmi:
The old slides are HTML, and the size is different (752x376 vs. 960x480). We have loosely replicated the looks with Flutter widgets because Flutter for Linux doesn't have support for WebView yet.

jpnurmi:
The left-side font is quite heavy compared to Ubiquity.
![image](https://user-images.githubusercontent.com/140617/218039481-87cffc4c-bb77-4ebe-8c59-2ff59e6c477a.png)

The situation is not that bad, actually. We should get away with minor font and layout tweaks.

![photos-uk](https://user-images.githubusercontent.com/140617/218038713-7c50e50d-2cca-4fbf-a428-b7b7cfe58112.png)
![photos-ru](https://user-images.githubusercontent.com/140617/218038723-6a5955fd-e3cc-429c-8554-eba2c92d7248.png)
![photos-pl](https://user-images.githubusercontent.com/140617/218038727-f591be3c-0140-462c-90b9-bfd590628a8a.png)
![photos-ja](https://user-images.githubusercontent.com/140617/218038731-d190aada-4729-441e-bd3f-3f2cd0ca8c04.png)
![photos-hu](https://user-images.githubusercontent.com/140617/218038735-5a3c547e-c8b1-4c87-9221-87b7dd0cbca0.png)
[...]

PR #1388, opened 2023-02-09
Installation slides: reduce the font size to avoid overflow

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2023-02-09 17-53-44 | Screenshot from 2023-02-09 17-53-52 |

Close: #1373


seb128:
I think the change makes sense and look fine so +1 from me but we should at least get @elioqoshi to review before merging.

seb128 (approved):

jpnurmi:
I wrote some tests to run all slides with all translations and it caught more overflows, also on other slides that wouldn't be fixed by this change alone:

- https://github.com/canonical/ubuntu-desktop-installer/pull/1389

elioqoshi:
Looks good to me!

PR #1399, opened 2023-02-13
Remove unused dio dependency

jpnurmi:
I noticed Renovate trying to offer an update but we're not even using dio in the installer.

- https://github.com/canonical/ubuntu-desktop-installer/pull/1398


seb128 (approved):
thanks

Issue #1390, opened 2023-02-10
Doesn't explain why the codec installation option is disabled


seb128:
The codecs packages can't be provided on the ISO so the option isn't selectable unless the system is online. The reason/solution might not be obvious to the user though, should we have a string hint about it?

@elioqoshi

elioqoshi:
We could add an information label saying the user is offline instead, I agree. @jpnurmi Should I mock something up or would you want to propose something?

jpnurmi:
@elioqoshi please do if you have time. :) for reference, there's a warning when on battery:

![image](https://user-images.githubusercontent.com/140617/218073537-715b74fe-073a-4aee-a90d-99b664c8811e.png)

P.S. Ugh, that red error color needs to be fixed.

elioqoshi:
I'd propose something like this as a warning. Red is a critical error semantic and too alarming for this.

![grafik](https://user-images.githubusercontent.com/5436686/218084464-4fb34a13-7bcf-4649-ab92-0a9af57eff3d.png)

elioqoshi:
Bonus: We can update the label to be more specific that it downloads something:

`Download and install support for additional media formats`

![grafik](https://user-images.githubusercontent.com/5436686/218084820-31685388-e789-4903-882b-ae4659088eaa.png)

jpnurmi:
I guess a red warning makes sense when running on battery because it could be dangerous if you run out of battery at a critical moment. Anyway, there's a chance that a system is both offline and on battery at the same time. Would you show both messages in a column, for example?

dbungert:
From that mockup I'm not sure the relationship between being offline and the codecs checkbox is going to be clear.  What about moving the warning closer to the checkbox?  Maybe the user does want the codecs but just skipped the network step and they plan to resolve it later, that should indicate that setting up network would give the option to fix the problem.

kenvandine:
I agree with @dbungert, it would also be cool to be able to take them back to the page to connect to wifi.

elioqoshi:
We can move the label closer to the option:

![grafik](https://user-images.githubusercontent.com/5436686/218253452-18cd6fe4-f22b-4006-bb72-70d7b28c0098.png)


I do not think we need another option for connecting to the wifi beyond the user going back manually, as that was a user's deliberate choice.

jpnurmi:
@elioqoshi Could it be an option to show the warning in a tooltip instead when hovering the disabled checkbox?

[...]

PR #1408, opened 2023-02-15
Updates and other software: explain disabled codecs with a tooltip

jpnurmi:
- The text was changed: "Install support for additional media formats" => "Download and install support for additional media formats"
- A "You are currently offline" tooltip is shown when hovering the disabled checkbox

| Light | Dark |
|---|---|
| image | image |

Close: #1390


jpnurmi:
@elioqoshi Is this proposal ok?

elioqoshi:
Structurally yes, but the tooltip is not accessible on light as orange lacks color contrast on white. It needs to be brownish in order to pass the contrast.

Also is this the standard tooltip we use? The shadow looks a bit weird

jpnurmi:
Ok, hmm... the warning color comes from Yaru. Actually, do we need to change the text color or would it be enough to only use it for the warning icon?

This is not the standard Flutter tooltip because it can only contain text. It looked like this :(

Screenshot from 2023-02-15 08-01-37

elioqoshi:
I think the standard tooltip looks fine as it's informational

jpnurmi:
Ok, thanks! I've changed it to use the basic tooltip.

d-loose (approved):
:+1:

PR #1415, opened 2023-02-16
Who are you: add "Use Active Directory" checkbox

jpnurmi:
NOTE: AD join result handling during installation follows as a separate step.

image

Ref: #41


OpenOS-Project-Creator-Founder:
> NOTE: Not connected to subiquity yet.
> 
> Screenshot from 2023-02-16 17-07-55
> 
> Close: #41
> 



Will this also be downstreamed to Kubuntu as a KCM System Settings Module?

jpnurmi:
> Will this also be downstreamed to Kubuntu as a KCM System Settings Module?

This is just a thin Flutter (Dart) GUI on top of Subiquity (Python) that implements the actual AD logic. I'm not sure how this could be a KCM module which is QML and C++ if I understood correctly. Perhaps the following Subiquity changes would help for adapting the logic at least, which is probably the most interesting part?

- https://github.com/canonical/subiquity/pull/1550
- https://github.com/canonical/subiquity/pull/1553
- https://github.com/canonical/subiquity/pull/1563

redgryphon:
Is it good to label the username textbox "Domain Administrator" when using that kind of users is not the right thing to do nowadays? Why not the more generic "Domain Join User"?

Feichtmeier:
@jpnurmi 

There is one problem with this screen in Ubiquity I noticed while enjoying to set up active directory with Ubuntu clients in our University:

Pressing the button "test connection" only tests if the domain name is valid and discovered. I think, it should test all three things, domain name, username, password. Which would also need moving the button beneath all three components.

How is the behavior with the flutter implementation? Does it also test the user name and password?

jpnurmi:
@Feichtmeier As far as I know, it's not possible to establish an actual AD connection at this point. All "Test connection" does is test that the given address is reachable. AD is actually set up sometime later when the relevant packages have been installed. Sounds a bit confusing to potentially get an error 15 minutes later but this is how I understood it works.

jpnurmi:
Actually, looking at the new API available in Subiquity, we might not have a separate "Test connection" button at all because we're supposed to validate the domain on the fly as the user types.

Feichtmeier:
Yes the problem is that when the credentials are wrong you have no GUI tool to correct your mistake later on as all the SSSD helper packages are removed again after the installation finished.

At least in powershell it is possible to test the credentials, no idea about [SSSD](https://ubuntu.com/server/docs/service-sssd-ad)

If it is not possible I would propose to change the label to something like "Test the domain name"

OpenOS-Project-Creator-Founder:
Would it be possible to deploy Active Directory as LXC/LXD?

What would be the benefits of this would be the Cluster-Mode within Intranets.

Especially for shared applications for schools, colleges, universities...

Also, for Offices, Small Businesses & Big Businesses...

Having AD as LXC/LXD also adds a few more layers of privacy & security...

OpenOS-Project-Creator-Founder:
> > Will this also be downstreamed to Kubuntu as a KCM System Settings Module?
> 
> This is just a thin Flutter (Dart) GUI on top of Subiquity (Python) that implements the actual AD logic. I'm not sure how this could be a KCM module which is QML and C++ if I understood correctly. Perhaps the following Subiquity changes would help for adapting the logic at least, which is probably the most interesting part?
> 
> - https://github.com/canonical/subiquity/pull/1550
> - https://github.com/canonical/subiquity/pull/1553
> - https://github.com/canonical/subiquity/pull/1563

I found this,

https://github.com/therecipe/flutter

Would this help with Kubuntu?

CarlosNihelton:
> Yes the problem is that when the credentials are wrong you have no GUI tool to correct your mistake later on as all the SSSD helper packages are removed again after the installation finished.
> 

With the current approach through subiquity the packages will remain installed on the system, so if troubleshooting or manual joining is required, at least you won't need to reinstall stuff. 

With the infra we have currently the only reliable way to test the credentials is by doing the actual join. We could mess with sssd configuration and attempt to authenticate without joining, but that can be fragile. And joining the live session is not desirable at this point. While that's something to keep researching for the near future, right now our priority is getting in feature-parity with Ubiquity (tiny improvements are welcome, but that one would certainly not be tiny at all 😉  )

jpnurmi:
@jibel a couple of AD-related naming proposals came up:

- @redgryphon proposed using "domain join user" instead of "domain administrator", and
- @Feichtmeier proposed using "test domain name" instead of "test connection"

are you able to help confirm if these changes are ok?

packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_model.dart
```diff
@@ -96,8 +100,15 @@ class WhoAreYouModel extends PropertyStreamNotifier {
   bool get autoLogin => _autoLogin.value;
   set autoLogin(bool value) => _autoLogin.value = value;
 
-  /// Returns true if there is a site-wide connectivity.
-  bool get isConnectedSite => _network.isConnectedSite;
+  /// Returns true if there is at least site-wide connectivity.
+  bool get isConnected => _network.isConnected || _network.isConnectedSite;
+
+  /// Whether Active Directory is support available or null when unknown.
```
CarlosNihelton, line 106:
```suggestion
  /// Whether Active Directory support is available or null when unknown.
```

CarlosNihelton (commented):
One minor nitpick and jibel's response about [this comment](https://github.com/canonical/ubuntu-desktop-installer/pull/1415#issuecomment-1458467213) and we should be good to go!

jpnurmi:
> and jibel's response about [this comment](https://github.com/canonical/ubuntu-desktop-installer/pull/1415#issuecomment-1458467213) and we should be good to go!

That part of the GUI was already prepared and merged in #1522 so I've created a separate ticket to track the naming proposals:
- https://github.com/canonical/ubuntu-desktop-installer/issues/1582

CarlosNihelton (approved):
👍🏽 then.

PR #1371, opened 2023-02-06
Show the window full screen if it cannot fit the workarea (- margins)

jpnurmi:

PR #1418, opened 2023-02-17
Try or install: use `yaru_window` instead of `window_manager`

jpnurmi:
Migrating to `yaru_window` will eventually allow us to show the window fullscreen when it doesn't fit on the screen (#1371).


jpnurmi:
The page looks like this. If you select _Try Ubuntu_ and press _Continue_, it'll close the window.

image

d-loose (approved):
:+1:

PR #1419, opened 2023-02-17
Turn off BitLocker: use `yaru_window`

jpnurmi:
Ref: #1371

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/turn_off_bitlocker/turn_off_bitlocker_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/turn_off_bitlocker/turn_off_bitlocker_model.dart
@@ -1,17 +1,17 @@
 import 'package:subiquity_client/subiquity_client.dart';
 import 'package:ubuntu_logger/ubuntu_logger.dart';
-import 'package:ubuntu_wizard/utils.dart';
 
 /// @internal
 final log = Logger('turn_off_bitlocker');
 
 /// View model for [TurnOffBitLockerPage].
-class TurnOffBitLockerModel with SystemShutdown {
+class TurnOffBitLockerModel {
   /// Creates an instance with the given client.
-  TurnOffBitLockerModel(this.client) {
+  TurnOffBitLockerModel(this._client) {
     log.info('BitLocker must be turned off');
   }
 
-  @override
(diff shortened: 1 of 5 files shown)


packages/ubuntu_desktop_installer/lib/pages/turn_off_bitlocker/turn_off_bitlocker_model.dart
```diff
@@ -1,17 +1,17 @@
 import 'package:subiquity_client/subiquity_client.dart';
 import 'package:ubuntu_logger/ubuntu_logger.dart';
-import 'package:ubuntu_wizard/utils.dart';
 
 /// @internal
 final log = Logger('turn_off_bitlocker');
 
 /// View model for [TurnOffBitLockerPage].
-class TurnOffBitLockerModel with SystemShutdown {
```
jpnurmi, line 9: The `SystemShutdown` "helper" is no longer very helpful because `yaru_window` is not a singleton so we can't/shouldn't call it straight from a model as we did with the `window_manager` which is a singleton. Once we move the `window.close()` call to the widget tree, the system shutdown helper doesn't really add any value anymore... :man_shrugging: 

```dart
/// Provides system shutdown and reboot functionality, and performs the
/// appropriate application cleanup routines to close the window, subiquity
/// socket, and the server if was started by this UI instance.
mixin SystemShutdown {
  /// The client to send shutdown and reboot requests to.
  SubiquityClient get client;

  /// Request system reboot and close the window.
  ///
  /// The [immediate] argument should be `true` if the system should be rebooted
  /// immediately before the installation has finished.
  Future<void> reboot({required bool immediate}) {
    log.info('request reboot (immediate: $immediate)');
    return client.reboot(immediate: immediate).then((_) => closeWindow());
  }

  /// Request system shutdown and close the window.
  ///
  /// The [immediate] argument should be `true` if the system should be shut
  /// down immediately before the installation has finished.
  Future<void> shutdown({required bool immediate}) {
    log.info('request shutdown (immediate: $immediate)');
    return client.shutdown(immediate: immediate).then((_) => closeWindow());
  }
}
```

jpnurmi:
The page looks like this btw just to get a feeling what we're talking about :)

image

d-loose (approved):
LGTM :+1:

Issue #1402, opened 2023-02-14
Superfluous radio button when there is no network


jibel:
Lunar 20230214

When the machine is not connected to a network (NIC deactivated), there is a single radio button with the label "I don't want to connect to the network"
This radio button shouldn't be displayed or should be greyed out since, in any case, you cannot connect to the network.

![image](https://user-images.githubusercontent.com/1928546/218698973-ca8de69b-7245-44fc-baf6-573742cf8eae.png)

PR #1423, opened 2023-02-17
Connect to internet: disable superfluous radio button

jpnurmi:
If both ethernet and wifi are disabled or unavailable, disable the "I don't want to connect to the internet just now" radio button because there's nothing to select.

Screenshot from 2023-02-17 14-20-34

Fixes: #1402


d-loose (approved):
:+1:

PR #1427, opened 2023-02-17
Remove old unused InstallationSlideModel.reboot()

jpnurmi:
We have a dedicated "installation complete" page and we haven't had the need to reboot straight from the installation slides for a very long time. This kills one use of the `SystemShutdown` mixin which is on the way out...


CarlosNihelton (approved):
Less is more! 👍🏽

PR #1430, opened 2023-02-20
WSL Applying changes: use `yaru_window`

jpnurmi:
The end goal is to migrate all top-level window access from `window_manager` to `yaru_window` because the latter is a) compatible with `handy_window` and b) we have full control over it so we are able to quickly fix & publish whenever needed.

The main API difference is that `yaru_window` is not a singleton but accessed from the widget tree instead using the context as a reference, following the direction of the ongoing top-level window work in Flutter. This makes the old `SystemShutdown` helper mixin useless, though, because we can't/shouldn't access the window directly from models anymore which was a bit questionable choice anyway, by yours truly. :grin: 

Ref: #1371


d-loose (approved):
LGTM :+1:

Issue #1432, opened 2023-02-20
Occasional storage-related exception on startup


jpnurmi:
Hot restart occasionally causes subiquity to throw an exception. The log reveals that on startup, `GET /storage` is called before `POST /source`.

```
Restarted application in 273ms.
flutter: INFO ubuntu_desktop_installer: Logging to /home/jpnurmi/Projects/canonical/ubuntu-desktop-installer/packages/ubuntu_desktop_installer/build/linux/x64/debug/bundle/.ubuntu_desktop_installer/ubuntu_desktop_installer.log
flutter: unhandled element defs; Picture key: AssetBundlePictureKey(bundle: ProxyAssetBundle#dc931(), name: "assets/welcome/logo.svg", colorFilter: null, theme: SvgTheme(currentColor: Color(0xff000000), fontSize: 14.0, xHeight: 7.0))
flutter: INFO subiquity_server: Starting server (PID: 41765) with args: [-m, subiquity.cmd.server, --dry-run, --machine-config=examples/simple.json, --storage-version=2]
flutter: INFO subiquity_server: Waiting server up to 30 seconds
flutter: ERROR subiquity_server: SocketException: Connection failed (OS Error: Connection refused, errno = 111), address = ../subiquity_client/subiquity/.subiquity/socket, port = 0
flutter: INFO subiquity_server: ApplicationState.WAITING
flutter: INFO subiquity_client: Opening socket to Endpoint(../subiquity_client/subiquity/.subiquity/socket )
flutter: INFO subiquity_status: null => ApplicationStatus(state: ApplicationState.WAITING, confirmingTty: , error: null, cloudInitOk: true, interactive: true, echoSyslogId: subiquity_echo.41765, logSyslogId: subiquity_log.41765, eventSyslogId: subiquity_event.41765)
flutter: DEBUG subiquity_client: POST http://localhost/meta/client_variant?variant=%22desktop%22
flutter: DEBUG subiquity_client: GET http://localhost/storage/v2?wait=true
flutter: DEBUG subiquity_client: GET http://localhost/storage/has_rst
flutter: DEBUG subiquity_client: GET http://localhost/storage/has_bitlocker
flutter: INFO welcome: Loaded 74 languages
flutter: DEBUG subiquity_client: ==> setVariant("desktop") null
flutter: DEBUG subiquity_client: POST http://localhost/source?source_id=%22ubuntu-desktop%22
[...]

PR #1433, opened 2023-02-20
Fix storage init on startup

jpnurmi:
Post-pone `getStorage()` until after `setSource()` to let Subiquity know what kind of installation it's dealing with before asking for storage details including how much space the installation requires.

Also, even if mockito's `untilCalled()` made the easy to test, it's generally a better idea to trigger asynchronous initialization routines from the outside than from the service's own constructor because there's no way to get ahold of the future...

Fix: #1432


d-loose (approved):
:+1:

PR #1436, opened 2023-02-20
README.md: update instructions

jpnurmi:
Running `make install_deps` as non-root is better as it depends on `gitdeps` which builds the local curtin & probert as the local user and `aptdeps` that uses sudo to install the relevant dependencies. The `installdeps.sh` script is somewhat intrusive because it runs dist-upgrade etc. and is supposedly primarily meant for the CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,7 @@ git submodule update --init --recursive
 
 Install Subiquity dependencies:
 ```sh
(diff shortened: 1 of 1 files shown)


d-loose (approved):
That's really good to know! :+1:

PR #1412, opened 2023-02-16
Initial autoinstall support

jpnurmi:
- The existing (welcome-like) loading screen is reused until we have a design for a proper loading screen that needs to be shown on startup until we get a status response from subiquity.
- If Subiquity requires a confirmation (no "autoinstall" kernel cmdline), the wizard jumps straight to the _Write changes to disk_ page and after that, to the installation slides.
- If Subiquity does not required a confirmation ("autoinstall" was supplied on the kernel cmdline), the wizard jumps straight to the installation slides.

| Startup | Confirm | Auto |
|---|---|---|
| image | image | Screenshot from 2023-02-17 18-17-55 |

NOTE: Semi-automated installation with interactive sections is not supported, yet.

PR #1437, opened 2023-02-20
Write changes to disk: fix storage API calling order

jpnurmi:
Split off from #1412 and coupled with a unit test.

Make sure to call `GET /storage/v2?wait=true` before calling `GET /storage/v2/orig_config`. The former blocks until Subiquity is done probing the hardware, whereas the latter throws if Subiquity is not done probing the hardware.

The correct order is important in autoinstall mode because the GUI jumps straight to the confirmation screen if Subiquity reports `NEEDS_CONFIRMATION` after establishing the connection. At that point, Subiquity might be still probing the hardware in the background.


d-loose (approved):
:+1:

PR #1438, opened 2023-02-20
Remove unused `SystemShutdown` mixin

jpnurmi:
All it did was to call `closeWindow()` after calling `reboot()` or `shutdown()` but it's no longer used after the gradual migration to `yaru_window`.


d-loose (approved):
:+1:

PR #1442, opened 2023-02-21
Migrate remaining top-level window management to `yaru_window`

jpnurmi:
This is the last step in the series of migrating to `yaru_window`:
https://github.com/canonical/ubuntu-desktop-installer/pulls?q=is%3Apr+yaru_window+

All remaining top-level window management has been migrated from `window_manager` to `yaru_window` because the latter is a) compatible with `handy_window` and b) we have full control over it so we are able to quickly fix & publish whenever needed.

After this, we'll be finally able to show the installer window in fullscreen mode in #1371 when it doesn't fit the available workarea.


jpnurmi:
I opened a ticket to improve yaru_window testability. Not happy with all those MethodChannel instances created all over the place in tests... :P

- https://github.com/ubuntu-flutter-community/yaru_window.dart/issues/14

d-loose (approved):
:+1:

PR #1444, opened 2023-02-21
Use handy_window without widget tree manipulation

jpnurmi:
Create a `HdyWindow` in `my_application.cc` instead of letting `handy_window` automagically reshuffle the widget tree under the hood, and restore the original order of showing the window and the view as it is in the Flutter app template.

Ref: #1371


d-loose (approved):
LGTM :+1:

PR #1450, opened 2023-02-22
Who are you: switch to a trailing "show/hide" password button

jpnurmi:
This PR implements a new design by @elioqoshi to make room for the upcoming "Use Active Directory" checkbox (#41 & #1415).

## Before

image

## After

[Screencast from 2023-02-22 16-20-34.webm](https://user-images.githubusercontent.com/140617/220667907-1b5c814b-4391-4b40-8dd8-a005c9a7a107.webm)


elioqoshi:
I have a problem with the active and error state colors but that's a separate thing from this issue, which I think is an improvement.

Jupi007:
@jpnurmi  Maybe can you add a yaru eye/hide icon depending on the state (like gnome-shell does in password inputs)?

jpnurmi:
@Jupi007 We had an eye icon in the beginning but we switched to a textual checkbox for accessibility reasons. :)

Jupi007:
@jpnurmi In fact I mean to add the icon before (or after) the label so it looks like this: `👀 Show`

elioqoshi:
If the space allows I think that makes sense

jpnurmi:
| Show | Hide |
|---|---|
| image | image |

Jupi007:
Looks very nice :+1:

jpnurmi:
@elioqoshi I added a screenshot above. Looks good to me and Paul, do you approve?

elioqoshi:
Looking good!

PR #1445, opened 2023-02-21
Add page indicator

d-loose:
[Screencast from 2023-02-21 15-30-17.webm](https://user-images.githubusercontent.com/113362648/220372982-2abb96b9-d8e5-4da6-bba6-915772c74b26.webm)

Depends on https://github.com/ubuntu-flutter-community/wizard_router/pull/30
Fixes #1348

PR #1455, opened 2023-02-23
Show battery warning in a snack bar

d-loose:
See https://github.com/canonical/ubuntu-desktop-installer/pull/1445#issuecomment-1441744399

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/updates_other_software/updates_other_software_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/updates_other_software/updates_other_software_page.dart
@@ -100,10 +100,22 @@ class _UpdatesOtherSoftwarePageState extends State<UpdatesOtherSoftwarePage> {
           ),
         ],
       ),
-      footer: model.onBattery
-          ? Html(
-              data: lang.onBatteryWarning(
-                  Theme.of(context).colorScheme.error.toHex()),
+      snackBar: model.onBattery
+          ? SnackBar(
+              width: 500,
+              dismissDirection: DismissDirection.none,
+              behavior: SnackBarBehavior.floating,
+              content: Html(
+                style: {
+                  'body': Style(
+                    color: Theme.of(context).colorScheme.onInverseSurface,
+                  )
+                },
+                data: lang.onBatteryWarning(
+                    Theme.of(context).colorScheme.error.toHex()),
+              ),
+              duration: const Duration(days: 1),
+              showCloseIcon: true,
             )
           : null,
       actions: <WizardAction>[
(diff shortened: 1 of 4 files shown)


elioqoshi:
@d-loose Would appreciate if for every PR we can also have a screenshot for light and dark versions :)

packages/ubuntu_desktop_installer/lib/pages/updates_other_software/updates_other_software_page.dart
```diff
@@ -100,10 +100,23 @@ class _UpdatesOtherSoftwarePageState extends State<UpdatesOtherSoftwarePage> {
           ),
         ],
       ),
-      footer: model.onBattery
-          ? Html(
-              data: lang.onBatteryWarning(
-                  Theme.of(context).colorScheme.error.toHex()),
+      snackBar: model.onBattery
+          ? SnackBar(
+              width: 500,
```
jpnurmi, line 105: did you try what happens if the text is too long to fit? does it wrap?

d-loose:
Screenshots:
| Light | Dark |
| - | - |
| image | image |

packages/ubuntu_desktop_installer/lib/pages/updates_other_software/updates_other_software_page.dart
d-loose: Yes, it wraps
image

jpnurmi (approved):
Sorry for the delay, LGTM! :+1: And thanks for splitting it off separately :)

jpnurmi:
@elioqoshi Do you think we could replace the red "Warning" label with an icon? It could potentially look a bit nicer

elioqoshi:
I think using an icon is fine here as well.

d-loose:
> The theming is off I think. Ideally they should be reverted with the snackbar background color slightly different from the screen color. What do you think?

It's probably an issue for yaru. In principle I agree with you, although this is currently in accordance with the [M3 specs](https://m3.material.io/components/snackbar/specs#3c1ed6ed-2314-408b-aa75-634de186ec0a).

jpnurmi:
@elioqoshi Do you think we could move forward with this or would you prefer we try to tweak the colors in yaru.dart first?

elioqoshi:
I think we should solve the root issue and not the overlying issue here so we should go for this as it is now and address yaru.dart separately, what do you think?

jpnurmi:
Ok, thanks! I think it looks decent enough to be on the daily image like this for a while. It's hard to say how long it might take to address this in yaru.dart. :)

elioqoshi:
@jpnurmi do you think we can address it in yaru.dart before 23.04 release?

jpnurmi:
@elioqoshi we can propose changes right away. would the looks of GNOME notifications work for us?

image

elioqoshi:
I think that would work yeah. I can finetune the design if needed.

PR #1458, opened 2023-02-24
Integration tests: simplify tapping buttons

jpnurmi:
CommonFinders.widgetWithText() uses find.byType(T) which does not test subclass types but we can use find.bySubtype<T>() to be able to support any ButtonStyleButton because we don't really care about the exact type of button in the high level integration tests. All those small details are tested in widget tests.


d-loose (approved):
LGTM :+1:

Issue #1446, opened 2023-02-21
Manual partitioning on smaller screens


jpnurmi:
Since #1371, we're showing the window in fullscreen mode if the window (960x680) doesn't fit the available workarea on the screen. With this, most pages work surprisingly well in 800x600 which is a common VM screen size. The manual partitioning page could ideally leave a bit more space for the table rows. It feels like too much space is eaten up by the various controls around, and the use of empty spacing is also pretty generous.

![image](https://user-images.githubusercontent.com/140617/220415330-5c0a2e8c-c828-4a6f-bdc3-fcee6d708f93.png)

PR #1474, opened 2023-02-28
Allocate disk space: improve layout for smaller screens

jpnurmi:
- reduce partition bar height and partition legend spacing
- reduce table header height to match disk/partition rows
- move the bootloader selector label as part of the input decoration

On an 800x600 screen, these changes result in roughly two extra disk/partition rows.

| Before | After |
|---|---|
| image | image |

Fix: #1446

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_page.dart
@@ -83,7 +83,7 @@ class _AllocateDiskSpacePageState extends State<AllocateDiskSpacePage> {
           Expanded(child: PartitionTable(controller: _scrollController)),
           const SizedBox(height: kContentSpacing / 2),
           const PartitionButtonRow(),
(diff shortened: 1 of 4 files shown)


jpnurmi:
Depends on
- https://github.com/canonical/ubuntu-flutter-plugins/pull/154

elioqoshi (approved):
Great improvement, LGTM!

elioqoshi:
Unrelated: The accent color dropdowns really look like error states and that worries me with the Ubuntu color (or red-ish accent colors):

grafik

Should address that in other issues

jpnurmi:
Agreed! :+1: It uses the same "input decoration" as text fields. I opened a ticket earlier but I might need some help to fix it :smile: 
- https://github.com/ubuntu/yaru.dart/issues/290

jpnurmi:
With reduced partition bar height and spacing we got an additional half-row visible in 800x600 which helps to see that the table content is scrollable.

image

Luckily enough, it matches half a line with the default 960x680 window size, too.

image

d-loose (approved):
:+1:

PR #1477, opened 2023-02-28
Parse autoinstall journal events

### LINKED ISSUE #1454: Autoinstall: copying files
The GUI is normally able to detect what Subiquity is doing, and presents different status messages during different phases of the installation process.

image
image

During automated installations, however, the message seems to be always "Copying files" which is a fallback used when the GUI doesn't know what Subiquity is doing:

image

d-loose:
Fix #1454

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
@@ -103,21 +103,13 @@ class InstallationSlidesModel extends SafeChangeNotifier {
   //     subiquity/Install/install/curtin_install/run: executing curtin install partitioning step
   // ```
   void _processEvent(String syslog) {
+    syslog = syslog.replaceFirst(RegExp(r'  subiquity/[\w/]+: '), '');
     final trimmed = syslog.trimLeft();
-    if (trimmed.startsWith('subiquity')) {
-      if (trimmed == syslog) return;
-      final tokens = trimmed.split(':').map((e) => e.trim()).toList();
-      if ('  $trimmed' == syslog && tokens.length >= 2) {
-        _event = InstallationEvent.fromString(tokens[1]);
-      } else if (tokens.length >= 2) {
-        _event = _event.copyWith(description: tokens[1]);
-      }
+    if (trimmed.startsWith('subiquity')) return;
+    if (trimmed == syslog) {
+      _event = InstallationEvent.fromString(syslog);
     } else {
-      if (trimmed == syslog) {
-        _event = InstallationEvent.fromString(syslog);
-      } else {
-        _event = _event.copyWith(description: trimmed);
-      }
+      _event = _event.copyWith(description: trimmed);
     }
     notifyListeners();
   }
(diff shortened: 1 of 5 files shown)


packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
```diff
@@ -21,6 +22,17 @@ class InstallationEvent {
       description: description ?? this.description,
     );
   }
+
+  String localize(AppLocalizations lang) {
```
jpnurmi, line 26: i'd prefer to keep translations out of the model. could we use an enum instead?

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_page.dart
```diff
@@ -145,7 +133,9 @@ class _InstallationSlidesPageState extends State<InstallationSlidesPage> {
                     Text(
                       model.hasError
                           ? lang.installationFailed
-                          : _formatEvent(model.event),
+                          : (model.event ??
+                                  const InstallationEvent('copying files'))
```
jpnurmi, line 137: perhaps we could return this as a fallback from the model?

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
```diff
@@ -70,12 +82,31 @@ class InstallationSlidesModel extends SafeChangeNotifier {
...
   //   downloading and installing security updates
   // ```
+  // In autoinstall mode syslog events start with "subiquity" and unindented
+  // entries contain no useful information.
+  // ```
+  // subiquity/Drivers/_list_drivers
+  //   subiquity/Install/install/unmount_target: umounting /target dir
+  //   subiquity/Install/install/curtin_install: installing system
+  //     subiquity/Install/install/curtin_install/run: executing curtin install initial step
+  //     subiquity/Install/install/curtin_install/run: executing curtin install partitioning step
```
jpnurmi, line 92: could we add a test with these lines?

packages/ubuntu_desktop_installer/lib/pages/installation_slides/installation_slides_model.dart
```diff
@@ -70,12 +82,31 @@ class InstallationSlidesModel extends SafeChangeNotifier {
...
+  // ```
   void _processEvent(String syslog) {
     final trimmed = syslog.trimLeft();
```
jpnurmi, line 95: a crazy idea; what if we would do something like
```dart
syslog.removeOne(RegExp(r'  subiquity/[\w/]+:'))
```
as the first step? would the rest of the function work as before?

jpnurmi: i meant `replaceFirst(rx, '')` :)

d-loose: good idea - works

jpnurmi (approved):
Nice! LGTM :+1:

Issue #1467, opened 2023-02-25
redesign 'Go Back' and 'Continue' buttons


elioqoshi:
Currently the primary and secondary button are styled similarly. Is there any background why the primary button is not styled as such (green)? @jpnurmi 

Considering that primary button is reserved for the one time Installation is started, could we see to distinguish Go Back and Continue with other kinds of buttons?

![image](https://user-images.githubusercontent.com/5436686/221368858-ffec5dc3-c459-4540-b031-74d761731241.png)

jpnurmi:
I don't know the background but the pattern is that buttons that modify the system are highlighted. "Go Back" and "Continue" are basically just navigating.

![image](https://user-images.githubusercontent.com/140617/221371865-d3bd5930-a48d-4c84-95cb-fb9abcc9f9a8.png)
![image](https://user-images.githubusercontent.com/140617/221371838-216abbee-458f-41e4-bc69-53516a12bbfe.png)
![image](https://user-images.githubusercontent.com/140617/221371848-281a6886-3298-43a8-b52d-476cd148efa8.png)
![image](https://user-images.githubusercontent.com/140617/221371946-bc943efb-c6a7-4139-b579-2fd4a923b465.png)

elioqoshi:
I'd still recommend we distinguish Go Back from Continuing. Can we use buttons without any outlines for Go Back maybe? What are the available button styles we have at our disposal again?

elioqoshi:
Since we are talking about this: ![grafik](https://user-images.githubusercontent.com/5436686/221372549-e0825493-1bd3-484d-a6f4-cdb08f5b25db.png)

This is likely my fault, however the order of Primary and Secondary button is inverted in the 'Continue Testing' screen. We should use the usual order, so "Secondary | Primary"

elioqoshi:
Thanks!

Can I propose to use Text buttons for secondary actions (Go Back) and filled buttons for primary actions (Continue)?

jpnurmi:
Sure, here's how that combination would look:

![image](https://user-images.githubusercontent.com/140617/221519494-bed5c140-01b6-4ba7-85ee-b5f14b496a56.png)

elioqoshi:
I think that looks better, though are we set on the accent color for the text button? The issue that might create is that it looks highlighted.

jpnurmi:
[...]

PR #1484, opened 2023-02-28
Move the _Go Back_ button

jpnurmi:
| Before | After |
|---|---|
| image | image |
| image | image |

Ref: #1467

### DIFF
--- a/packages/ubuntu_wizard/lib/src/widgets/wizard_page.dart
+++ b/packages/ubuntu_wizard/lib/src/widgets/wizard_page.dart
@@ -22,7 +22,6 @@ class WizardPage extends StatefulWidget {
     this.content,
     this.contentPadding = kContentPadding,
     this.contentSpacing = kContentSpacing,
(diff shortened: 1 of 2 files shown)


elioqoshi:
@jpnurmi I want to make sure this looks okay for all instances and all screens. Would it be too much trouble to ask for screenshots (one theme is enough) of the more exotic screens (i.e partitioning) to ensure this looks good there too?

jpnurmi:
Thanks for asking, it's a fair request for a change like this and I must admit I didn't even check all screens myself! (oops)

Styling and label changes will follow separately. :)

image
image
image
image
image
image
image
image
image
image
image
image

image
image
image
image
image

elioqoshi:
@jpnurmi this looks great! I don't see any issues at first sight with these layouts. There might be a few things we can improve on specific pages but I am following up with them separately. This is good to go I think

elioqoshi (approved):

Issue #1466, opened 2023-02-25
Differ animations for buttom buttons across screens


elioqoshi:
While the slide and fade out animations feels quite crisp, it might look a bit more coherent with the buttom buttons (Go Back, Continue) not animating, as they are consistently at the same place. Them moving and being replaced by the same buttons feels rather awkward.

[Screencast from 2023-02-25 16-37-24.webm](https://user-images.githubusercontent.com/5436686/221368719-9f747d16-8660-4b8d-bf62-caba067d265a.webm)

jpnurmi:
The original idea was to smoothen the transition from "Continue" to a highlighted "Start installing" that can be double the size in some languages.

![image](https://user-images.githubusercontent.com/140617/221369020-d5c38bf9-b7c5-498e-a7a7-2d8e1ebf637d.png)
![image](https://user-images.githubusercontent.com/140617/221369024-968f7063-19c2-4fb4-8b24-8d8053880e05.png)

jpnurmi:
I must add that the page transitions were different back then. Flutter had some rather subtle vertical page transitions during the early days of the installer.

At some point, thanks to Android and Material design changes, the default page transitions were changed in Flutter. The new transitions felt out of place in the installer. We tried iOS-style horizontal slide transitions for a while but they had a bit too strong mobile phone feeling so we came up with the shorter horizontal transitions we have today.

That's the history. :)

elioqoshi:
Yeah I get that. Would it be possible to use different animations for these buttons than the content uses?

jpnurmi:
Yes, we can try different animations :)

jpnurmi:
I hope we don't need different animations now that the button sizes have been made stable and the bottom bar is no longer animated:
- https://github.com/canonical/ubuntu-desktop-installer/pull/1485

PR #1485, opened 2023-02-28
WizardPage: keep the bottom bar in place during page transitions

jpnurmi:
We're in the process of making sure buttons have consistent sizes between pages (#1487) so we don't need to (ab)use animations to hide the difference anymore.

Keeping the bottom bar in place during page transitions requires a lot fewer changes than implementing different animations for the buttons so I'd suggest we start with this. It looks and feels nice and natural, especially with all the button changes that are halfway through.

## Before

[Screencast from 2023-03-01 06-39-42.webm](https://user-images.githubusercontent.com/140617/222054587-8a9362e6-6d9f-42b6-97a2-ddd3b5bfe2d4.webm)

## After

[Screencast from 2023-03-01 06-38-17.webm](https://user-images.githubusercontent.com/140617/222054416-a2969043-cb55-42d1-93af-ac4c7a1d0d66.webm)

Ref: #1466

### DIFF
--- a/packages/ubuntu_wizard/lib/src/widgets/wizard_page.dart
+++ b/packages/ubuntu_wizard/lib/src/widgets/wizard_page.dart
@@ -115,26 +115,29 @@ class _WizardPageState extends State<WizardPage> {
             SizedBox(height: widget.contentSpacing),
           ],
         ),
-        bottomNavigationBar: Padding(
-          padding: widget.footerPadding,
-          child: Row(
-            children: [
-              _buildAction(context, leading) ?? const SizedBox.shrink(),
-              const Spacer(),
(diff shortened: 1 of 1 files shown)


jpnurmi:
Depends on #1484 to avoid conflicts.

jpnurmi:
@elioqoshi can we start with this easy and simple change? it feels great now that we have fixed the buttons. 🙂 we can always try adding some special animations for the buttons later if this doesn't feel right

elioqoshi (approved):

elioqoshi:
Yeah this is a great change as mentioned. Thanks for flagging again, I missed some issues in the last days with all that back and forth :D

d-loose (approved):
LGTM :+1:

Issue #1397, opened 2023-02-13
Consisten graphical glitches on window corners throughout the installer


local-optimum:
![image](https://user-images.githubusercontent.com/55029393/218415324-efdc28ee-b683-440e-a530-42ea155e61ef.png)

Regardless of dark or light mode, the corners of the installer windows glitch frequently throughout, sometimes on mouseover and sometimes when clicking actions, this produces a lot of flickering during the installer slideshow as well.

[Screencast from 13-02-23 08:56:58.webm](https://user-images.githubusercontent.com/55029393/218415567-465d3dc8-e1fd-47ef-b65d-e41111b1adfc.webm)

Potentially unrelated but when transitioning from the slideshow to the 'installation complete' screen the slideshow is rescaled to about 2/3 the size of the main window oriented to the top left briefly as it fades out creating an ugly transition

jpnurmi:
We use the [Handy](https://gnome.pages.gitlab.gnome.org/libhandy/) library to round the bottom corners of the installer window:

| Handy window | Plain GTK window |
|---|---|
| <image src="https://raw.githubusercontent.com/ubuntu-flutter-community/handy_window/main/images/handy-window.png" width="430"/> | <image src="https://raw.githubusercontent.com/ubuntu-flutter-community/handy_window/main/images/flutter-window.png" width="400"/> |

I don't know what is it with the combination of libhandy and Flutter that is causing this but I'm afraid our best bet is to drop libhandy and use a plain GTK window instead. Rounded bottom corners may look nice and modern but there's very little value in having them if it screws up the other corners...

local-optimum:
Agreed.

jpnurmi:
The bug is reproducible on X11 without Flutter and even without libhandy. I was able to trim the Flutter embedder down to a plain GTK test case. People with AMD cards are unaffected. Turning on VSync helps with Intel at least. Nvidia is unclear.

- https://github.com/jpnurmi/oh-my-corners
- https://gitlab.gnome.org/GNOME/gtk/-/issues/5600
- https://warthogs.atlassian.net/browse/DT-1045 (internal)

elioqoshi:
I had the same issues when I was testing last week. Doing another round for better documentation.

With the current states of things I'd also highly advise to drop libhandy if it causes these issues. For what's worth, dropping round corners (also every where on Desktop) is something I agree with but might need further inspections

jpnurmi:
[...]

PR #1491, opened 2023-03-01
Snap: set `GDK_RENDERING=image` to avoid graphical glitches on X11

jpnurmi:
See:
- #1397
- https://gitlab.gnome.org/GNOME/gtk/-/issues/5600
- https://gitlab.gnome.org/GNOME/gtk/-/issues/4704#note_1676890


kenvandine (approved):
Looks good!

Issue #1469, opened 2023-02-25
Redesign Windows Bitlocker screen


elioqoshi:
The Windows Bitlocker screen is quite neutral and could benefit from an improved layout, as well as semantic Warning/Information

![image](https://user-images.githubusercontent.com/5436686/221369503-6cda78d8-d49d-4997-819a-dca8c762d84c.png)

(TODO for @elioqoshi )

cc @local-optimum it's also rather not ideal looking that help.ubuntu.com/bitlocker links to our Discourse:
https://discourse.ubuntu.com/t/ubuntu-installation-on-computers-running-windows-and-bitlocker-turned-on/15338.

It doesn't feel well thought-through and at worst could make people distrust us. Can we spin up a proper tutorial page for this?

elioqoshi:
Also, as a note, there is no Confirmation dialog after this screen. The machine restarts right away. We would likely need to introduce a little friction here to allow for user errors.

elioqoshi:
After going through this process myself, I am not sure if I did something wrong because of being careless or because the instructions are unclear. It seems that I have overwritten my Windows installation (fresh Windows 10 Pro coming with my Intel NUC). Luckily I had nothing important there yet as it's for testing purposes, but I wonder how this could play out for end-users.

@local-optimum @jpnurmi Could we investigate this a bit further? Unless I did something completely wrong, this seems rather crucial to address.

jpnurmi:
@elioqoshi Could you ZIP and attach the logs in `/var/log/installer` please? If Windows was correctly detected, it should have by default offered to install Ubuntu _alongside_ Windows:

![image](https://user-images.githubusercontent.com/140617/221495410-b44b4b09-5fac-49e0-8cfc-dc72193d1777.png)
![image](https://user-images.githubusercontent.com/140617/221495432-180e906d-906c-4244-af62-5170ec8ebbb5.png)

jpnurmi:
P.S. If we decide to redesign the page, _Turn off RST_ currently looks the same :)

![image](https://user-images.githubusercontent.com/140617/221496019-47099c31-898c-4281-91c2-1b60196c8ddb.png)

elioqoshi:
Correct me if I am not mistaken, but if Bitlocker is on, Ubuntu can't recognize there is a Windows installation, no? @jibel ?

I will have a look and for the log files

kenvandine:
[...]

PR #1492, opened 2023-03-01
Show "Turn off BitLocker" after "Installation type"

jpnurmi:
This fixes a critical issue brought up in #1469: "Turn off BitLocker" was shown after "Write changes to disk" i.e. after the user had already pressed "Start installing".

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/installer.dart
+++ b/packages/ubuntu_desktop_installer/lib/installer.dart
@@ -397,6 +397,10 @@ class _UbuntuDesktopInstallerWizard extends StatelessWidget {
           userData: InstallationStep.type.index,
           onNext: (settings) => _nextStorageRoute(service, settings.arguments),
         ),
+        Routes.turnOffBitlocker: WizardRoute(
+          builder: TurnOffBitLockerPage.create,
+          userData: InstallationStep.storage.index,
+        ),
         Routes.installAlongside: WizardRoute(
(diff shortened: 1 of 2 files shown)


elioqoshi:
We should ensure that this is the right approach and check with QA and others as well in case we are missing something. I want to ensure that we are not missing edge cases here.

packages/ubuntu_desktop_installer/lib/installer.dart
```diff
@@ -457,12 +456,16 @@ class _UbuntuDesktopInstallerWizard extends StatelessWidget {
 
   String? _nextStorageRoute(DiskStorageService service, dynamic arguments) {
     if (arguments == InstallationType.manual) {
-      return Routes.allocateDiskSpace;
+      return service.hasBitLocker
```
jpnurmi, line 459: this is covered by the integration test but we're not collecting test coverage because it essentially runs through the whole app so it'd be too easy to achieve ~100%

jpnurmi:
Hi @linjuiche, we've made a fatal mistake by showing the _Turn off BitLocker_ page after the user has already pressed _Start install_. Would it break test cases if we reorder the pages?

d-loose (approved):
:+1:

linjuiche:
@jpnurmi 

My part of test is manual. SO it will not break. If you are asking automation test part, then we need @mz2 help to check it.

PR #1494, opened 2023-03-01
Fix wizard button padding

jpnurmi:
Remove bogus padding around hidden actions.

## Before

[Screencast from 2023-03-01 17-53-01.webm](https://user-images.githubusercontent.com/140617/222207844-0295a15c-bd9f-4fa4-829c-cd2b190eece8.webm)

## After

[Screencast from 2023-03-01 17-52-07.webm](https://user-images.githubusercontent.com/140617/222207651-975d4d75-ccd9-4146-bbfe-1ab04166c383.webm)


d-loose (approved):
:+1:

PR #1496, opened 2023-03-01
Installation complete: fix alignment

jpnurmi:
| Before | After |
|---|---|
| image | image |

The alignment went unnoticed because my product name is long enough to push it all the way to the left and even wrap... :)

image


elioqoshi (approved):
+1

PR #1497, opened 2023-03-01
Allocate disk space: capitalize "Free space" & "Limit reached"

jpnurmi:
Just updating the English translation which is used as a template for all others. The bot will automatically regenerate the untranslated entries and open a PR. For already translated entries, Weblate will show a warning that they are out of date.

| Before | After |
|---|---|
| image | image |


d-loose (approved):
:+1:

Issue #1294, opened 2023-01-06
Correct Btrfs filesystem description journaling on selection


emanuc:
Hello,
When choosing manual partitioning you have a selection of filesystems to choose from, when selecting Btrfs it is incorrectly described as: "Btrfs journaling file system" [1], it is actually a COW filesystem and does not have journaling [2].

1:  
![installer](https://user-images.githubusercontent.com/27281121/211023910-a7e8deaf-90f7-4833-9128-bd9985a2aae0.png)

2:  
> btrfs is a modern copy on write (CoW) filesystem for Linux
https://btrfs.wiki.kernel.org/index.php/Main_Page

jpnurmi:
Interesting, thanks for bringing it up! It had been copied from the old installer. :)

![image](https://user-images.githubusercontent.com/140617/220425240-9477e56f-1d6f-4019-9431-3aefe9df7897.png)

PR #1501, opened 2023-03-01
Drop repetitive and long FS descriptions in favor of short names

jpnurmi:
Just updating the English translation which is used as a template for all others. The bot will automatically regenerate the untranslated entries and open a PR. For already translated entries, Weblate will show a warning that they are out of date.

```
$ apt show btrfs-progs 2>/dev/null | grep Desc
Description: Checksumming Copy on Write Filesystem utilities
```

Screenshot from 2023-03-01 18-41-44

Close: #1294

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/allocate_disk_space/allocate_disk_space_dialogs.dart
@@ -136,49 +136,6 @@ Future<void> showCreatePartitionDialog(
   );
 }
 
-extension _PartitionFormatLang on PartitionFormat {
-  String localize(AppLocalizations lang) {
-    switch (this) {
-      case PartitionFormat.none:
-        return lang.partitionFormatNone;
-      case PartitionFormat.btrfs:
-        return lang.partitionFormatBtrfs;
-      case PartitionFormat.ext2:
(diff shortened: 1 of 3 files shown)


jpnurmi:
My personal favorite is the Subiquity way :)

image

@elioqoshi Do you have any thoughts on this? IMO the long descriptions make it almost harder to find what you're looking for. There's so much text that it takes (unnecessary?) extra time to scan through the content.

image

elioqoshi:
@jpnurmi could you elaborate what I should be looking at in comparison? I am not sure what the impact is here.

jpnurmi:
@elioqoshi Here's the current version which already looks somewhat heavy with pretty long descriptions, a proposal that makes the Btrfs description even longer, and a counter-proposal that drops all those descriptions to make it look like Subiquity. :) Yet another option could be to move the description to a second line with a smaller font and weaker color so it becomes less prominent and eats up less space...

| Current | Proposal | Alternative | Yet another |
|---|---|---|---|
| image | image | image | image |

elioqoshi:
Ah yeah, I think the Subiquity way is better, with no descriptions. Descriptions won't help users who don't know what they are doing anyway and users who know what they are looking for have no use for it. So the alternative proposal looks best to me I think

jpnurmi:
Getting back to this while we still can... should we go with the short names?

image

xnox (approved):

Issue #1479, opened 2023-02-28
change "Go back" in buttons to "Back"


elioqoshi:
(similar to https://github.com/canonical/ubuntu-desktop-installer/issues/1476 )

"Go" seems redundant in "Go Back" as the action of the button is self-explanatory. "Back" should suffice and is a standard used across UX copy in various applications.

<img width="166" alt="grafik" src="https://user-images.githubusercontent.com/5436686/221819404-0b19adfb-dfff-4d8d-bbf9-b48bd825b861.png">

jpnurmi:
What would be a nice minimum width for the buttons?

![image](https://user-images.githubusercontent.com/140617/221822157-27a8ade1-f9bd-4c80-a23e-9d2da996814e.png)

Perhaps 120?

![image](https://user-images.githubusercontent.com/140617/221823620-e3e98e41-114a-417f-b22d-2a2c02b800c4.png)
![image](https://user-images.githubusercontent.com/140617/221823919-97d9a8b8-b394-48a7-915a-da437617fd1d.png)
![image](https://user-images.githubusercontent.com/140617/221823955-30ed474e-b4a8-4abb-89b4-5fbe29b5741b.png)
![image](https://user-images.githubusercontent.com/140617/221823984-bff4e923-1041-423f-954c-ae12c9c6232c.png)

elioqoshi:
I wanted to address this in another ticket but the higher minimum width looks good!

jpnurmi:
I got a bit sidetracked because the "Back" button became so small that it started looking awkward. :) We could start with the minimum width and come back to the label changes after we've fixed the width.

jpnurmi:
<img width="166" alt="grafik" src="https://user-images.githubusercontent.com/5436686/221819404-0b19adfb-dfff-4d8d-bbf9-b48bd825b861.png">

is this 268 wide @2x so 134 logical pixels?

elioqoshi:
I think this looks okay. Let's see how this behaves with longer widths. We will likely have to check this in all major languages in the end.

elioqoshi:
Reopening this as we had this challenged in this issue:
https://github.com/canonical/ubuntu-desktop-installer/issues/1585#issuecomment-1469478278

The consensus is to go with `Previous` and `Next`

@jpnurmi pretty please?

PR #1507, opened 2023-03-01
Change "Go back" in buttons to "Back"

jpnurmi:
Use `UbuntuLocalizations.backLabel` from the new [stock labels](https://github.com/canonical/ubuntu-flutter-plugins/pull/63). The continue button needs the same treatment but it's used all over the place so let's leave it to a separate step.

| Before | After |
|---|---|
| image | image |

Fixes: #1479


elioqoshi:
LGTM!

elioqoshi (approved):

Issue #1401, opened 2023-02-14
Username reset on page change


jibel:
Lunar 20230214

The username is reset when the user switches from the location page to the user page but not when it switches from the look & feel page to the user page.

## Test Case
1. Proceed with the installation and go to the user page
2. Fill the fields
3. Continue to next page (Look & Feel)
4. Go back to the user page
=> Verify that the fields still contain the value you entered
5. Go back to the location page
6. Continue to the user page
=> The fields are empty

## Expected behaviour
The content for the fields of the user page should be preserved in all cases.

PR #1517, opened 2023-03-02
ValidatedFormField: allow initial value changes when focused

jpnurmi:
Don't blindly reject initial value changes when focused because most values are initialized asynchronously whenever entering a page. If a field was (auto)focused, it would never get its initial value.

## Before

[Screencast from 2023-03-02 20-16-36.webm](https://user-images.githubusercontent.com/140617/222529200-abd157cd-5533-4c83-8d07-ccf4eda43a10.webm)

## After

[Screencast from 2023-03-02 20-18-39.webm](https://user-images.githubusercontent.com/140617/222529590-7cfd36f3-dbd2-40fd-8041-a755af399b00.webm)

Fixes: #1401


d-loose (approved):
Works :+1:

jpnurmi:
It's not too pretty but I guess we can live with this "for now"

PR #1521, opened 2023-03-03
Pass USERNAME to setInputSource()

jpnurmi:
Subiquity runs as root. It doesn't know the desktop live session user. The GUI must pass the appropriate username to make it possible for Subiquity to change the desktop input source for the live session user. It defaults to 'ubuntu' which won't work for flavors that may have a different live session username.

Ref: canonical/ubuntu-flavor-installer#10


d-loose (approved):
:+1:

PR #1522, opened 2023-03-03
Add "Configure Active Directory" page

jpnurmi:
This is the _Configure Active Directory_ page alone extracted from #1415 without changes to other pages. A _Use Active Directory_ checkbox will be added on the _Who are you_ page as a separate step when things are ready. This PR does not include a way to navigate to the _Configure Active Directory_ page through the GUI but the page can be tested by passing `--initial-route=/active-directory` command-line argument or by changing the initial route in `routes.dart`.

| Light | Dark |
|---|---|
| image | image |

Ref: #41


packages/ubuntu_desktop_installer/lib/pages/active_directory/active_directory_l10n.dart
```diff
@@ -0,0 +1,54 @@
...
+      case AdDomainNameValidation.END_DOT:
+        return lang.activeDirectoryDomainEndDot;
+      case AdDomainNameValidation.START_HYPHEN:
+        return lang.activeDirectoryDomainStartHyphen;
+      case AdDomainNameValidation.END_HYPHEN:
+        return lang.activeDirectoryDomainEndHyphen;
+      case AdDomainNameValidation.MULTIPLE_DOTS:
+        return lang.activeDirectoryDomainMultipleDots;
+      default:
+        throw UnimplementedError(name);
```
jpnurmi, line 27: missing translation for REALM_NOT_FOUND
image

CarlosNihelton: image

versus

image

CarlosNihelton: Honestly: much better than the alternatives.

packages/ubuntu_desktop_installer/lib/pages/active_directory/active_directory_widgets.dart
```diff
@@ -0,0 +1,90 @@
...
+      labelText: lang.activeDirectoryDomainLabel,
+      initialValue: domainName,
+      successWidget: const SuccessIcon(),
+      validator: CallbackValidator(
+        (_) =>
+            validation == null ||
+            validation.singleOrNull == AdDomainNameValidation.OK,
+        errorText: validation?.firstOrNull?.localize(lang) ?? '',
+      ),
+      onChanged: context.read<ActiveDirectoryModel>().setDomainName,
```
CarlosNihelton, line 35: This worries me a little bit because when we set the domain name in the mode, we call for validation, and if the string is valid, then a network call is made to confirm the domain is reachable. It's usually fast in the positive scenario, but can easily span for more than 5 seconds if the domain name is not found. Chances are we won't see that often, because subiquity provides you with domain auto discovery, so a user requesting to join AD might not even need to type the domain name, in such case we should be fine.

CarlosNihelton (approved):
LGTM so far! I realize the mechanism for watching the join result will be a topic for its own PR and I appreciate that.

packages/ubuntu_desktop_installer/lib/pages/active_directory/active_directory_widgets.dart
jpnurmi: Hmm, I'd also feel more confident about this if we could split the basic regexp-based input validation and connection testing to separate endpoints. Or is it already possible with the latest API?

CarlosNihelton: I'll take care of that split in subiquity and ping you back once done.

jpnurmi: I've updated it to ping the domain controller on editing complete or when pressing _Test connection_ 

image

jpnurmi:
Alright, time to take the first step! I'll add the checkbox next.

PR #1534, opened 2023-03-06
WhoAreYouModel: prepare site-wide connectivity check for AD

jpnurmi:
This property will be used to enable/disable the upcoming "Use Active Directory" checkbox.

[`NM_STATE_CONNECTED_SITE`](https://networkmanager.dev/docs/api/latest/nm-dbus-types.html#NMState)
> There is only site-wide IPv4 and/or IPv6 connectivity. This means a default route is available, but the Internet connectivity check (see "Connectivity" property) did not succeed. The graphical shell should indicate limited network connectivity.

Ref: #41

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/who_are_you/who_are_you_model.dart
@@ -1,7 +1,6 @@
 import 'dart:io';
 
 import 'package:flutter/foundation.dart';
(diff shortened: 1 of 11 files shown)


CarlosNihelton (approved):
IIUC you are just paving the way to be able to use the `isConnectedSite` getter later when adding the checkbox to use AD. It makes sense to me. 👍🏽

jpnurmi:
Yes, just a preparation step to reduce some noise from the final AD changes. :)

Issue #1533, opened 2023-03-04
How to hide "choose your look" on non-gnome shell installers?


fossfreedom:
The dark / light theme option is specific to gnome-shell.

For Ubuntu Budgie I don't think we can configure this page via ubuntu-flavor-installer - am I mistaken here?

Is there a way that this page can only be displayed for gnome shell only?

Or alternatively can this page be added to be configurable for flavors?

At the moment selecting light breaks the budgie desktop visually since yaru has little support for budgie desktop.

jpnurmi:
You're right that in the longer term we probably need a way to control which pages are shown for flavors but we could start with a simple fix to make the installer respect the system theme.

The installer clearly shouldn't blindly switch the theme between "Yaru" and "Yaru-dark". It's been like this since the early days of the new installer and simply wasn't paid attention to because it didn't cause issues until now. :)

We could easily make the installer query the current system theme and then add or remove the "-dark" suffix, for example. That way it would switch between "Pocillo" and "Pocillo-dark". Would it do the right thing on Budgie?

fossfreedom:
Hi. Adding a suffix to the system theme would make sense from a  live session short-term point of view.

  I don't have any insight though how that choice carries forward after the installation for the new login that is created and for any other user account that is subsequently created.

jpnurmi:
> I don't have any insight though how that choice carries forward after the installation for the new login that is created and for any other user account that is subsequently created.

This is handled by the post-installation scripts:
https://github.com/UbuntuBudgie/ubuntu-budgie-installer/blob/main/snap/local/postinst.d/10_override_desktop_settings

It might be good to review to scripts to see if there's anything you want/need to customize for Budgie:
https://github.com/UbuntuBudgie/ubuntu-budgie-installer/tree/main/snap/local/postinst.d

jpnurmi:
The next round of updates should help:
- https://github.com/UbuntuBudgie/ubuntu-budgie-installer/pull/4

jpnurmi:
There's currently no easy way to hide the screen but the default desktop settings service no longer hardcodes Yaru _and_ there's a way to pre-register a custom implementation of the desktop service:

```dart
abstract class DesktopService {
  Future<void> inhibit();
  Future<void> setTheme(Brightness brightness);
  Future<void> close();
}
```
For example:
```dart
void main() {
  registerService<DesktopService>(MyDesktopService.new);
  runInstallerApp(...);
}
```

PR #1535, opened 2023-03-06
Share InheritedLocale via ubuntu_wizard

jpnurmi:
In #897, we added `InheritedLocale` to `ubuntu-wsl-setup` to avoid using `Settings` because the `gsettings.dart` dependency was not desired when running the WSL setup in Windows. This PR moves the `InheritedLocale` implementation into the shared `ubuntu_wizard` library and takes it into use in `ubuntu_desktop_installer`, too.

The old `Settings` class will be left with such desktop settings that flavors want/need/should customize. It will be made possible to pass a custom settings backend via the `runInstallerApp()` entry-point later, and perhaps renamed to `DesktopSettings`...?

Ref: #1533


d-loose (approved):
LGTM :+1:

Issue #1476, opened 2023-02-28
change "Start installation" button label to "Install"


elioqoshi:
The long label of "Start installation" is redundant and also technically not the complete story. It will install Ubuntu, not just start it. We don't use buttons with labels like "Start deleting" either so we should get rid of the `Start ` here.

Frontloading the keyword to the beginning of a label is also more useful, as "Install" can be easier recognized visually.

<img width="849" alt="grafik" src="https://user-images.githubusercontent.com/5436686/221801391-7b375831-a092-4013-95a4-6268746fa002.png">

Beyond that, it will also help with the consistency of the button label length, especially in longer languages like German or Danish.

PR #1537, opened 2023-03-06
Write changes to disk: change "Start installing" to "Install"

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2023-03-06 10-07-20 | Screenshot from 2023-03-06 10-08-02 |

Close: #1476


elioqoshi (approved):
`Start installing` is dead. Long live `Install`

d-loose (approved):
:+1:

Issue #1500, opened 2023-03-01
Screen size detection is unreliable


jpnurmi:
If the screen size (with some margins reduced) is considered too small to fit the fixed 960x680 window, it's shown full screen. The detection is not reliable as can be seen in the screenshot below. The resolution is clearly much higher than what is needed to fit a 960x680 window but it was still shown full screen.

![image](https://user-images.githubusercontent.com/140617/222216867-fc186985-fd4e-4231-906f-f8131f4afc49.png)

PR #1539, opened 2023-03-06
Revise workarea detection & fix window size comparison

jpnurmi:
First of all, use the default window size (960x680) instead of the size allocation including shadows (1012x732) for determining whether it fits on the screen.

Secondly, given that gdk_monitor_get_workarea() is not reliable early on startup and we have to subtract some arbitrary margins, make sure to subtract the margins from the full geometry instead of the workarea to avoid subtracting double if and when the workarea is correctly reported when restarting the installer after startup.

Fixes: #1500


d-loose (approved):
LGTM :+1:

PR #1542, opened 2023-03-06
Flavor compatible desktop settings

jpnurmi:
Following up on what was promised in #1535. :)

Make the desktop settings a plain service because it's no longer notifying anything now that the app locale was moved out, give it a bit more explicit name, and let flavors pass their own into `runInstallerApp()`.

The default implementation for GNOME has been also generalized a tiny bit to make it able to switch between `Foo` and `Foo-dark` themes where `Foo` can be anything instead of being hardcoded to Yaru. :)

Ref: #1533


jpnurmi:
Hmm, I've got an idea. Instead of adding more and more `runInstallerApp()` arguments, perhaps we could let flavors pre-register services before calling the method. This could be more flexible and powerful...

```dart
import 'package:ubuntu_desktop_installer/installer.dart';

import 'budgie_slides.dart';

Future<void> main(List<String> args) {
  return runInstallerApp(
    args,
    flavor: FlavorData(
      name: 'Ubuntu Flavor',
      theme: yaruFooLight,
      darkTheme: yaruFooDark,
      localizationsDelegates: AppLocalizations.localizationsDelegates,
    ),
    slides: [...],
    settings: BudgieSettings(), // <==
  );
}
```
vs.
```dart
import 'package:ubuntu_desktop_installer/installer.dart';
import 'package:ubuntu_service/ubuntu_service.dart';

import 'budgie_slides.dart';
import 'budgie_settings.dart';

Future<void> main(List<String> args) {
  registerService(BudgieSettings.new);  // <==

  return runInstallerApp(
    args,
    flavor: FlavorData(
      name: 'Ubuntu Flavor',
      theme: yaruFooLight,
      darkTheme: yaruFooDark,
      localizationsDelegates: AppLocalizations.localizationsDelegates,
    ),
    slides: [...],
  );
}
```

PR #1547, opened 2023-03-07
Add --source-catalog command-line argument for testing purposes

jpnurmi:
Forwarded to subiquity in dry-run mode, similarly to `--machine-config`. Defaults to [`example/mixed-sources.yaml`](https://github.com/canonical/subiquity/blob/main/examples/mixed-sources.yaml) which comes with a desktop source.


d-loose (approved):
:+1:

PR #1548, opened 2023-03-07
Switch to conditional service registration if not already registered

jpnurmi:
This allows flavors and tests to register their own preferred service implementations before calling `runInstallerApp()`.

Ref: #1533
Ref: #1542


d-loose (approved):
LGTM :+1:

Issue #1550, opened 2023-03-07
Language choosing gets wrong


bittin:
Is it known that the installer sometimes decides to choose English even if i choose Swedish? 
some screens in it sometimes randomly changes back to English

Logs:

https://pastebin.com/ewcSDYNH 
https://pastebin.com/XXabeMWW

jpnurmi:
Thanks, I was able to reproduce the issue. I had Swedish selected when entering the timezone page.

![image](https://user-images.githubusercontent.com/140617/223417571-8460e121-7dcb-47ab-8e39-a01e15888699.png)

PR #1551, opened 2023-03-07
Move `InheritedLocale` up in the tree to avoid losing language selection

jpnurmi:
This fixes a regression introduced by #1535. The app object is rebuilt whenever Subiquity's status changes, so we cannot store InheritedLocale there. :(

Fix: #1550


d-loose (approved):
LGTM :+1:

PR #1554, opened 2023-03-07
Disable screensaver, screen blanking, automounting

### LINKED ISSUE #1543: Disable screensaver, screen blanking, power manager?
Ubiquity:
```py
        self.disable_volume_manager()
        self.disable_screensaver()
        self.disable_screen_blanking()
        self.disable_powermgr()
```
https://git.launchpad.net/ubiquity/tree/ubiquity/frontend/gtk_ui.py#n842

d-loose:
Disables the screensaver, screen blanking and automatic volume mounting when the installer is run, and restores the previous settings when its closed.

I mostly followed https://git.launchpad.net/ubiquity/tree/ubiquity/frontend/gtk_ui.py#n842 with two exceptions:
- I didn't implement `disablePowermgr` yet. In ubiquity this is done by setting `active=false` in `org.gnome.settings-daemon.plugins.power`, however gsettings doesn't list an `active` key for that interface on my system (22.10).
- `org.gnome.nautilus.desktop` doesn't exist on my system either, so I left that part out of `disableAutoMount`

I'll do more research on those things tomorrow, but please feel free to enlighten me if anyone knows more about this.

Ref #1543
Edit: I'll tackle power management in a separate PR.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -32,7 +32,7 @@ jobs:
         run: |
           sudo apt update
           sudo apt install -y clang cmake curl libgtk-3-dev ninja-build pkg-config unzip xvfb
-          sudo apt install -y dbus-x11 network-manager upower
+          sudo apt install -y dbus-x11 network-manager ubuntu-desktop-minimal upower
           make install_deps
 
       - name: Install Flutter
--- a/packages/ubuntu_desktop_installer/lib/services/desktop_service.dart
+++ b/packages/ubuntu_desktop_installer/lib/services/desktop_service.dart
@@ -29,11 +29,14 @@ class GnomeService implements DesktopService {
   /// Creates a GNOME settings instance using the given GSettings as a backend
   /// for storing the settings.
   GnomeService({
+    @visibleForTesting GSettings? dingSettings,
     @visibleForTesting GSettings? interfaceSettings,
     @visibleForTesting GSettings? mediaHandlingSettings,
     @visibleForTesting GSettings? sessionSettings,
     @visibleForTesting GSettings? screensaverSettings,
-  })  : _interfaceSettings =
+  })  : _dingSettings =
+            dingSettings ?? GSettings('org.gnome.shell.extensions.ding'),
+        _interfaceSettings =
             interfaceSettings ?? GSettings('org.gnome.desktop.interface'),
         _mediaHandlingSettings = mediaHandlingSettings ??
             GSettings('org.gnome.desktop.media-handling'),
@@ -42,6 +45,7 @@ class GnomeService implements DesktopService {
         _sessionSettings =
             sessionSettings ?? GSettings('org.gnome.desktop.session');
 
+  final GSettings _dingSettings;
   final GSettings _interfaceSettings;
   final GSettings _mediaHandlingSettings;
   final GSettings _screensaverSettings;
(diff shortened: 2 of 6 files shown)


jpnurmi:
looking great so far! feel free to slit off the finished/working screen saver and blanking parts we could already merge

jpnurmi:
it just came to my mind that we have to double-check that the [post-inst script](https://github.com/canonical/ubuntu-desktop-installer/blob/main/snap/local/postinst.d/10_override_desktop_settings) doesn't transfer these settings to the target system :slightly_smiling_face:

d-loose:
It seems like the settings I mentioned are outdated:
- the ability to start/stop individual gnome-settings plugins (with the `active` keys) was removed [here](https://github.com/GNOME/gnome-settings-daemon/commit/caf51f50f332aabe7af77f9641ac288dd88af553)
- `org.nautilus.desktop` was removed when [nautilus dropped desktop support](https://gitlab.gnome.org/GNOME/nautilus/-/merge_requests/46)
@sergio-costas should we instead set `org.gnome.shell.extensions.ding show-volumes` to `false` here?

sergio-costas:
@d-loose About ding, "show-volumes" only determines whether mounted external drives are shown or not. To fully disable the desktop, you should remove "ding" from the list of enabled extensions: /org/gnome/shell/extensions/enabled-extensions.

Also, there is a disabled-extensions entry; I have "ding" set there because I'm using the upstream "ding", which has to have a different id when installed in Ubuntu.

sergio-costas:
BTW: as I already commented to J-P, currently, to disable the blank screen the dconf key is /org/gnome/desktop/session/idle-delay , setting it to zero.

d-loose:
> About ding, "show-volumes" only determines whether mounted external drives are shown or not. To fully disable the desktop, you should remove "ding" from the list of enabled extensions: /org/gnome/shell/extensions/enabled-extensions.

I think it's sufficient to just hide the drive icons, to make sure the user doesn't accidentally mount anything during the installation (at least I think that was the intention of https://git.launchpad.net/ubiquity/tree/ubiquity/frontend/gtk_ui.py#n743). WDYT?

d-loose:
> it just came to my mind that we have to double-check that the [post-inst script](https://github.com/canonical/ubuntu-desktop-installer/blob/main/snap/local/postinst.d/10_override_desktop_settings) doesn't transfer these settings to the target system slightly_smiling_face

Just to document somewhere that we have discovered they aren't transferred: https://github.com/canonical/ubuntu-desktop-installer/blob/main/snap/local/postinst.d/10_override_desktop_settings#L23-L27 doesn't copy any of the settings involved in this PR :)

sergio-costas:
> I think it's sufficient to just hide the drive icons, to make sure the user doesn't accidentally mount anything during the installation (at least I think that was the intention of https://git.launchpad.net/ubiquity/tree/ubiquity/frontend/gtk_ui.py#n743). WDYT?

Good point... if before it only disabled the volumes, then I fully agree.

sergio-costas:
But... what about remote mounts? Those are managed in a different setting...

d-loose:
> But... what about remote mounts? Those are managed in a different setting...

They shouldn't interfere with partitioning, but we can hide them as well, just to be safe.

.github/workflows/integration_test.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
         run: |
           sudo apt update
           sudo apt install -y clang cmake curl libgtk-3-dev ninja-build pkg-config unzip xvfb
-          sudo apt install -y dbus-x11 network-manager upower
+          sudo apt install -y dbus-x11 network-manager ubuntu-desktop upower
```
jpnurmi, line 35: i don't know how much it saves but did you give it a try with ubuntu-desktop-minimal?

packages/ubuntu_desktop_installer/lib/services/desktop_service.dart
```diff
@@ -15,29 +28,106 @@ abstract class DesktopService {
...
+    final previousShowNetworkVolumes =
+        await _dingSettings.get('show-network-volumes');
+    await _mediaHandlingSettings.set('automount', const DBusBoolean(false));
+    await _mediaHandlingSettings.set(
+        'automount-open', const DBusBoolean(false));
+    await _mediaHandlingSettings.set('autorun-never', const DBusBoolean(true));
+    await _dingSettings.set('show-volumes', const DBusBoolean(false));
+    await _dingSettings.set('show-network-volumes', const DBusBoolean(false));
+    restoreSettings.add(() async {
+      await _mediaHandlingSettings.set('automount', previousAutoMount);
```
jpnurmi, line 74: would reset() work or are these explicitly set in the beginning?

jpnurmi (commented):
very cool! thanks for working on this. :) i left a couple of questions

.github/workflows/integration_test.yaml
d-loose: I'll check - if it still takes so long we can just add gnome-shell-extension-desktop-icons-ng

packages/ubuntu_desktop_installer/lib/services/desktop_service.dart
d-loose: I just mimicked what ubiquity is doing, but I guess it's a little overkill. Does anyone fiddle with screensaver and automounting settings in a live system, run the installer, and then decide to continue using the live system, expecting all their precious settings are restored?

jpnurmi (approved):
Thanks! :tada: LGTM :+1:

PR #1555, opened 2023-03-08
Allocate disk space: present "leave unformatted" as "none" format

jpnurmi:
Previously, we presented the "Leave unformatted" option as `null`. However, we want to have an additional separator which will be supported by the upcoming `MenuButtonBuilder` in `ubuntu_widgets`. The separator will be presented as `null` so let's present the "Leave unformatted" option with its own dedicated value instead, to leave `null` for the separator.

NOTE: I just added the Subiquity screenshot for reference and noticed that it actually has two separators but it's irrelevant for this PR because this does not change the visuals. This is just a preparation step for being able to present both separators (`null`) and the "Leave unformatted" option (`PartitionFormat.none`).

| Current | Future | Subiquity|
|---|---|---|
| Screenshot from 2023-03-08 07-40-55 | Screenshot from 2023-03-08 07-37-50 | image |

See also:
- #1335
- canonical/ubuntu-flutter-plugins#148


d-loose (approved):
LGTM :+1:

Issue #1563, opened 2023-03-08
Replace DropdownBuilder with MenuButtonBuilder


jpnurmi:
We should replace all `DropdownBuilder` instances with `MenuButtonBuilder` from `ubuntu-flutter-plugins`.

- Based on the new overlay-based [MenuAnchor](https://api.flutter.dev/flutter/material/MenuAnchor-class.html) which does not have hard-coded menu/dropdown route transitions that look and feel off on desktop.
- Unlike the old non-themable `DropdownButton`, the new `MenuAnchor` + `MenuItemButton` are themable. They look like Yaru popups, and it's possible to change the mouse cursor.

## Before

[Screencast from 2023-03-08 21-33-08.webm](https://user-images.githubusercontent.com/140617/223843209-22c9b34d-f236-4e1e-ad42-35a9f9d11cdb.webm)

## After

[Screencast from 2023-03-08 21-31-58.webm](https://user-images.githubusercontent.com/140617/223843006-25801b7a-6cd3-4746-b1ae-0de2d103c0c4.webm)

See also:
- https://github.com/canonical/ubuntu-flutter-plugins/pull/148
- https://github.com/ubuntu/yaru_widgets.dart/issues/557
- https://github.com/canonical/ubuntu-desktop-installer/issues/1532

PR #1568, opened 2023-03-09
StorageSizeBox: replace DropdownBuilder with MenuButtonBuilder

jpnurmi:
Looks like a Yaru menu, doesn't have hard-coded mobile-oriented transitions, and allows customizing the mouse cursor later.

| Before | After |
|---|---|
| image | image |
| image | image |
| image | image |
| image | image |

Ref: #1563


jpnurmi:
@elioqoshi This may look like a nice-to-have at first sight but it's actually an important step that allows us to take over the look'n'feel of these dropdowns. This PR changes only one dropdown but the idea is to exchange them all, of course. :)

elioqoshi:
Thanks for the context, yeah I see. This is a no brainer then, looks good!

elioqoshi (approved):

d-loose (approved):
:+1:

Issue #1570, opened 2023-03-09
Contrast accessibility issues


Horlix:
When selecting High Contrast mode in the accessibility options the installer window ignores the contrast change and this remains the same.

jpnurmi:
Reopening this because the issue is not fully fixed yet. The app is now respecting the system setting but we still need to fix high-contrast Yaru themes. Some relevant tickets:
- https://github.com/ubuntu/yaru.dart/issues/305
- https://github.com/ubuntu/yaru.dart/issues/302

jpnurmi:
There's some progress. Dennis helped to fix the borders to stand out in HC themes:

- https://github.com/ubuntu/yaru.dart/pull/319

jpnurmi:
A few remaining issues with high-contrast themes:

1. Selected list items don't stand out

| Normal | High-contrast |
|---|---|
| ![image](https://user-images.githubusercontent.com/140617/234800884-2497d501-e5e2-4af6-a05f-acde865a4b40.png) | ![image](https://user-images.githubusercontent.com/140617/234800824-53433c94-908b-458b-acc4-040087ca1ec6.png) |

2. The labels of disabled buttons are weak

| Normal | High-contrast |
|---|---|
| ![image](https://user-images.githubusercontent.com/140617/234801297-a7676177-c590-4c7d-b4ca-daae6ab27a43.png) | ![image](https://user-images.githubusercontent.com/140617/234801344-0b74ba75-e65b-47b8-99bc-fdcc68a309df.png) |

3. Theme selection highlight is unclear

| Normal | High-contrast |
|---|---|
| ![image](https://user-images.githubusercontent.com/140617/234801859-54008a10-3c2e-4a4e-ac21-ea56f3663866.png) | ![image](https://user-images.githubusercontent.com/140617/234801799-5d823013-94f7-4a95-b83c-bdb75cfc81fa.png) |

PR #1573, opened 2023-03-09
Specify high contrast themes

jpnurmi:
Ref: #1570

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/installer.dart
+++ b/packages/ubuntu_desktop_installer/lib/installer.dart
@@ -225,6 +225,8 @@ class _UbuntuDesktopInstallerAppState extends State<UbuntuDesktopInstallerApp> {
                 },
                 theme: theme?.customize(),
                 darkTheme: darkTheme?.customize(),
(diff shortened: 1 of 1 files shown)


jpnurmi:
Yaru's high-contrast themes don't look correct anymore:
| Light | Dark |
|---|---|
| image | image |

- https://github.com/ubuntu/yaru.dart/pull/148

jpnurmi:
I was mostly concerned about the dark background that has changed from black to "jet" but it's the same with GCC so perhaps it's fine after all?

image

elioqoshi:
This seems okay to me. But for what's worth, there is a lot of things which are not great for high color contrast upstream in Yaru, but that's a separate thing.

jpnurmi:
We'll fix the high-contrast themes in yaru.dart. Some relevant tickets:
- https://github.com/ubuntu/yaru.dart/issues/305
- https://github.com/ubuntu/yaru.dart/issues/302

d-loose (approved):
:+1:

PR #1581, opened 2023-03-10
Active directory join result

jpnurmi:
image

Ref: #41

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/active_directory/active_directory_page.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/active_directory/active_directory_page.dart
@@ -7,6 +7,7 @@ import 'package:yaru_widgets/yaru_widgets.dart';
 
 import '../../l10n.dart';
 import '../../services.dart';
+import 'active_directory_dialogs.dart';
 import 'active_directory_model.dart';
 import 'active_directory_widgets.dart';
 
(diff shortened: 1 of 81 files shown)


packages/ubuntu_desktop_installer/lib/pages/active_directory/active_directory_page.dart
```diff
@@ -69,7 +70,15 @@ class _ActiveDirectoryPageState extends State<ActiveDirectoryPage> {
         WizardAction.next(
           context,
           enabled: context.select((ActiveDirectoryModel m) => m.isValid),
-          onNext: context.read<ActiveDirectoryModel>().save,
+          onNext: () async {
+            final model = context.read<ActiveDirectoryModel>();
+            await model.save();
+            model.getJoinResult().then((result) {
+              if (mounted && result == AdJoinResult.JOIN_ERROR) {
```
CarlosNihelton, line 77:
```suggestion
               if (mounted && (result == AdJoinResult.JOIN_ERROR || result == AdJoinResult.PAM_ERROR)) {
 ```
 
 If PAM_ERROR is returned is likely that the user won't find his home directory when logging with her AD account, which is not a good thing. Inspecting ubiquity code it seems to display the same error message if that error happens.

I personally never seen it happen but I guess we should warn about it. Maybe in the future we may decide to show a different set of instructions, but for now, for simplicity, I'd just show the same dialog.

CarlosNihelton (changes requested):
One small thing and we should be good to go!

packages/ubuntu_desktop_installer/lib/pages/active_directory/active_directory_page.dart
jpnurmi: awesome, thanks! :+1:

CarlosNihelton (approved):
LGTM once melos is pinned to version 2.9

PR #1604, opened 2023-03-13
CI: speed up the integration test workflow

jpnurmi:
Split the jobs because they have different needs. For example, only the desktop installer test needs ubuntu-desktop-minimal, and the slide test doesn't really need subiquity dependencies at all.


jpnurmi:
It's a bit more repeat but we barely need to touch these so probably doesn't matter much...?

d-loose (approved):
:+1:

PR #1605, opened 2023-03-13
Page indicator: increase relative size of selected dot

d-loose:
| Before | After |
| - | - |
| image | image |


jpnurmi (approved):
That was quick. :) LGTM :+1:

Issue #1502, opened 2023-03-01
Partition dialog size constraints


jpnurmi:
![image](https://user-images.githubusercontent.com/140617/222221474-1308aa21-1924-4610-8c71-1d6b38af99aa.png)

jpnurmi:
@elioqoshi is 600px a reasonable width for this kind of dialog or does it look too stretched when it's empty?

![image](https://user-images.githubusercontent.com/140617/223186447-0c3b1ae4-4766-43bf-bfd4-dd1cd068e6d8.png)

elioqoshi:
@jpnurmi Yeah, growing the dialog dynamically is really awkward.

How about a much bigger width but not confine all the elements inside to the same width?

![grafik](https://user-images.githubusercontent.com/5436686/223363379-790cc38e-4428-4e9c-88a6-4e746a964df5.png)

PR #1611, opened 2023-03-13
Partition dialog size constraints

jpnurmi:
The storage size spinbox is not expanded to full width as proposed in https://github.com/canonical/ubuntu-desktop-installer/issues/1502#issuecomment-1457737146.

| Before | After |
|---|---|
| image | image |

Close: #1502


d-loose (approved):
:+1:

PR #1620, opened 2023-03-14
Use lzo compression

kenvandine:


seb128:
Did we mesure the impact on storage vs startup?

kenvandine:
Adds 40M of size

Issue #1618, opened 2023-03-14
Updated layout for language selection and keyboard layout


elioqoshi:
I updated the layout for the loading screen, language selection and keyboard layout.

Loading screen:
![grafik](https://user-images.githubusercontent.com/5436686/224988696-d14b5e6d-b8b4-4d0f-95bb-f458b366dd7c.png)

Language selection:
![grafik](https://user-images.githubusercontent.com/5436686/224988908-389e7987-aec9-4647-b6b5-733d73a8522b.png)

Keyboard layout:
![grafik](https://user-images.githubusercontent.com/5436686/224988979-590896f6-95f3-41a5-a7df-25acde4aae18.png)

This should also address the keyboard accessibility issues addressed

jpnurmi:
Could we have the same width for the language and keyboard layout lists? It could look jumpy with different sizes when animations (page transitions) are disabled. GNOME defaults to no animations when running with software rendering, or they can be explicitly disabled in the accessibility settings.

elioqoshi:
@jpnurmi This looks indeed better, thanks!

tim-hm:
A question/suggestion for the future and only if we have data to back it up. When someone selects their language how often do they then go on to change their keyboard layout and variant? 

For example, every time i select English (UK) I accept the default keyboard layout + variant. I'd be curious to understand what percentage of users are similar. If its the majority then we could consider a UX where the keyboard layout + variant are not a separate step from the initial language selection but rather a single button on that step for manual layout + variant selection.

elioqoshi:
@tim-hm Great minds think alike :)

I asked this yesterday as I was designing this based on assumptions as we have neither data on this, neither time to do usability testing with humans. My assumption is similar to yours but that was challenged by @seb128 and @jibel (do FR keyboards use alternate layouts more often? it might depends on the local context, which is not English-specific probably).

We currently do not collect telemetry on keyboard layouts via ubuntu-report so we cannot answer that. I requested that this be tracked (anonymously of course): https://github.com/ubuntu/ubuntu-report/issues/43

Maybe we will get tangible data for 23.10.

PR #1621, opened 2023-03-14
Keyboard layout

jpnurmi:
| Before | After |
|---|---|
| image | image |

Ref: #1618


packages/ubuntu_desktop_installer/test/keyboard_layout/keyboard_layout_page_test.dart
```diff
@@ -85,12 +85,14 @@ void main() {
...
-      await tester.tap(variantTile);
+      await tester.tap(find.byType(MenuButtonBuilder<int>));
+      await tester.pumpAndSettle();
+      final variantItem = find.widgetWithText(MenuItemButton, 'Variant $i');
+      expect(variantItem, findsOneWidget);
+      await tester.tap(variantItem);
       verify(model.selectVariant(i)).called(1);
     }
-  });
+  }, skip: true); // TODO: fix the test
```
jpnurmi, line 95: i'm having trouble figuring out why `tap(MenuButtonBuilder)` doesn't work :(

```

Warning: A call to tap() with finder "exactly one widget with type "MenuButtonBuilder<int>" (ignoring offstage widgets): MenuButtonBuilder<int>(dependencies: [Directionality, MediaQuery, _InheritedTheme, _LocalizationsScope-[GlobalKey#79b53]], state: _MenuButtonBuilderState<int>#71692)" derived an Offset (Offset(609.0, 451.0)) that would not hit test on the specified widget.
Maybe the widget is actually off-screen, or another widget is obscuring it, or the widget cannot receive pointer events.
```

d-loose: it works when I only test the very first variant without a loop - I suspect the issue is that the test tries to tap the `MenuButtonBuilder<int>` again, while the popup is visible

d-loose: Moving
```dart
await tester.tap(find.byType(MenuButtonBuilder<int>));
await tester.pumpAndSettle();
```
out of the loop works

jpnurmi: ahh, i'm an idiot :D thank you!

d-loose (approved):
Ah, that's the better solution of course!

LGTM :+1:

jpnurmi:
it's funny that in the test it's indeed possible to click through all items before the popup closes itself (next frame?)

PR #1622, opened 2023-03-15
Active Directory: add missing page indicator

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2023-03-15 07-38-46 | Screenshot from 2023-03-15 07-39-05 |


d-loose (approved):
LGTM :+1:

PR #1624, opened 2023-03-15
Turn off XXX: remove page indicators

jpnurmi:
When the user is taken to the page with instructions to turn off RST or BitLocker, there's nowhere to progress but the only available action is to restart the computer.

| Before | After |
|---|---|
| Screenshot from 2023-03-15 07-43-40 | Screenshot from 2023-03-15 08-20-39 |
| Screenshot from 2023-03-15 07-45-54 | Screenshot from 2023-03-15 08-20-59 |


elioqoshi (approved):
LGTM!

Issue #1631, opened 2023-03-15
Application-wide minimum button width


jpnurmi:
We applied a minimum width of 136px for all wizard buttons:
- https://github.com/canonical/ubuntu-desktop-installer/pull/1487

We could probably apply it in the application-wide theme for all other buttons too. For example, the Ok and Cancel buttons in dialogs should be wider:

![image](https://user-images.githubusercontent.com/140617/225291175-d809f4b3-ee16-4e8c-818e-ff20b4755d47.png)

![image](https://user-images.githubusercontent.com/140617/225641473-f77e4585-403f-450a-b898-57c83e0738f9.png)

PR #1643, opened 2023-03-15
Apply application-wide minimum button size

jpnurmi:
Previously, we specified 136px as a minimum width for the wizard buttons. Some buttons, most notably the Ok and Cancel buttons in dialogs, are still awkwardly small. This PR promotes 136px as a minimum width for all push buttons (elevated, filled, and outlined) in the whole app.

| Before | After |
|---|---|
| Screenshot from 2023-03-15 20-04-06 | Screenshot from 2023-03-15 20-03-55 |

Close: #1631


jpnurmi:
Hmm, not good... There are several exceptions. Perhaps we need to go case by case.

Issue #1591, opened 2023-03-10
Review page titles


elioqoshi:
The current page titles are off quite often. I did an audit of them and provided suggestions. These are not perfect but should improve all the titles. Ideally this needs to be reviewed by a proper copy editor.

| Current Title                                      | Proposed Title                              | Notes                                                                                             |
| -------------------------------------------------- | ------------------------------------------- | ------------------------------------------------------------------------------------------------- |
| Welcome                                            | Welcome to Ubuntu                           | mention Ubuntu explicitly                                                                         |
| Try or install                                     | Try or install Ubuntu                       | mention Ubuntu explicitly                                                                         |
| Keyboard layout                                    | Keyboard layout                             |                                                                                                   |
| Detect Keyboard Layout                             | Detect Keyboard Layout                      |                                                                                                   |
| Connect to internet                                | Connect to a network                        | Connect to internet sounds awkward. Network is mostly used                                        |
| Updates and other software                         | Install additional packages                 |                                                                                                   |
| Installation type                                  | Type of installation                        | worth to frontload the keyword here, which is 'type'                                              |
| Advanced features                                  | Advanced features                           | This will be legacy soon so keeping it as it is                                                   |
| Choose a security key                              | Security Key                                |                                                                                                   |
| Install Ubuntu Lunar Lobster alongside Windows 11  | Install alongside Windows 11                | remove the Ubuntu version to make it shorter                                                      |
| Install Ubuntu Lunar Lobster alongside them        | Install alongside current operating systems | this is wrong in this context, needs to explain better what it is about                           |
| Write changes to disk                              | Ready to install                            | Emphasize that this is the last step to confirm                                                   |
| Turn off Bitlocker                                 | Bitlocker is enabled                        | Change wording to status (adjective) instead of action (verb()                                    |
| Turn off RST                                       | RST is enabled                              |                                                                                                   |
| Where are you?                                     | Select your time zone                       |                                                                                                   |
| Who are you?                                       | Set up your account                         |                                                                                                   |
| Choose your look                                   | Choose your theme                           | look' is too off tone-wsie                                                                        |
| Welcome to Ubuntu                                  | Ubuntu 23.04                   | This is for the slideshow.                                                                        |
| Installation complete                              | Installation complete                       |                                                                                                   |
| Allocate disk space                                | Manual partitioning                         | Specifying what this really is                                                                    |
| Create partition                                   | Create partition                            |                                                                                                   |
[...]

PR #1649, opened 2023-03-15
Allocate disk space: change page title to "Manual partitioning"

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2023-03-15 21-54-08 | Screenshot from 2023-03-15 21-54-42 |

Ref: #1591


elioqoshi (approved):

seb128 (approved):
Ack, let's land that and I will notify translators

PR #1653, opened 2023-03-16
Refactor product info service

jpnurmi:
The old "ProductInfoExtractor" with static data was painful to work with and test. We need the product info on more pages so this change paves the road by making the product info available via an injectable service.


d-loose (approved):
LGTM :+1:

Issue #1405, opened 2023-02-14
Red error labels look bad in dark theme


jpnurmi:
![image](https://user-images.githubusercontent.com/140617/218073537-715b74fe-073a-4aee-a90d-99b664c8811e.png)

_Originally posted by @jpnurmi in https://github.com/canonical/ubuntu-desktop-installer/issues/1390#issuecomment-1425674837_

PR #1655, opened 2023-03-16
Tweak error colors and text field borders

jpnurmi:
- Remove accent color from focused text field border (orange accent vs. red error are too close)
- Use a lighter red shade for errors in the dark theme to increase contrast and make error labels readable

| Before | After |
|---|---|
| image | image |
| image | image |
| image | image |
| image | image |

Close: #1405


elioqoshi (approved):
This will be a good improvement for now, thanks!

PR #1657, opened 2023-03-16
Filled "show password" button

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2023-03-16 12-56-16 | Screenshot from 2023-03-16 12-56-27 |
| Screenshot from 2023-03-16 12-56-11 | Screenshot from 2023-03-16 12-56-31 |


elioqoshi:
Not entirely sure this will be the final solution long-term but it is an improvement so +1

elioqoshi (approved):

jpnurmi:
we can try other fill colors if you want?

PR #1658, opened 2023-03-16
Help & support slide: increase padding

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2023-03-16 13-32-36 | image |


elioqoshi:
Much better, thanks!

PR #1662, opened 2023-03-16
Allocate disk space: change dialog labels + text

### LINKED ISSUE #1325: New partition table confirmation
@elioqoshi Are _Go Back_ and _Continue_ the best choice of responses for a confirmation dialog like this?

To me, they feel a bit confusing because the same _Go Back_ and _Continue_ actions, which are also visible on the page below, are used on almost every page to navigate the installation wizard. The semantics of these two buttons in the confirmation dialog is completely different, though. These are about reseting the partition configuration on the current page, not about navigating. I was going to fix it and change the buttons to either _Ok_ and _Cancel_, or _Yes_ and _No_ but I noticed it had been copied from Ubiquity. What shall we do?

| Flutterstaller | Ubiquity |
|---|---|
| image | image |

d-loose:
- Change 'Go Back'/'Continue' to 'Cancel'/'OK'.

This is the only dialog left using different labels.

@elioqoshi WDYT?

image

Close #1325


elioqoshi:
I just meant to do changes to this dialog, give me 30min! :D

d-loose:
Ah sure, no rush :)

jpnurmi:
@d-loose don't mind the button sizes and styles. we can fix them in the next step. :)

- #1631

jpnurmi (approved):
Implementation LGTM :+1:

jpnurmi:
@elioqoshi is this good to go? it only changes the text and button labels. we'll fix the button sizes and styles in a separate PR

PR #1664, opened 2023-03-16
Revisit dialog buttons

jpnurmi:
- Use an outlined button for _Cancel_
- Use a filled button for _OK_/_Yes_/_No_
- Give them them same 136px width as the wizard buttons

| Before | After |
|---|---|
| image | image |
| image | image |
| image | image |
| image | image |

Close: #1631


jpnurmi:
@elioqoshi Wizard buttons are 136px but it felt a bit too much for these dialog buttons so I went down to 120px. We could perhaps make it even smaller, or what do you think?

d-loose:
Do we need an 'Outlined' version as well?

jpnurmi:
There are some annoyingly small buttons outside dialogs too so I renamed it to `PushButton` and added `elevated` `filled` and `outlined` variants.

elioqoshi (approved):

Issue #1577, opened 2023-03-09
Inconsistent disk and partition name formatting


jpnurmi:
"sda1", "disk-sda1", "/dev/sda1", "#sda1", ...

![image](https://user-images.githubusercontent.com/140617/224047747-fa8aba05-5ee3-4d9f-b62c-dcdbd3f81bb2.png)
![image](https://user-images.githubusercontent.com/140617/224047844-c31dc564-18b8-446a-ab16-f1e526c5d339.png)
![image](https://user-images.githubusercontent.com/140617/224047903-e066990a-3adb-417c-a352-93a7047e7545.png)

jpnurmi:
Oops, "formatting" sounds confusing in this context. I mean the names, not erasing. :)

elioqoshi:
I'd double check with @local-optimum or @jibel but I feel `/dev/sda1` might be best to stick to

elioqoshi:
@jibel mentioned it will always be `dev` so we might want to drop that and go with `sda1` as a format. Unless we are missing something

elioqoshi:
Quoting @jibel :

> If you look at the kernel or systemd messages, they always use the short form (without /dev/) IMHO if you're referring to the device or partition, use the short form. If you're referring to the mount point use the full path.

I think we should go for this.

elioqoshi:
@jpnurmi did we address this in a PR?

jpnurmi:
Should we use everywhere bold **sda** and **sda1** for disks and partitions, respectively?

elioqoshi:
Yes, everywhere except the mount point which has the full name

jpnurmi:
We missed some

![image](https://user-images.githubusercontent.com/140617/225922371-ccc82ad5-399d-4245-8d72-0aebc9d47ef7.png)

PR #1667, opened 2023-03-16
Manual partitioning: use simplified disk and partition names

jpnurmi:
| Before | After |
|---|---|
| image | Screenshot from 2023-03-16 20-07-36 |

Ref: #1577


elioqoshi (approved):
`designer/approved` is dead
long live `/approved`

PR #1674, opened 2023-03-17
Added enable-patchelf in subiquitydeps and os-prober parts

kenvandine:
Added enable-patchelf in subiquitydeps and os-prober parts, where stage-packages are being added


jpnurmi (approved):
LGTM :+1:

PR #1682, opened 2023-03-20
Write changes to disk: fix inconsistent partition formatting

jpnurmi:
| Before | After |
|---|---|
| image | Screenshot from 2023-03-20 12-40-03 |

Fixes: #1577


CarlosNihelton (approved):
LGTM 👍🏽

PR #1685, opened 2023-03-20
Allow installing alongside BitLocker if there's free space

jpnurmi:
Subiquity handles BitLocker partitions fine for guided partitioning. It correctly leaves out the "resize" option and only offers a "use gap" option when there's enough free space. There is no reason to force users to turn off BitLocker when installing into an unused area alongside a partition with BitLocker enabled.


jpnurmi:
PXL_20230320_140727159
PXL_20230320_141135763
PXL_20230320_141349387

dbungert (approved):
Conceptually this is fine, I think it just boils down to convincing ourselves that we have enough testing.

elioqoshi (approved):
What Dan said. Looks good on the surface, have to test out edge cases and ensure we did cover all scenarios.

kenvandine (approved):
Conceptually this looks good, let's test this thoroughly

PR #1686, opened 2023-03-20
stage libibus-1.0-5 and libglib2.0-bin

### LINKED ISSUE #1680: Theme selection did not apply on the first boot
I selected the dark theme during installation but the target system has the light theme.

Screenshot from 2023-03-19 11-34-58

kenvandine:
stage libibus-1.0-5 and libglib2.0-bin to ensure we use binaries that match our version of glib

Fixes #1680


seb128 (approved):
Thanks!

PR #1696, opened 2023-03-23
Use a fake desktop service to speed up integration testing in the CI

jpnurmi:
This way, we don't need to install `ubuntu-desktop-minimal` which can sometimes take several minutes.


d-loose (approved):
:+1:

PR #1691, opened 2023-03-21
Add integration test for dumping screenshots

jpnurmi:
This integration test allows dumping screenshots of ~all pages in the installer

https://github.com/canonical/ubuntu-desktop-installer-screenshots

PR #1697, opened 2023-03-23
Integration test: extract page test helpers into a separate file

jpnurmi:
This is a preparation step for #1691 to allow reusing the logic for taking screenshots of all pages.


d-loose (approved):
LGTM :+1:

PR #1703, opened 2023-03-24
Present a fake "install alongside" option that leads to BitLocker instructions

jpnurmi:
If Subiquity doesn't offer any "install alongside" option and we detect BitLocker on the system, assume that the option got filtered out due to BitLocker. At this point, we're already past the minimum installation size check so we can safely assume that there's enough disk space in total. In this case, we want to present a fake "install alongside" option that leads to BitLocker instructions.

NOTE: The BitLocker instructions will be updated separately. This PR is only about the flow change.

The following "before" and "after" screenshot and recording are taken with a machine configuration that has one large partition with BitLocker enabled.

## Before

Subiquity does not offer any "alongside" option because there's no free space and the BitLocker partition cannot be resized.

Screenshot from 2023-03-24 12-38-51

## After

We insert a fake "alongside" option that leads to BitLocker instructions.

[Screencast from 2023-03-24 12-56-24.webm](https://user-images.githubusercontent.com/140617/227518333-f6445a3f-6dde-4c9a-ab65-839f96b49708.webm)

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/installer.dart
+++ b/packages/ubuntu_desktop_installer/lib/installer.dart
@@ -468,7 +468,9 @@ class _UbuntuDesktopInstallerWizard extends StatelessWidget {
           ? Routes.turnOffBitlocker
           : Routes.allocateDiskSpace;
     } else if (service.guidedTarget == null) {
-      if (arguments == InstallationType.erase) {
+      if (arguments == InstallationType.bitlocker) {
+        return Routes.turnOffBitlocker;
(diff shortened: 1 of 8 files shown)


elioqoshi (approved):
Thanks, this looks sensible as discussed

d-loose (approved):
LGTM :+1:

mwhudson:
FWIW we should discuss in Prague or somewhere/when about how subiquity can present options that are not available for some reason. We don't really want the client guessing like this in the long run :-)

PR #1704, opened 2023-03-24
Prevent editing partitions with BitLocker enabled

jpnurmi:
BitLocker'd partitions cannot be resized or mounted. There's no reason to let the user open the partition edit dialog at all. Notice that deleting such a partition is allowed by pressing the "-" button, though.

| Before | After |
|---|---|
| Screenshot from 2023-03-24 14-32-41 | Screenshot from 2023-03-24 14-32-57 |


elioqoshi (approved):

elioqoshi:
LGTM!

PR #1724, opened 2023-03-28
Correct spelling of Ethernet

robert-ancell:
Ethernet should use a capital letter.


seb128 (changes requested):
The issue with patching translations is that it creates conflict in weblate which doesn't know how to handle the fact that its own records get different from what is in the vcs.

I've forced an export of the currently pending change from weblate, could we at least merge !1717 first and rebase to try to minimize the risk of conflict?

jpnurmi:
Alternatively, we could only change `app_en.arb` and let Weblate warn translators about the case mismatch.

seb128:
> Alternatively, we could only change `app_en.arb` and let Weblate warn translators about the case mismatch.

do you know how weblate and flutter handle that case? in a traditional gettext world the translations would be flagged as fuzzy and not be used anymore which would mean a visible regression for users until the translators go and revalidate the localized string

jpnurmi:
I saw Weblate automatically proposing fixes to similar mismatches before so I thought it could take care of Ethernet vs. ethernet mismatches too but I'm probably wrong because I don't see any case mismatch on the list of translation checks. :(

https://docs.weblate.org/en/latest/user/checks.html#translation-checks

jpnurmi (approved):
LGTM :+1:

seb128 (approved):
thanks!

PR #1739, opened 2023-03-29
CI: enable PR workflows on non-main branches

jpnurmi:
There will be `ubuntu/foo` branches in the future, and we might want to have some short-lived canary branches too. The CI should run whenever opening PRs against those branches.


seb128 (approved):
Right, that should be fine, it could increase our usage and create quota issues with the snap artifacts maybe if we were pushing wip to feature branches and doing merge requests for those but that's not the case today

jpnurmi:
We discussed it when we were about to add support for snap artifacts and ended up doing it so that the snap is built for every PR but the artifact is only stored when manually requested.

PR #1744, opened 2023-03-29
KeyboardLayoutDetector: remove subiquity_client export

jpnurmi:
Exporting all of `subiquity_client.dart` from `keyboard_layout_detector.dart` makes VS Code automatically offer `keyboard_layout_detector.dart` import whenever using types from `subiquity_client.dart` so it accidentally ended up being imported in completely unrelated places such as AD & confirmation.


d-loose (approved):
:+1:

PR #1745, opened 2023-03-29
Move WizardBar out of WizardPage

jpnurmi:
A single list of actions was convenient when they all used to be on the right. Since then, we've moved the back button to the other side, added a page indicator, and we might soon want to run animations too. This is a simple preparation step for giving individual pages more control over the bottom bar.


d-loose (approved):
LGTM :+1:

PR #1747, opened 2023-03-29
snapcraft: set non-snap environment

dbungert:
`netplan apply` wants to run without this environment setup. Lay the groudwork for a subiquity patch that will run netplan without these environment changes.


dbungert:
Companion change: https://github.com/canonical/subiquity/pull/1623

kenvandine (approved):
Thanks!

PR #1749, opened 2023-03-30
Improve resized and pre-cached mascot avatar

d-loose:


packages/ubuntu_desktop_installer/lib/widgets/mascot_avatar.dart
```diff
@@ -1,24 +1,28 @@
...
 
+const _kMascotAsset = AssetImage('assets/mascot.png');
+
 class MascotAvatar extends StatelessWidget {
-  const MascotAvatar({super.key, this.radius = 80});
+  const MascotAvatar(
+      {super.key, this.image = _kMascotAsset, this.height = 160});
+
+  final ImageProvider<Object> image;
+  final double height;
```
jpnurmi, line 10: should we either expose `width` or call it `size` instead and pass it as width to make it safe to use this in horizontal layouts too?

jpnurmi (approved):
Thanks for sorting it out! LGTM :+1:

PR #713, opened 2022-02-28
Proxy asset bundle

jpnurmi:
This change replaces the easy-to-forget FlavorImage class with a proxy
asset bundle that hides the vanilla vs. flavor asset selection under
the hood. The proxy asset bundle is injected to the widget tree with
help of DefaultAssetBundle. That way, Image.asset() and alike work
without any extra consideration.

Ref: #700

PR #1761, opened 2023-03-31
ProxyAssetBundle: use synchronous IO lookups

jpnurmi:
`ProxyAssetBundle` (#713) takes care of looking up image assets so that flavors can optionally override them. For example, `ubuntu-flavor-installer/assets/mascot.png` would override `packages/ubuntu-desktop-installer/assets/mascot.png` but it would fall back to the U-D-I asset if the flavor did not provide its own mascot image.

While resolving each asset, `ProxyAssetBundle` was doing up to two asynchronous `File.exists()` calls. While it's generally a good idea to use asynchronous APIs to avoid blocking the GUI, these asynchronous lookups would cause us to miss frames whenever looking up assets because the result would never be available within the same frame that requested the asset.

See also:
- #1743
- #1746
- #1749


d-loose (approved):
LGTM :+1:

PR #1763, opened 2023-03-31
Allow passing arbitrary CLI args to subiquity

jpnurmi:
Separated by `--`, the rest are forwarded to subiquity. For example, `--foo` goes to UDI whereas `--bar --baz` goes to subiquity:

```
./ubuntu-desktop-installer --foo -- --bar --baz
```

This way, we don't need to duplicate subiquity's arguments in the GUI but one can pass arguments straight to subiquity and that way test more exotic use cases with help of dry-run mode, custom machine config, and socket path even in a live session. We might still need to tweak the socket path lookup on the GUI side to avoid arbitrary paths but that comes separately if even necessary.


d-loose (approved):
LGTM :+1:

PR #1765, opened 2023-03-31
WelcomeModel: avoid calling .length on a lazy iterable

jpnurmi:
Micro-optimization: call `.length` on the final `List`, not on the lazy `Iterable` returned by `loadLocalizedLanguages()`. Also, calling `loadLanguages()` again should not accumulate. :)


d-loose (approved):
:+1:

PR #1775, opened 2023-04-04
scripts/update-subiquity: remove git submodule calls

jpnurmi:
Dedicate this script for updating the sha1's of subiquity dependencies and let the user decide which subiquity_client revision is checked out.

The git submodule calls are not only unnecessary whenever checking out dependabot branches that have already bumped the submodule but can also create problems because `git submodule update --remote` defaults to the `origin` remote which may point to a personal fork when following GitHub's recommended workflows. In that case, it would have to use the `upstream` remote instead but we cannot assume that either because it's just a recommendation.

### DIFF
--- a/scripts/update-subiquity-deps
+++ b/scripts/update-subiquity-deps
@@ -1,15 +1,6 @@
 #!/usr/bin/env python3
 
-import subprocess
 from ruamel.yaml import YAML
-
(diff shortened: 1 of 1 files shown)


d-loose (approved):
:+1:

PR #1776, opened 2023-04-04
Replace hard-coded SubiquityXxx instances with conditional services

jpnurmi:
To allow calling `runInstallerApp()` with pre-registered fake services for testing and screenshotting purposes.


d-loose (approved):
LGTM :+1:

PR #1781, opened 2023-04-05
postinst: don't apply default input source on the target

jpnurmi:
Let it be picked from `/etc/default/keyboard` instead.


seb128 (approved):
Thanks, that makes sense, we should inherit the default from the system so if a sysadmin change it in /etc new user will get the right value in the graphical session

Issue #1782, opened 2023-04-05
Empty language list on second run


jpnurmi:
Close and re-run the installer. Only reproducible in the live session.

A regression caused by:
- https://github.com/canonical/ubuntu-desktop-installer/pull/1773

```
flutter: ERROR ubuntu_desktop_installer: Unhandled exception
	UnsupportedError: Unsupported operation: Infinity or NaN toInt
#0      RenderBox.size (package:flutter/src/rendering/box.dart)
#1      _PositionedListState._schedulePositionNotificationUpdate.<anonymous closure> (package:scrollable_positioned_list/src/positioned_list.dart:338)
#2      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1289)
#3      SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1227)
#4      SchedulerBinding.scheduleWarmUpFrame.<anonymous closure> (package:flutter/src/scheduler/binding.dart:942)
#5      _rootRun (dart:async/zone.dart:1390)
#6      _CustomZone.run (dart:async/zone.dart:1300)
#7      _CustomZone.runGuarded (dart:async/zone.dart:1208)
#8      _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1248)
#9      _rootRun (dart:async/zone.dart:1398)
#10     _CustomZone.run (dart:async/zone.dart:1300)
#11     _CustomZone.bindCallback.<anonymous closure> (dart:async/zone.dart:1232)
#12     Timer._createTimer.<anonymous closure> (dart:async-patch/timer_patch.dart:18)
#13     _Timer._runTimers (dart:isolate-patch/timer_impl.dart:398)
#14     _Timer._handleMessage (dart:isolate-patch/timer_impl.dart:429)
#15     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:192)
```

![image](https://user-images.githubusercontent.com/140617/230041583-b8801696-2096-4938-af21-00a7a46d61e7.png)

jpnurmi:
It's empty even after closing the installer, hiding the dock again, and re-login. The only practical difference is that subiquity is already up and running at startup time.

PR #1773, opened 2023-04-03
Remove animated auto-scrolling in language & keyboard layout lists

jpnurmi:
### Before

[auto-scrolling-before.webm](https://user-images.githubusercontent.com/140617/229713925-f83ec2a8-5f89-4159-b2e5-74e2c90fa6ad.webm)

### After

[auto-scrolling-after.webm](https://user-images.githubusercontent.com/140617/229713938-b6740ae4-3579-4b8e-8ec6-85ac78bbd70c.webm)

P.S. I switched to the [scrollable_positioned_list](https://pub.dev/packages/scrollable_positioned_list) package because it supports an initial scroll position and unlike the previously used [scroll_to_index](https://pub.dev/packages/scroll_to_index), it's also able to jump [without a delay](https://github.com/quire-io/scroll-to-index/blob/be171b267f318e47436fdfb134e30fd5ce7b2efb/lib/scroll_to_index.dart#L216).

Close: #1499

PR #1783, opened 2023-04-05
Revert "Remove animated auto-scrolling in language & keyboard layout lists"

jpnurmi:
Reverts canonical/ubuntu-desktop-installer#1773 because the language list appears empty on the second run:

- #1782


d-loose (approved):
Hope we find another solution
:+1:

jpnurmi:
I have a potential workaround
- https://github.com/canonical/ubuntu-desktop-installer/pull/1787

I'll continue debugging it a bit further and try to understand the root cause of why the entire welcome page is briefly built at 0x0 size.

PR #1787, opened 2023-04-05
ListWidget: guard against empty size constraints

jpnurmi:
I'm not sure why this only happens in snap+live but the entire app gets briefly built with 0x0 size which breaks the initial offset calculation of `ScrollablePositionedList`.

Close: #1782
Close: #1783


d-loose (approved):
LGTM :+1:

PR #1790, opened 2023-04-05
DiskStorageService.setGuidedStorage(): remove unused return value

jpnurmi:
This is a noisy preparation step for applying the storage config from setGuidedStorage().


d-loose (approved):
:+1:

PR #1791, opened 2023-04-05
Apply storage config after sending the guided choice

jpnurmi:
For guided partitioning, the GUI collects input across several screens:

- Installation type: whether to use LVM or encryption
- Select disk: which disk to erase
- Install alongside: how much space to free up
- Choose security key: encryption password

These bits and pieces of information cannot be sent to Subiquity from each screen but must be collectively sent as a "guided choice". This is done when entering the confirmation screen because at that point all the necessary input has been collected.

When the guided storage target had been chosen, the user was presented with the resulting storage config which was then applied when the user pressed "Install". The problem was that udev events could invalidate the storage config while sitting at the confirmation screen, before pressing "Install".

Therefore, we have to change the flow so that in case of guided partitioning, the storage config is already applied when entering the confirmation screen, right after choosing the guided storage target.

#### Guided

- Choose a guided storage target when entering the confirmation screen
- Apply the storage config after pressing "Install"

#### Manual

- Apply the storage config when done with manual partitioning, before entering the confirmation screen

#### Guided

- Choose a guided storage target _and_ apply the storage config when entering the confirmation screen

#### Manual

- No changes

Fixes: [LP#2012077](https://bugs.launchpad.net/subiquity/+bug/2012077)

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/write_changes_to_disk/write_changes_to_disk_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/write_changes_to_disk/write_changes_to_disk_model.dart
@@ -43,7 +43,6 @@ class WriteChangesToDiskModel extends SafeChangeNotifier {
 
   /// Starts the installation process.
   Future<void> startInstallation() async {
-    await _service.setStorage();
     _service.securityKey = null; // no longer needed
     await _client.confirm('/dev/tty1');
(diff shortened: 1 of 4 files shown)


ogayot (approved):
Yep, looks good from my standpoint. This is closer from the implementation in the text installer.

Thanks!

PR #1794, opened 2023-04-06
DiskStorageService.setStorage(): remove confusing unused arg

jpnurmi:
The argument is an old left-over which is no longer used. This method applies the current storage configuration, whatever has been requested using either the guided or manual partitioning APIs. Dropping the unused arg helps to eliminate some confusion in #1791.


d-loose (approved):
:+1:

PR #1795, opened 2023-04-06
Replace ScrollablePositionedList with a plain ListView

jpnurmi:
ScrollablePositionedList creates problems with focus traversal and is probably overkill for what we need, anyway. Using a plain ListView and calculating the offsets ourselves helps to make arrow key navigation work sensibly in list views again. All alignment tests added in #1773 are passing.


d-loose (approved):
Nice, LGTM! :+1:

PR #1808, opened 2023-04-13
ListWidget: use OrderedTraversalPolicy

d-loose:
Ensures that the keyboard focus doesn't suddenly jump out of list widgets due to Flutter's default 'reading order' focus traversal policy.


jpnurmi (approved):
Thanks! This looks safe yet a big improvement. LGTM :+1:

PR #1809, opened 2023-04-13
Postpone source & storage calls

jpnurmi:
The storage call is necessary on startup to fetch such information that is needed to build the wizard (has rst, total disk space, ...). A valid installation source must be selected before the storage call, however, it must only be done in manual installation mode to avoid conflicting with autoinstall source config.

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/installer.dart
+++ b/packages/ubuntu_desktop_installer/lib/installer.dart
@@ -114,17 +114,6 @@ Future<void> runInstallerApp(
 
   final subiquityClient = getService<SubiquityClient>();
   await subiquityClient.setVariant(Variant.DESKTOP);
-  await subiquityClient.setSource(InstallationMode.normal.source);
-
-  // Use the default values for a number of endpoints
-  // for which a UI page isn't implemented yet.
-  await subiquityClient.markConfigured([
-    'mirror',
-    'proxy',
-    'ssh',
-    'snaplist',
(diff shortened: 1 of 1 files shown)


ogayot (commented):
I'm not familiar with the code so I can't give you very useful feedback.
That said, the PR description sounds fine. 

One question though, are we sure that the call to /source is going to happen _before_ the user selects normal vs minimal install?

d-loose (approved):
LGTM :+1:

jpnurmi:
> I'm not familiar with the code so I can't give you very useful feedback.
> That said, the PR description sounds fine. 

Sorry, it would've been more appropriate to /cc rather than request a review. :)

> One question though, are we sure that the call to /source is going to happen _before_ the user selects normal vs minimal install?

Yes, the /source call is right after the loading spinner disappears but before the user has a chance to interact with the GUI. Selecting the source comes a few pages later.

ogayot:
> > I'm not familiar with the code so I can't give you very useful feedback.
> > That said, the PR description sounds fine.
> 
> Sorry, it would've been more appropriate to /cc rather than request a review. :)
> 
> > One question though, are we sure that the call to /source is going to happen _before_ the user selects normal vs minimal install?
> 
> Yes, the /source call is right after the loading spinner disappears but before the user has a chance to interact with the GUI. Selecting the source comes a few pages later.

Sounds good to me then!

PR #1812, opened 2023-04-13
(fix) WSL windows tests - `pumpAndSettle` no longer needed

CarlosNihelton:
> This is test code fix. It shouldn't affect any other part of the apps.

It seems to be at least since we moved to YaruWindow. The test would hang forever waiting the circular progress indicator to finish, but it's infinite. Removing the `await tster.pumpAndSettle()` allows the window to close when the server exits, as it should.

### DIFF
--- a/packages/ubuntu_wsl_setup/integration_test/e2e/setup_test.dart
+++ b/packages/ubuntu_wsl_setup/integration_test/e2e/setup_test.dart
@@ -28,6 +28,5 @@ void main() {
       confirmedPassword: 'password123',
     );
     await testApplyingChangesPage(tester, expectClose: true);
-    await tester.pumpAndSettle();
   });
 }
--- a/packages/ubuntu_wsl_setup/integration_test/e2e/setup_with_prefill_test.dart
+++ b/packages/ubuntu_wsl_setup/integration_test/e2e/setup_with_prefill_test.dart
@@ -24,6 +24,5 @@ void main() {
       confirmedPassword: 'password123',
     );
     await testApplyingChangesPage(tester, expectClose: true);
-    await tester.pumpAndSettle();
   });
 }


jpnurmi (approved):
LGTM :+1: 

Sorry for breaking the test! Is there a possibility to run it in the CI?

CarlosNihelton:
> Is there a possibility to run it in the CI?

Running anything for real in WSL in CI is currently very complicated. We need to dedicate more time in the test preparation to make it simpler to consume in other repos. Then it will be a pleasure to open a PR in here to run it.

jpnurmi:
Ok, np. Are there any special steps needed for running the test if I setup WSL locally?

CarlosNihelton:
You'll need to have a self-signed certificate to allow signing the appx package. You'll also need the go language and the current version of Visual Studio installed in your machine.

Then you'll need to:

1. checkout the `ubuntu/WSL` repository.
2. run the `.\e2e\pre-e2e-testing.ps1` script from the Developer Powershell to build the appx for automated testing. Something like:
```
.\e2e\pre-e2e-testing.ps1 -AppID "UbuntuPreview" -RootfsX64 "https://cloud-images.ubuntu.com/wsl/lunar/current/ubuntu-lunar-wsl-amd64-wsl.rootfs.tar.gz" -CertificateThumbprint <your-testing-certificate>
```
That script will download the root filesystem, build an appx and install into the windows machine.

3. export an environment variable pointing to the root of the repository. Some test cases refer to that directory.
`$env:LAUNCHER_REPO_ROOT=$PWD`

4. Run the test cases.
`go test .\launchertester\ -timeout 15m -run TestBasicSetup --distro-name Ubuntu-Preview --launcher-name ubuntupreview.exe`

The test case performs a real setup of a WSL distro instance.

It's what happens in the `ubuntu/WSL` CI: https://github.com/ubuntu/WSL/blob/main/.github/workflows/e2e.yaml#L68-L87

Issue #1817, opened 2023-04-14
chainging the format of existing partition requires the Format value to be set


dbungert:
Suppose an existing partition of filesystem type A, and we use manual partitioning to change that partition to filesystem type B.  In UI this is possible (unrelated - this shows some of the filesystem options we didn't want to offer) - but it doesn't enforce the format checkbox.

Attempting to proceed in this state will yield a "changing partition format requires a wipe value" traceback.  Please enforce the checkbox being set for that case.

jpnurmi:
Changing in relation to the original config? If I go and change it back to A, should I be allowed to leave it unformatted?

dbungert:
Yes, if the filesystem type is set back to the original value, A in this case, the format checkbox being checked is no longer required.

jpnurmi:
Hmm, and presumably this only applies to preserved partitions? What about newly created partitions, are there any restrictions whether they must be formatted or can they be left unformatted?

dbungert:
Yes, this is about preexisting partitions that would start off in a `preserved: true` state.  Newly created partitions are either going to be left unformatted, or have an implied wipe due to getting the mkfs tool run on them.

PR #1820, opened 2023-04-17
Installation summary: present all disks with any partition modifications

jpnurmi:
The installation summary aims to present all file system changes. Make sure to include all disks that have any sort of partition modifications to avoid confusing and mismatching lists of disks vs. partitions. In other words, use the same filter condition for finding disks with modified partitions and the flattened list of modified partitions.

| Before | After |
|--------|--------|
| Screenshot from 2023-04-17 14-08-30 | Screenshot from 2023-04-17 14-08-55 | 

Ref: #1817


seb128 (approved):
Thanks makes sense but please check with the release team if that's ok to land still/if we will want to respin the ISO for the change otherwise we should hold on merging for now

Issue #1819, opened 2023-04-17
wrong NVME partition name shown in disk summary


ogayot:
Hello,

When using NVME storage, the storage summary screen shows incorrect partition names:

![Screenshot from 2023-04-17 11-43-43](https://user-images.githubusercontent.com/4038023/232483382-65028b55-a9f5-446b-b13d-6469529940b0.png)

For the device nvme0n1 (where 1 is the namespace), the partition names are expected to be:

* nvme0n1p1
* nvme0n1p2
* ...
* nvme0n1p9

https://bugs.launchpad.net/subiquity/+bug/2016389

As far as I know, the desktop installer is generating those names from the partition indexes returned by subiquity.

Thanks,
Olivier

asacha-ru:
I can confirm that the issue is fixed in 23.04 release

PR #1821, opened 2023-04-17
Installation summary: fix NVME partition formatting

jpnurmi:
"nvme0n11" vs. "nvme0n1p1"

For historical reasons, we used "{disk}{partition}" because it used to be something like "{disk} #{partition}". However, this does not result in the correct formatting with NVME disks where there needs to be an extra "p". It's not possible to change the translations on the release week but luckily all existing translations have kept it as "{disk}{partition}" so we can safely pass an empty disk argument for the time being, and remove the unused argument later. The "Partition.path" property from subiquity provides the correct value, and we have a "sysname" convenience wrapper for that to take the filename part of the path.

| Before | After |
|---|---|
| Screenshot from 2023-04-17 14-26-37 | Screenshot from 2023-04-17 14-25-27 |

Fixes: #1819


seb128 (approved):
Seems fine to me, as for the other one we should have a +1 from release team before landing

d-loose (approved):
LGTM :+1:

jpnurmi:
Agreed with @sil2100

PR #1823, opened 2023-04-17
Manual partitioning: ensure "format" selection when appropriate

jpnurmi:
Force-select the "Format partition" -checkbox if the FS type is changed. For example, if you take an existing EXT4 partition and change it to BTRFS, the format checkbox must be selected or else Subiquity will throw an exception. Notice that we must fetch the _original_ storage config to be able to compare the FS type of preserved partitions.

| Before | After |
|---|---|
| Screenshot from 2023-04-17 15-59-45 | Screenshot from 2023-04-17 15-59-24 |

Fixes: #1817


d-loose (approved):
LGTM :+1:

seb128 (approved):
Yes please, we should try to get that in the release

jpnurmi:
@sil2100, last fix for the respin merged

PR #1827, opened 2023-04-18
Manual partitioning: fix initial GUI state update

jpnurmi:
Fixes a regression caused by yesterday's quick fix:
- #1823

`GET storage/v2` and `GET storage/v2/orig_config` requests were sent in parallel but only the former notified the GUI to update. Depending on the order of completion, the GUI could have left in a wrong initial state that confusingly stated that existing partitions would be formatted because it didn't see the original storage config. Reorder the calls and await for the `orig_config` to make sure both configs are available before notifying the GUI to update.

Fixes: #1828


d-loose (approved):
LGTM :+1:

seb128 (approved):
Thanks!

PR #1829, opened 2023-04-18
Installation type: remove initial default selection

jpnurmi:
Don't hard-code the initial selection to "Erase" (and then potentially switch to "Alongside") but leave it unselected until available guided partitioning options have been received from Subiquity. This is not the final solution but makes the GUI feel slightly less awkward in such cases where the pre-selected "Alongside" option appears with a slight delay.

Show-case recorded with an extra 500ms delay injected:

## Before

[installation-type-before.webm](https://user-images.githubusercontent.com/140617/232820354-9439f836-4eb7-4a5d-8708-0ba8456986f0.webm)

## After

[installation-type-after.webm](https://user-images.githubusercontent.com/140617/232820312-245dac8b-0709-4186-adba-f5f134bf9636.webm)


jpnurmi:
Turned into a draft. Not supposed to be included in the release.

d-loose (approved):
LGTM :+1:

PR #1831, opened 2023-04-18
Disable storage auto-mounting

jpnurmi:
Unlike Ubiquity, the new GUI is not running as root so we cannot use udisks2-inhibit as a wrapper so we ended up copying the commands.

### DIFF
--- a/snap/local/launcher
+++ b/snap/local/launcher
@@ -40,4 +40,11 @@ fi
 
 [ "$needs_update" = true ] && echo "SNAP_DESKTOP_LAST_REVISION=$SNAP_REVISION" > "$SNAP_USER_DATA/.last_revision"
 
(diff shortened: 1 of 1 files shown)


snap/local/launcher
```diff
@@ -40,4 +40,11 @@ fi
 
 [ "$needs_update" = true ] && echo "SNAP_DESKTOP_LAST_REVISION=$SNAP_REVISION" > "$SNAP_USER_DATA/.last_revision"
 
+# Disable storage automounting (/usr/lib/udisks2/udisks2-inhibit)
```
kenvandine, line 43: @jpnurmi is the path in that comment relevant?

kenvandine (approved):
Great catch!

PR #1834, opened 2023-04-19
launcher: fix udev cleanup

dbungert:
With the exec call, the previous trap will not run.  Leave the process around so that we can run the trap on exit.


jpnurmi (approved):
LGTM :+1:

PR #1838, opened 2023-04-20
Switch to subiquity/main

jpnurmi:
The most notable difference is that in the main branch, a concept of "guided capabilities" has been introduced as a more flexible replacement for "bool use_lvm" and friends.


d-loose (approved):
LGTM :+1:

Issue #1841, opened 2023-04-24
Source selection on startup


jpnurmi:
We are unconditionally selecting the "ubuntu-desktop" source on startup just to be able to call Subiquity's storage API that would hang if no source was selected.

There are a number of issues with this approach:
- First of all, we should not to hardcode sources in the GUI but fetch the available sources from Subiquity.
- Secondly, it conflicted with autoinstall as it came up in #1809.
- Last but not least, we are incorrectly using the minimum installation size reported for "ubuntu-desktop" while the user may later choose to install "ubuntu-desktop-minimal" instead.

Proposal:
- Move "Not enough disk space" after "Updates and other software"
- Remove unconditional source selection on startup

See also:
- #1809

PR #1842, opened 2023-04-24
Move "Not enough disk space" after "Updates and other software"

jpnurmi:
Let the user select between normal and minimal installation before checking the minimum installation size because it depends on the selected source.

Ref: #1841


d-loose (approved):
:+1:

PR #1843, opened 2023-04-24
Fetch source catalogs from Subiquity

jpnurmi:
Remove the hardcoded "normal" vs. "minimal" enum, declare known "ubuntu-desktop" and "ubuntu-desktop-minimal" source IDs, fetch the effective list of sources from Subiquity, present them in the desired order, and use localized descriptions if available.

There's no visual difference by default when the sources are defined as they are in [`livecd-rootfs`](https://git.launchpad.net/livecd-rootfs/tree/live-build/auto/config#n766) (and [`desktop-sources.yaml`](https://github.com/canonical/subiquity/blob/main/examples/desktop-sources.yaml) for dry-run) but this does allow the GUI to run with any source catalogs e.g. the previously used `mixed-sources.yaml` would present a server source.

| Before | After |
|---|---|
| Screenshot from 2023-04-24 11-55-07 | Screenshot from 2023-04-24 11-54-50 |

Ref: #1841

### DIFF
--- a/packages/ubuntu_desktop_installer/integration_test/test_pages.dart
+++ b/packages/ubuntu_desktop_installer/integration_test/test_pages.dart
@@ -136,14 +136,14 @@ Future<void> testConnectToInternetPage(
 
 Future<void> testUpdatesOtherSoftwarePage(
   WidgetTester tester, {
-  InstallationMode? mode,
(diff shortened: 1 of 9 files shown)


d-loose (approved):
LGTM :+1:

Issue #1844, opened 2023-04-24
Desktop vs. server steps


jpnurmi:
Server:
- filesystem
- identity
- keyboard
- locale
- network
- source
- timezone
- ...

Desktop:
- allocate_disk_space
- connect_to_internet
- installation_type
- keyboard_layout
- select_guided_storage
- updates_other_software
- welcome
- where_are_you
- who_are_you
- ...

Proposal:
- Change the Desktop Installer to follow Subiquity's naming as close as possible
- Use nested wizards to e.g. group all filesystem-related pages into one top-level wizard step
- Adding support for partially automated installs becomes much easier when there's some sensible mapping between the server & desktop steps...

See also:
- #1591

PR #1847, opened 2023-04-24
Rename WhereAreYouXxx to TimezoneXxx

jpnurmi:
The original page title was "Where are you?" and that was used as the name for the page, route, and model. Things started getting weird after the page title was changed to "Select your timezone" in #1591. The most sensible name is "timezone" to match with Subiquity. This makes it easier to add support for partially automated installs when Subiquity may report that e.g. "timezone" must be manually configured.

See also:
- #38
- #1591
- #1844


d-loose (approved):
:+1:

Issue #26, opened 2021-03-08
Screen 1 - Choose Language


seb128:
The design for screen n°1, Choose Language

![language](https://user-images.githubusercontent.com/10974330/110310609-07db4580-8003-11eb-83c6-a9c1febf557f.png)

Design details https://people.canonical.com/~platform/desktop/installer/1.%20Choose%20Language/

madsrh:
I wonder why the list is centered in the mockup above 🤔 Looking at the other mockups, I see that all of them are left-justified.

...OR perhaps the other screens should be centered too 😉 

[Connect to the internet](https://github.com/canonical/ubuntu-desktop-installer/issues/30)
![Wi-Fi-mockup](https://user-images.githubusercontent.com/3986894/110615052-4a338c80-8193-11eb-8c14-03eb7c9d8a17.jpg)

[Updates and other software](https://github.com/canonical/ubuntu-desktop-installer/issues/31)
![Updates-mockup](https://user-images.githubusercontent.com/3986894/110615107-57e91200-8193-11eb-8bf9-ff7b2b4c2ee4.jpg)

long-chung:
All screens should be left aligned (Except for Choose language) as per the designs.
This is because it requires more effort to read central aligned text.

fitojb:
Yep, the margins should make everything look centered but the text should be flush left or right depending on the language.

Mannshoch:
Why not add on the left side a chooser for the Keyboard?
![grafik](https://user-images.githubusercontent.com/7795891/111594077-e6304a00-87ca-11eb-921a-ea3a22a5e34d.png)

May a better solution.
![grafik](https://user-images.githubusercontent.com/7795891/111594721-b6ce0d00-87cb-11eb-89a4-3bb57142b980.png)

jpnurmi:
This was the first page that was implemented and should be now fully functional. I'm closing the ticket as agreed to make it easier to follow the status.

PR #1849, opened 2023-04-24
Rename WelcomeXxx to LocaleXxx

jpnurmi:
Match the name of the underlying implementation (page, route, model) with Subiquity. This makes it easier to add support for partially automated installs when Subiquity may report that e.g. "locale" must be manually configured.

See also:
- #26
- #1591
- #1844


packages/ubuntu_desktop_installer/lib/installer.dart
```diff
@@ -307,7 +307,7 @@ class _UbuntuDesktopInstallerLoadingPage extends StatelessWidget {
 }
 
 enum InstallationStep {
-  welcome,
+  language,
```
d-loose, line 310: should this be `locale` as well?

jpnurmi: thanks, good catch!

PR #1850, opened 2023-04-24
Group everything FS-related into filesystem/

jpnurmi:
To match with "filesystem" in Subiquity. This is a preparation step for semi-automated installation support including nested filesystem wizard that will follow separately.

Ref: #1844


d-loose (approved):
:+1:

PR #1856, opened 2023-04-25
rename TurnOffBitLockerXxx to BitLockerXxx

d-loose:

See also:
* #26
* #1591
* #1844


jpnurmi (approved):
Thanks, LGTM :+1:

PR #1858, opened 2023-04-25
rename ChooseYourLookXxx to ThemeXxx

d-loose:

See also:
* #26
* #1591
* #1844


packages/ubuntu_desktop_installer/lib/routes.dart
```diff
@@ -11,7 +11,7 @@ abstract class Routes {
   static const updatesOtherSoftware = '/update-other-software';
   static const installationComplete = '/installation-complete';
   static const writeChangesToDisk = '/write-changes-to-disk';
-  static const chooseYourLook = '/choose-your-look';
+  static const theme = '/choose-your-look';
```
jpnurmi, line 14: could you rename the route as well?

jpnurmi (commented):
could you rename test/choose_your_look and the files in it too?

jpnurmi (approved):
LGTM :+1:

PR #1864, opened 2023-04-26
Add forwarding Makefile for conveniently installing Subiquity deps

jpnurmi:
Before:
```sh
$ make -C packages/subiquity_client/subiquity install_deps
```

After:
```sh
$ make install_deps
```


d-loose (approved):
:+1:

PR #1866, opened 2023-04-26
README.md: remove manual live-run instructions

jpnurmi:
There are (safe) instructions to test live images above. It's better not to give (dangerous) instructions to run the installer in live-mode on a normal desktop. Also, emphasize that `flutter run` is a dry run.


d-loose (approved):
:+1:

PR #1870, opened 2023-04-26
ubuntu_desktop_installer: use package-syntax for barrel imports

jpnurmi:
To get rid of the ugly "../../foo.dart" imports. Import them using the package syntax instead, as if they were standalone. Keep using local imports for local files in the same directory or in subdirectories.

As a bonus, this will help to reduce the noise when restructuring half a dozen pages into filesystem/ in #1850 where the relative imports would become one level deeper...


d-loose:
should we add the `avoid_relative_lib_imports` linter rule as well?

jpnurmi:
Yes please, +1 from me! :+1: I think there are still some imports left to migrate before we can turn it on in the shared analysis options.

PR #1872, opened 2023-04-26
Nested filesystem wizard

jpnurmi:
This is a preparation step for semi-automated installs. If Subiquity reports that "filesystem" requires configuration, the user is taken to the subwizard. This is just an implementation detail, though. Visually, it looks the same as before in manual installs.

Ref: #1844


d-loose (approved):
:+1:

PR #1874, opened 2023-04-26
Move has_bitlocker call from startup to installation_type

jpnurmi:
This takes us one step closer to removing the unconditional source selection on startup. The one and only screen that needs this information is the _Installation type_ screen. We can fetch it there instead of doing it on startup.


d-loose (approved):
:+1:

PR #1878, opened 2023-04-27
Nested welcome wizard

jpnurmi:
This subwizard is only shown in manual installation mode and does not have a respective step in Subiquity. In the future, this subwizard may also include a "Repair Ubuntu" page if/when the functionality is available in Subiquity.

Ref: #1844


d-loose (approved):
:+1:

PR #1879, opened 2023-04-27
Remove Wizard(Action).done

jpnurmi:
It was used to finish the telemetry report and write it on the disk but that approach didn't work because Subiquity copies it to /target when it finishes installing and it's too late to write anything from the GUI when it quits. Also, writing any telemetry data from early exits is irrelevant because it doesn't end up stored anywhere anyway.


jpnurmi:
For reference, `_UbuntuDesktopInstallerWizardObserver.onDone` was removed in:
- #1202

d-loose (approved):
:+1:

Issue #1891, opened 2023-04-28
Split ubuntu_test


jpnurmi:
A proposal to split `ubuntu-desktop-installer/packages/ubuntu_test` into:
- `subiquity_test`
  Subiquity-related test utils and pre-generated mocks and placed into a `packages/subiquity_test` subdir at [canonical/subiquity_client.dart](https://github.com/canonical/subiquity_client.dart)
- `ubuntu_test`
  Generic ubuntu(_localizations)-related test extensions and placed into a `packages/ubuntu_test` subdir next to `ubuntu_localizations` at [canonical/ubuntu-flutter-plugins](https://github.com/canonical/ubuntu-flutter-plugins)
- `yaru_test`
  Generic Yaru/widget-related finders and matchers at [ubuntu/yaru_test.dart](https://github.com/ubuntu/yaru_test.dart)

PR #1892, opened 2023-04-28
Demote test-specific mocks from ubuntu_test to desktop_service_test

jpnurmi:
ubuntu_test has been used for sharing the subiquity client mocks that are used almost everywhere but these mocks can be generated locally because they are specific to one test.

Ref: #1891


d-loose (approved):
:+1:

PR #1898, opened 2023-04-29
Streamline TextField tests

jpnurmi:
- Add and use `isEnabled`, `isDisabled`, `hasFocus`, and `hasNoFocus` matchers
- Add and use `find.textField(str)` in favor of `find.widgetWithText(TextField, str)`
- Use `tester.enterText()` + `find.textField()` in favor of `tester.enterTextValue()` with confusing semantics due to a hidden null check


d-loose (approved):
:+1:

PR #1906, opened 2023-05-02
Add additional convenience finders for buttons and alike

jpnurmi:
We are using a lot of check buttons, radio buttons, icon buttons, and list tiles across the GUI. These finders make it more pleasant and less verbose to test them.

Note: Eventually, all these new common finders and matchers will be extracted into a separate package.


d-loose (approved):
LGTM :+1:

PR #1908, opened 2023-05-02
upgrade ubuntu_wizard to 1.0.0-beta

d-loose:
Since `WizardObserver` was removed from ubuntu_wizard, the telemetry service now uses a `NavigatorObserver`.

There's no method corresponding to `onInit()` - is it enough to call `addStage()` in `didPush()`?


jpnurmi (approved):
LGTM :+1:

PR #1912, opened 2023-05-03
screenshot_test: run with only one theme by default

jpnurmi:
The test is much more convenient to work with if it doesn't run all four themes by default but only when explicitly requested.


d-loose (approved):
LGTM :+1:

PR #1922, opened 2023-05-08
refactor(locale): decouple from subiquity

jpnurmi:
Get and set the locale via a service class that can be implemented with alternative backends, such as org.freedesktop.locale1 aka systemd-localed, in addition to the subiquity-based one.


d-loose (approved):
LGTM :+1:

PR #1923, opened 2023-05-08
refactor(timezone): decouple from subiquity

jpnurmi:
Get and set the timezone via a service class that can be implemented with alternative backends, such as org.freedesktop.timedate1 aka systemd-timedated, in addition to the subiquity-based one.


d-loose (approved):
LGTM :+1:

PR #1927, opened 2023-05-08
ci: run all UDI integration tests

jpnurmi:
The screenshot test kept breaking because it wasn't run in the main UDI repo but only in the screenshot repo.


d-loose (approved):
LGTM :+1:

PR #1929, opened 2023-05-08
refactor(source): merge "not enough disk space" into /source sub-wizard

jpnurmi:
An issue exposed by the newly added tests: _Not enough disk space_ page was missing a back-button.

If the user selects _Normal_ installation source and ends up on the "not enough disk space" page, there must be a back button to allow them to go back and select _Minimal_ source instead...

Ref: #1844


d-loose (approved):
LGTM :+1:

PR #1933, opened 2023-05-08
fix(network): use overrideWithValue() to avoid disposing multiple times

jpnurmi:
The list of devices and access points are owned by the model, whereas [overrideWith()](https://pub.dev/documentation/flutter_riverpod/latest/flutter_riverpod/ChangeNotifierProvider/overrideWith.html) must return a new value because it will be automatically disposed when appropriate.


d-loose (approved):
LGTM :+1:

PR #1937, opened 2023-05-09
refactor(active-directory): decouple from subiquity (#1930)

jpnurmi:
Manage Active Directory via a service class that can be implemented with alternative backends e.g. directly accessing realmd, in addition to the existing subiquity-based service.


d-loose (approved):
LGTM :+1:

PR #1940, opened 2023-05-09
refactor(storage): rename (Disk)StorageService to match Subiquity

jpnurmi:
It's easier to understand the relationships between UDI & Subiquity when the terms and concepts match.


d-loose (approved):
LGTM :+1:

PR #1946, opened 2023-05-09
refactor(keyboard): remove unused streams and fix service variable name

jpnurmi:
The streams were used for auto-scrolling the two list views side by side but we don't have a separate list view for variants anymore.


d-loose (approved):
LGTM :+1:

PR #1947, opened 2023-05-09
refactor(identity): decouple from subiquity

jpnurmi:
Get and set identity via a service class that can be implemented with alternative backends e.g. directly accessing org.freedesktop.accounts, in addition to the existing subiquity-based service.


d-loose (approved):
LGTM :+1:

PR #1926, opened 2023-05-08
migrate filesystem view models to riverpod

d-loose:
Continuation of #1918.
Migrates the remaining view models of the `filesystem`-related pages to riverpod.

PR #1952, opened 2023-05-09
refactor: move model providers

jpnurmi:
We started with page-scoped static providers:
- https://github.com/canonical/ubuntu-desktop-installer/pull/1918

But ended up with a more idiomatic riverpod approach:
- https://github.com/canonical/ubuntu-desktop-installer/pull/1926

This PR applies the latter approach to the providers added in the former.


d-loose (approved):
LGTM :+1:

Issue #1736, opened 2023-03-28
ux: improve interface fluidity across different loads


tim-hm:
## background

Several steps in the installer are not as "fluid" as they could be. This is particularly noticeable when navigating to into/out of the "keyboard layout" screen (the language list jumps) and the "connect to a network" (the radios quickly toggle on/off before settling). 

Here's an example of the language list loading: 

[Screencast from 2023-03-28 17-40-13.webm](https://user-images.githubusercontent.com/1335525/228316414-de1b7334-4e56-4ace-9be7-480404c9a620.webm)

<br />

On some hardware this will be worse -- especially if a live session consumes most of the system's resources. Additionally, as the installer matures with more features there will be more opportunities for loading jitter to sneak in.

Its likely that assuming communicating with subiquity would always be fast was incorrect (its over a unix socket so not totally unreasable). Therefore, we should adjust our approach to prevent loading jitters.

## related issues

- https://github.com/canonical/ubuntu-desktop-installer/issues/1396
- https://github.com/canonical/ubuntu-desktop-installer/issues/1627
- https://github.com/canonical/ubuntu-desktop-installer/issues/1503
- https://github.com/canonical/ubuntu-desktop-installer/issues/1499

## suggestion

I think we can boil it down to a single requirement: **the installer should expect async request delays and have a common approach to dealing with them**.

As I haven't dug into the architecture of the flutter installer I'm unable to suggest an approach. I will try to dig into the architecture this week and come back with a suggestion.

PR #1954, opened 2023-05-10
refactor(wizard): replace abstract WizardAction by WizardButton widget

jpnurmi:
Each individual button needs to be able to manage its state (enabled, disabled, loading, ...) and rebuild as appropriate. Make them widgets rather than abstract actions so that they get their own contexts, and present the busy loading state with a spinner but only after a 100ms delay to avoid flashing spinners on every Next-click.

See also:
- #1736
- https://github.com/ubuntu-flutter-community/wizard_router/pull/42#issuecomment-1541894446


jpnurmi:
A proof of concept with fake 2s delays for keyboard & network:

[Screencast from 2023-05-10 13-54-59.webm](https://github.com/canonical/ubuntu-desktop-installer/assets/140617/1a4d2159-a57c-4075-b993-d162e031c5ab)

d-loose (approved):
LGTM :+1:

PR #1956, opened 2023-05-10
feat: show error text for invalid mount point

d-loose:
Adds an error message to the mount point validation logic from #1706

image


jpnurmi (approved):
Not that it adds much info but makes it clear at least :) LGTM :+1:

jpnurmi:
Alternatively, we could have a multi-validator instead and report "must start with /" and "cannot contain spaces" separately?

jpnurmi (approved):
LGTM :+1:

PR #1957, opened 2023-05-10
refactor(timezone): prepare for model pre-loading

jpnurmi:
The relationship between the view model and a separate controller managed by the widget state would be problematic for pre-loading the model from `WizardRoute.onLoad`. Restructure it so that we will be able to do a full init before building the page.

Ref: #1736


d-loose (approved):
LGTM :+1:

PR #1959, opened 2023-05-10
feat: pre-load view models

jpnurmi:
Fetch data before entering pages.

Ref: #1736


jpnurmi:
Some pages are much harder than others and smacking it all into one PR makes me miss things so I'll go with smaller separate PRs instead...

PR #1960, opened 2023-05-11
refactor(rst): make use of the new WizardRoute.onLoad

jpnurmi:
This makes the RST route fully standalone and makes sure it's correctly shown even if/when the welcome aka try-or-install route is skipped, which was missed in #1878.


d-loose (approved):
LGTM :+1:

PR #1961, opened 2023-05-11
test: clean up, modernize, extend legacy page & model tests

jpnurmi:
The old `setUpApp` + `late MaterialApp` approach is getting in the way of adapting the tests to pre-load models. Switch to the common `buildModel` + `buildPage` + `buildApp` approach where the latter is shared among all page tests.

See also:
- #1959


d-loose (approved):
LGTM :+1:

PR #1962, opened 2023-05-12
refactor(secure-boot): use WizardRoute.onLoad

jpnurmi:
Make the route fully standalone in a way that it can be anywhere in the wizard and it can be even configured out without changing other routes.


d-loose (approved):
LGTM :+1:

PR #1973, opened 2023-05-15
refactor(installer): use SubiquityClient.monitorStatus()

jpnurmi:
One thing less to abstract away in non-Subiquity mode.


d-loose (approved):
LGTM :+1:

Issue #1628, opened 2023-03-15
Loading screen transitions


jpnurmi:
_Originally posted by @jpnurmi in https://github.com/canonical/ubuntu-desktop-installer/issues/1599#issuecomment-1469696702_

> A small shortcut was taken with the transition because a slide transition would've needed structural changes to avoid sliding titles and buttons (we'd need to move the loading spinner inside the welcome screen instead of being two separate screens).

### Expected

https://user-images.githubusercontent.com/5436686/224758885-fc9d2340-d02c-46aa-a972-e466771f3fbc.mov

### Actual

[Screencast from 2023-03-15 10-46-34.webm](https://user-images.githubusercontent.com/140617/225273404-940d2a84-7933-4272-aaf2-77c6af7b3b1d.webm)

P.S. The welcome screen below looks different because the new design was not implemented yet but this issue is specifically about the transition. :)

fabiancrx:
Hi @jpnurmi is this issue (or any other in the repo ) up for grabs. I'd like to contribute to the project.

jpnurmi:
Hi @fabiancrx, thanks for your interest!

We've just passed the UI freeze for 23.04 but we've intentionally not branched off ubuntu/lunar yet. We didn't want to get side-tracked with any features or refactoring but keep full focus on stabilizing things. Feel free to put up any PRs regardless if you don't mind any GUI changes laying around for a while until we've branched off ubuntu/lunar.

Any stabilizing fixes that don't affect the GUI in the sense that it could potentially invalidate any translations, documentation, screenshots, recordings, etc. that are already being prepared for the release would be most welcome at this point. :wink:

PR #1974, opened 2023-05-15
refactor(loading): move LoadingPage into the wizard

jpnurmi:
- Allows pre-loading the locale model from `WizardRoute.onLoad` (#1959)
- Moves out a bunch of previously hard-coded Subiquity assumptions from the app startup

Close: #1628


d-loose (approved):
Nice, LGTM :+1:

PR #1987, opened 2023-05-17
chore: add flutter_lints to the top-level pubspec

jpnurmi:
This fixes the issue that VS Code keeps complaining about:
```yaml
include: package:flutter_lints/flutter.yaml
```
in the shared [`analysis_options.yaml`](https://github.com/canonical/ubuntu-desktop-installer/blob/main/analysis_options.yaml). Now that we have a top-level `pubspec.yaml` thanks to Melos 3, we may add the `flutter_lints` dependency there as well to make tooling happy.

| Before | After |
|---|---|
| Screenshot from 2023-05-17 14-36-45 | Screenshot from 2023-05-17 14-36-25 |

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -81,6 +81,14 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "6.1.4"
+  flutter_lints:
+    dependency: "direct dev"
+    description:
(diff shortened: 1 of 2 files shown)


d-loose (approved):
LGTM :+1:

PR #1991, opened 2023-05-17
refactor(install): precache assets & preload view model

jpnurmi:
Ref: #1736


d-loose (approved):
LGTM :+1:

jpnurmi:
> This change seems to make the CI hang. Investigating...

It was because `onLoad` wouldn't get called due to using `--initial-route`.

PR #1998, opened 2023-05-22
test(screenshots): jump to initialize the timezone view model

jpnurmi:
`--initial-route` is no longer suitable and should probably be removed because it won't pre-load the respective view model and therefore things are not initialized properly.


d-loose (approved):
LGTM :+1:

PR #1999, opened 2023-05-22
refactor: remove 'part of' anti-pattern

jpnurmi:
There was a time when UDI had the `public_member_api_docs` lint enabled. As a consequence of that, this dirty little trick was used to make it possible to move widgets into separate files without having to write full-blown docs for every single member just to pass the CI. :)


CarlosNihelton (approved):
LGTM 👍🏽 

Those seem to be the last `part`/`part of` pairs.

PR #2002, opened 2023-05-23
test: migrate to yaru_test

jpnurmi:
The former `yaru_window_test` became `yaru_test` and gained some of the common finders and generic matchers (adapted from `packages/ubuntu_test`) that are helpful for testing any yaru/flutter apps.

Ref: #1891


d-loose (approved):
LGTM :+1:

PR #2004, opened 2023-05-23
test: expose helpers for testing various pages and models

jpnurmi:
This is a preparation step for adding missing widget tests for `installer.dart`. Such test requires all relevant mock models so make it possible to import `test_foo.dart` with ready-made factory methods to avoid repeating massive amounts of mocking.

Note: there's still more coming (there are several network & filesystem models) but the PR has already grown massive so I'd prefer to get the first batch in and follow up with the rest separately. :)

PR #2012, opened 2023-05-24
test: installer wizard flow

jpnurmi:
With mock view models now available outside page-specific tests, we can run the top-level wizard flow as a widget test and collect coverage (instead of executing awfully slow integration tests).

In essence, this brings back some of the coverage we lost during Riverpod/WizardRoute.onLoad migration, by testing that the init() methods of various view models get called as expected from the wizard routes.

- #2004
- #2011
- #2014
- #2016


d-loose (approved):
LGTM :+1:

PR #2015, opened 2023-05-24
test: prepare for common finders in ubuntu_test

jpnurmi:
Promote common finders from `udi/test_utils.dart` to `packages/ubuntu_test` to prepare for the final `ufp/ubuntu_test` migration which is still looking for a better solution for l10n...


d-loose (approved):
LGTM :+1:

PR #2024, opened 2023-05-29
refactor(installer): simplify app vs. background

jpnurmi:
Hide stack & details into the internal background helper widget. This is part of the effort to simplify, streamline, and de-couple the top-level app vs. wizard building blocks so that we'll be able to mix and match them.


d-loose (approved):
LGTM :+1:

PR #2026, opened 2023-05-29
chore: make WSL part of the CI aware of pubspec lockfile

CarlosNihelton:
We aim with this experiment with committing the `pubspec.lock` file into the repository. Since WSL is smaller and final, it's a good use case to try first.

Currently I'm trying not to restrict the SDK version too much. That can lead to noise. At least the CI should create a PR automatically when divergence appears.

Let's see how this goes.

---

With those changes in place PR's like #2027 will be created against the issuer's branch right away.


jpnurmi (approved):
LGTM :+1:

Thanks for experimenting with this. It will be interesting to see if it gets noisy or not. Automatically updating it in the CI is nice to reduce the likelihood of undesired local modifications. :)

.github/workflows/wsl.yaml
```diff
@@ -22,10 +22,24 @@ jobs:
...
 
+      - name: Update pubspec.lock
+        working-directory: packages\ubuntu_wsl_setup\
+        shell: bash
+        run: flutter pub upgrade
+
+      - name: Create PR
+        uses: peter-evans/create-pull-request@v5
+        with:
+          add-paths: packages/ubuntu_wsl_setup/pubspec.lock
```
jpnurmi, line 33: should we also have something like `add-paths: **/*.mocks.dart` for mocks and `add-paths: **/l10n/*.dart` for l10n?

- https://github.com/canonical/ubuntu-desktop-installer/pull/2031
- https://github.com/canonical/ubuntu-desktop-installer/pull/2032
- https://github.com/canonical/ubuntu-desktop-installer/pull/2033

CarlosNihelton: Yeah! I think we should have something like that 😜

PR #2030, opened 2023-05-30
refactor(logging): de-couple logging options from subiquity setup

jpnurmi:
Part of an ongoing effort to de-couple the app setup routines from Subiquity.


d-loose (approved):
LGTM :+1:

PR #2034, opened 2023-05-30
ci: specify path filters for mocks & l10n

jpnurmi:
We've checked in `ubuntu_wsl_setup/pubspec.yaml` in:
- #2026 

The idea was to make the CI keep the lock file up to date but it resulted in the regenerated lock file being submitted as mock and l10n:

- #2031
- #2033

vs.

- #2032

### DIFF
--- a/.github/scripts/check-outdated-files.sh
+++ b/.github/scripts/check-outdated-files.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 if [ -n "$(git status --porcelain)" ]; then
-  git diff
-  for f in $(git ls-files --modified); do
+  git diff -- ':!**/pubspec.lock'
+  for f in $(git ls-files --modified -- ':!**/pubspec.lock'); do
     echo "::warning ::$f may be outdated"
   done
   for f in $(git ls-files --others --exclude-standard); do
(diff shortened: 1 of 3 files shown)


jpnurmi:
> Should we also change the check-outdated-files script to ignore that pubspec.lock file?

I just learned that `git ls-files --exclude` only applies to untracked files, not to files checked in. :eyes:

CarlosNihelton (commented):
One minor doubt

jpnurmi:
I think the reason #2027 worked is because #2027 was created from a branch in the main repo but I'm afraid it won't work with forks that we may not have permission to push nor should we even try because it can create confusion and conflicts. :)

```
  "C:\Program Files\Git\bin\git.exe" symbolic-ref HEAD --short
  fatal: ref HEAD is not a symbolic ref
  [...]
  "C:\Program Files\Git\bin\git.exe" -c protocol.version=2 fetch --no-tags --progress --no-recurse-submodules --unshallow --force origin ci-generate:ci-generate
  fatal: couldn't find remote ref ci-generate
```

https://github.com/canonical/ubuntu-desktop-installer/actions/runs/5122122269/jobs/9210747930

PR #2035, opened 2023-05-30
refactor(installer): remove unsupported --initial-route

jpnurmi:
The hidden `--initial-route` argument which was previously used for testing is no longer supported because `WizardRoute.onLoad` would not get called. Remove handling from UDI and demote the command-line option from the shared package to WSL setup.


CarlosNihelton (approved):
LGTM 👍🏽

PR #2036, opened 2023-05-31
refactor: promote generic parts of UdiApp to the shared library

jpnurmi:
The upcoming welcome wizard needs the same styling and theming etc.


d-loose (approved):
LGTM :+1:

PR #2038, opened 2023-05-31
refactor: move Subiquity-specific --dry-run CLI argument

jpnurmi:
The desktop installer and WSL setup both have a `--dry-run` argument which is why it was included in the shared command-line parsing method but the same does not apply to non-Subiquity-based wizards...


d-loose (approved):
LGTM :+1:

PR #2041, opened 2023-05-31
refactor(wsl): add runWslSetupApp()

jpnurmi:
This shares some more logic between the two WSL entry-points and removes the use of `runWizardApp()` which is going away because there are non-Subiquity-based wizards coming and the Subiquity dependency in the shared `ubuntu_wizard` package is being removed.


CarlosNihelton (approved):
This looks great!

PR #2051, opened 2023-06-01
feat(welcome): import app skeleton

jpnurmi:
Just an empty app skeleton so far, using yaru, handy_window, ubuntu_wizard etc. and ready for inserting pages.

image


d-loose (approved):
:tada:

PR #2053, opened 2023-06-01
feat(welcome): add locale & identity

jpnurmi:
Note: auto-login is ignored for the time being.

[Screencast from 2023-06-01 17-16-49.webm](https://github.com/canonical/ubuntu-desktop-installer/assets/140617/b48daa8a-0c27-482d-be4e-7e7ec0016878)

PR #2057, opened 2023-06-02
fix(identity): use DBusInt32 in CreateUser

d-loose:


d-loose:
Just saw that this will be fixed by #2053

jpnurmi:
Verified, with this we're able to create a user :+1:

jpnurmi (approved):
LGTM :+1:

PR #2072, opened 2023-06-06
WIP: feat(keyboard_service): implement XdgKeyboardService

d-loose:
Fully implements `setKeyboard` (via `org.freedesktop.locale1.SetX11Keyboard`) and `setInputSource` (via `gsettings set org.gnome.desktop.input-sources sources ...`).

`getKeyboard` is implemented by reading the respective properties from `org.freedesktop.locale1`, however the DBus interface doesn't expose the available keymaps and variants that can be found with `localectl list-*` commands. As far as I can tell, subiquity maintains its own list of keymaps, generated by [this script](https://github.com/canonical/subiquity/blob/607eb80bdc1ff2ce9d45876603394be9fc956834/scripts/keyboard-scraper.sh).

`getKeyboardStep` isn't implemented yet - do we want/need to support keyboard detection for the XDG service as well?


jpnurmi:
> `getKeyboardStep` isn't implemented yet - do we want/need to support keyboard detection for the XDG service as well?

Maybe 'later' if anyone ever asks. :) For now, we can make it return null or add another getter that describes whether keyboard layout detection is supported by the backend so we can hide the _Detect_ button when running with XDG.

d-loose:
Sounds good!
What do we do about the list of available keyboard layouts and variants, though?

jpnurmi:
Are `/usr/share/X11/xkb/rules/base.lst` and `/usr/share/X11/xkb/rules/base.xml` the same data in different formats? Is either of those convenient to parse?

jpnurmi:
Alternatively, we could even consider stealing Subiquity's data if it's better structured than the alternatives. :grin:

packages/ubuntu_desktop_installer/lib/services/keyboard_service.dart
```diff
@@ -45,3 +52,114 @@ class SubiquityKeyboardService implements KeyboardService {
...
+
+  Future<String> _getLanguage() async {
+    await _client.connect();
+    var lang = _client.locale['LANG'] ?? 'C';
+
+    if (lang.contains('.')) {
+      lang = lang.split('.').first;
+    }
+
+    WidgetsFlutterBinding.ensureInitialized();
```
jpnurmi, line 80: any chance to remove this? could we have this in the test instead?

jpnurmi (approved):
Cool! LGTM :+1:

PR #2073, opened 2023-06-06
refactor(locale_service): use XdgLocaleClient

d-loose:
I've moved everything related to `org.freedesktop.locale1` into a [dedicated package](https://github.com/d-loose/xdg_locale.dart) ~~(which still needs to be moved into a proper namespace and published)~~.


jpnurmi (approved):
Removes a nice chunk of D-Bus boilerplate :+1:

PR #2074, opened 2023-06-07
refactor(installer): remove UdiApp in favor of WizardApp

jpnurmi:
This change removes the last remaining bits of the old monolithic `UbuntuDesktopInstallerApp` and uses the new more generic `WizardApp` directly to eliminate an unnecessary layer of indirection, property drilling, and some minor code duplication, too, because `WizardApp` takes care handling localization delegates for flavors.


d-loose (approved):
LGTM :+1:

PR #2076, opened 2023-06-07
refactor(storage): rename the page/route/model

jpnurmi:
For semi-automated installs, the list of interactive sections reported by Subiquity are [autoinstall keys](https://ubuntu.com/server/docs/install/autoinstall-reference), which are close to but not always 1:1 match with Subiquity's controller, model, and endpoint names.

As for "filesystem" vs. "storage", I'd propose we go with storage in the desktop GUI to make it a direct match with the autoinstall key and the Subiquity endpoint. Subiquity calls the respective model and controller "filesystem" but it's an implementation detail that is irrelevant to the desktop GUI.

See also:
- https://github.com/canonical/subiquity/pull/1682

Close: #1844


d-loose (approved):
LGTM :+1:

PR #2080, opened 2023-06-07
ci: specify flutter version as an environment variable

jpnurmi:
To avoid repeating it for every job.


CarlosNihelton (approved):
This is great!

PR #1527, opened 2023-03-03
snap: copy shader_lib/*.glsl artifacts for arm64

jpnurmi:
The `shader_lib` directory contains Flutter engine [artifacts](https://github.com/flutter/engine/tree/main/impeller/compiler/shader_lib) that are missing
for arm64. The files are platform-agnostic GLSL files that have been copied
from the x64 tarball.

The files can be updated to the latest stable by running `update-artifacts.sh`
or to a specific version by running `update-artifacts.sh x.y.z`. The script
downloads the given Flutter release, extracts `shader_lib` from it, and updates
the local copy.

See:
- https://github.com/canonical/ubuntu-desktop-installer/issues/1519
- https://github.com/flutter/flutter/issues/116703#issuecomment-1403956612

Fixes: #1519

PR #2083, opened 2023-06-07
snap: remove flutter's shader_lib artifacts for arm64

jpnurmi:
The artifacts were added as a temporary workaround in:
- #1527

The upstream issue was already fixed in March before the Flutter 3.10 release in May:
- https://github.com/flutter/flutter/issues/116703
- https://github.com/flutter/flutter/issues/122346


CarlosNihelton (approved):
LGTM 👍🏽 Good that upstream fixed it.

Issue #2085, opened 2023-06-08
Product name is wrong on Lenovo's products


binli:
Installed the daily image on ThinkPad, ThinkCentre or ThinkStaion, I found the hostname is wrong, and in 20.04 it could get the correct name like 'ThinkPad P17 Gen 1' while not '20SNZ5GNUS'.

$ cat /sys/devices/virtual/dmi/id/bios_vendor 
LENOVO
$ cat /sys/devices/virtual/dmi/id/product_name 
20SNZ5GNUS
$ cat /sys/devices/virtual/dmi/id/product_version 
ThinkPad P17 Gen 1

PR #2086, opened 2023-06-08
Support the correct product name for Lenovo platforms (#2085)

binli:
Installed the daily image on ThinkPad, ThinkCentre or ThinkStaion, I found the hostname is wrong, and in 20.04 it could get the correct name like 'ThinkPad P17 Gen 1' while not '20SNZ5GNUS'.

$ cat /sys/devices/virtual/dmi/id/bios_vendor
LENOVO
$ cat /sys/devices/virtual/dmi/id/product_name
20SNZ5GNUS
$ cat /sys/devices/virtual/dmi/id/product_version
ThinkPad P17 Gen 1

https://github.com/canonical/ubuntu-desktop-installer/issues/2085

### DIFF
--- a/packages/ubuntu_desktop_installer/lib/pages/identity/identity_model.dart
+++ b/packages/ubuntu_desktop_installer/lib/pages/identity/identity_model.dart
@@ -216,12 +216,24 @@ class IdentityModel extends PropertyStreamNotifier {
 /// The DMI product name file. Available for testing.
 @visibleForTesting
 const kDMIProductNameFile = '/sys/devices/virtual/dmi/id/product_name';
+@visibleForTesting
+const kDMIProductVersionFile = '/sys/devices/virtual/dmi/id/product_version';
 
-Future<String> _readProductName() {
-  return File(kDMIProductNameFile)
+Future<String> _readDmiFile(String path) {
+  return File(path)
       .readAsString()
       .then((value) => value.trim().sanitize())
       .catchError((_) => '');
+}
+
+Future<String> _readProductName() async {
+  /// Lenovo uses product_version as product name
+  /// and it's empty on HP and DELL
+  var productName = await _readDmiFile(kDMIProductVersionFile);
+  if (productName.isEmpty) {
+    productName = await _readDmiFile(kDMIProductNameFile);
+  }
+  return productName;
 }
 
 extension _IdentityDescription on IdentityData {


jpnurmi:
Thanks for the fix. Looks like `/sys/devices/virtual/dmi/id/product_version` is empty on my Dell system. Let's try to gather some more data on how `product_name` and `product_version` look on various systems. Perhaps we could simply default to `product_version` and fall back to `product_name` if empty? This way we wouldn't need any specific vendor checks. :)

binli:
> Thanks for the fix. Looks like `/sys/devices/virtual/dmi/id/product_version` is empty on my Dell system. Let's try to gather some more data on how `product_name` and `product_version` look on various systems. Perhaps we could simply default to `product_version` and fall back to `product_name` if empty? This way we wouldn't need any specific vendor checks. :)

Yes, I also checked some HP and DELL platforms, the product_version is empty, but I'm not sure if any other brands.

And I checked the ubiquity code, it uses the simliar way to get manufacturer first, then check the vendor. Anyway, I agree with your idea, I've updated a new version of change.

            if 'lenovo' in manufacturer or 'ibm' in manufacturer:
                key = 'system-version'
            else:   
                key = 'system-product-name'
            proc = subprocess.Popen(   
                ['dmidecode', '--quiet', '--string', key],
                stdout=subprocess.PIPE,          
                universal_newlines=True)

packages/ubuntu_desktop_installer/lib/pages/identity/identity_model.dart
```diff
@@ -216,14 +216,25 @@ class IdentityModel extends PropertyStreamNotifier {
...
-Future<String> _readProductName() {
-  return File(kDMIProductNameFile)
+Future<String> _readDmiFile(String path) {
+  return File(path)
       .readAsString()
       .then((value) => value.trim().sanitize())
       .catchError((_) => '');
 }
 
+String _readProductName() async {
```
jpnurmi, line 229:
```suggestion
Future<String> _readProductName() async {
```

packages/ubuntu_desktop_installer/lib/pages/identity/identity_model.dart
```diff
@@ -168,7 +168,7 @@ class IdentityModel extends PropertyStreamNotifier {
     _hostname.value ??= identity.hostname.orIfEmpty(null);
     _username.value ??= identity.username.orIfEmpty(null);
     log.info('Loaded identity: ${identity.description}');
-    _productName.value = await _readProductName();
+    _productName.value = _readProductName();
```
jpnurmi, line 171:
```suggestion
    _productName.value = await _readProductName();
```

packages/ubuntu_desktop_installer/lib/pages/identity/identity_model.dart
```diff
@@ -216,14 +216,25 @@ class IdentityModel extends PropertyStreamNotifier {
...
       .then((value) => value.trim().sanitize())
       .catchError((_) => '');
 }
 
+String _readProductName() async {
+  /// Lenovo uses product_version as product name
+  /// and it's empty on HP and DELL
+  var productName = await _readDmiFile(kDMIProductVersionFile);
+  if (productVersion.isEmpty)
+    productName = await _readDmiFile(kDMIProductNameFile);
```
jpnurmi, line 234:
```suggestion
  if (productVersion.isEmpty) {
    productName = await _readDmiFile(kDMIProductNameFile);
  }
```

packages/ubuntu_desktop_installer/lib/pages/identity/identity_model.dart
```diff
@@ -216,14 +216,26 @@ class IdentityModel extends PropertyStreamNotifier {
...
       .readAsString()
       .then((value) => value.trim().sanitize())
       .catchError((_) => '');
 }
 
+Future<String> _readProductName() async {
+  /// Lenovo uses product_version as product name
+  /// and it's empty on HP and DELL
+  var productName = await _readDmiFile(kDMIProductVersionFile);
+  if (productVersion.isEmpty) {
```
jpnurmi, line 233: oops there's still one variable name mismatch here :) productVersion -> productName

jpnurmi: other than this, LGTM :+1:

jpnurmi (approved):
Thanks, LGTM :+1:

binli:
@jpnurmi thanks for your suggestions, I've merged them, and rebased all the commits into one final commit. Please help review it again. Thanks!

packages/ubuntu_desktop_installer/lib/pages/identity/identity_model.dart
binli: Yes, I found it, and fixed, :)

jpnurmi:
Thanks! The fix should be available in the next daily-live build.

binli:
Tried Ubuntu 23.10 "Mantic Minotaur" - Daily amd64 (20230612) , this issue is fixed.

jpnurmi:
Great, thanks for the update!

PR #2088, opened 2023-06-08
refactor(active-directory): de-couple from navigation

jpnurmi:
Previously, Active Directory would not be used by default but it would be up to the user to interactively toggle it on the identity page, the preceding page in manual installation flow:

| Identity | Active Directory |
|---|---|
| image | image |

The selection would be passed as an argument for the router to make a decision on whether to navigate to the Active Directory page or jump over it. This approach was fine in manual installation flows but doesn't work when doing semi-automated installations because then it's up to Subiquity to decide whether the Active Directory page is shown. Therefore, this PR moves the logic to the service level and removes the navigation argument so that the Active Directory page can be shown independently of the identity page.


d-loose (approved):
LGTM :+1:

PR #2096, opened 2023-06-12
feat(keyboard): add canDetectLayout

d-loose:
Split off from #2072:

Makes keyboard layout detection an optional feature of the `KeyboardModel` and hides the 'Detect' button in the UI if layout detection is not supported


jpnurmi (approved):
Yay, thanks for splitting it off! :) LGTM :+1:

PR #2101, opened 2023-06-13
refactor(confirm): loosen coupling with subiquity

jpnurmi:
Call `InstallerService.start()` instead of `SubiquityClient.confirm('/dev/tty1')` (:woozy_face:) and export relevant storage types from the storage service to get rid of the `subiquity_client.dart` import in `confirm_model.dart`.


d-loose (approved):
LGTM :+1:

PR #2102, opened 2023-06-13
WIP: refactor: move xdg services to ubuntu_welcome

d-loose:


jpnurmi:
Could we call them `xdg_foo_service.dart` to avoid having three instances of `foo_service.dart` later? It will be easier to find the file you're looking for. :)

jpnurmi (approved):
Thanks! LGTM :+1:

PR #2103, opened 2023-06-13
refactor: use stock button labels from ubuntu_localizations

jpnurmi:
With fewer dependencies on `AppLocalizations`, it will be easier to modularize later.


jpnurmi:
I'm curious to find out how Weblate behaves if we remove entries from the main `app_en.arb` template but leave them to others. I'm hoping Weblate will detect this and offer to remove the dangling entries...

d-loose (approved):
LGTM :+1:

PR #2106, opened 2023-06-14
refactor(test): make better use of ubuntu_test helpers for buttons

jpnurmi:
`tapNext()` and friends would fail if they were ambiguous so we can leave out the repetitive `findsOneWidget` checks.


d-loose (approved):
LGTM :+1:

PR #2108, opened 2023-06-14
fix(ubuntu_wizard): show spinner during onActivated

d-loose:
This makes sure `WizardButton` shows the loading indicator while awaiting the result of the `onActivated` callback.


jpnurmi (approved):
LGTM :+1:
LGTM :+1:

PR #2110, opened 2023-06-14
fix(locale_page): await applyLocale before proceeding

d-loose:
This ensures that the `LocalePage` finishes setting the locale before proceeding to the next page. In the welcome wizard this is necessary because the `KeyboardPage` uses the system locale to show a localized list of keyboard layouts.


jpnurmi (approved):
LGTM :+1:

PR #2112, opened 2023-06-14
feat(ubuntu_wizard): add Wizard[Route]Data class

d-loose:
Wraps the `userData` of the `Wizard`s and their `WizardRoutes` in a class, so that we can add more data besides the page index, e.g. whether a route is allowed to navigate back.


jpnurmi (approved):
LGTM :+1:

PR #2113, opened 2023-06-14
refactor(ubuntu_test): switch to ubuntu-flutter-plugins/ubuntu_test

jpnurmi:
There's no such thing as `WidgetTester.ulang` in the new [ubuntu_test](https://github.com/canonical/ubuntu-flutter-plugins/tree/main/packages/ubuntu_test) so it was added to `test_utils` to make it possible to do the switch without hassle.


d-loose (approved):
LGTM :+1:

PR #2114, opened 2023-06-14
feat: control page navigation options in routes

d-loose:
Moves the control over whether or not the next/previous buttons are enabled from the pages to the `WizardRoute` using the `userData` property.


d-loose:
Oh wait.. pages should be able to at least control whether we can proceed or not

jpnurmi (commented):
The patch looks great but I'm wondering about the important things like naming because camel-case with short words makes it looks like AlTeRnAtInG cApS... :) Could they be called `hasNext` and `hasPrevious`? because they basically override `Wizard.hasNext` and `Wizard.hasPrevious`?

d-loose:
:+1: yEs, gOOd PoINt!

jpnurmi (approved):
LGTM :+1:

Issue #1893, opened 2023-04-28
Merge guided partitioning pages


jpnurmi:
We have two different pages related to guided partitioning:

| "Select guided storage" | "Install alongside" |
|---|---|
| ![image](https://user-images.githubusercontent.com/140617/235098813-22f4dbd6-895d-4666-b49b-9fdf755792f0.png) | ![image](https://user-images.githubusercontent.com/140617/235098838-9bf664a8-9eeb-438c-893f-8fd427329af5.png) |

The one on the left is missing a note about existing partitions that would be erased, and the link to manual partitioning. I think these two should in fact be implemented as one "guided partitioning" page that offers all "use gap" _and_ "resize" types of guided storage targets and the part in the middle swaps between an interactive vs. non-interactive view.

PR #2116, opened 2023-06-15
refactor(storage): rename guided storage pages/routes/models

jpnurmi:
This PR continues where #2094 left off and renames the guided storage pages, routes, and models in a similar way that was already done for the manual storage page. Ideally, there would be just one "guided storage" page (#1893) but the merge is not trivial so "guided reformat" and "guided resize" pages are the next best options we can have for now.

See also:
- #1844


jpnurmi:
The end-goal is to allow configuring nested routes:

```yaml
routes:
- locale
- keyboard
- storage/guided ### <== currently, only "storage" is allowed
- timezone
- identity
```

d-loose (approved):
LGTM :+1:

PR #2117, opened 2023-06-15
refactor(storage): remove non-existent "reinstall" option

jpnurmi:
There's no such feature in Subiquity. The radio button has been commented out from the GUI for 1.5 years. Maintaining the non-existent option in the view model causes unnecessary confusion while refactoring. Adding such a radio button will be trivial if/when there's support in the backend.


d-loose (approved):
LGTM :+1:

PR #2118, opened 2023-06-15
fix(welcome): override subiquity_client/test deps while they exist

jpnurmi:
Override the correct local dependencies from the Git submodule until these undesired dependencies go away after the upcoming modularization, to avoid breaking due to API changes in subiquity_client.

See also:
- https://github.com/canonical/subiquity_client.dart/pull/140
- https://github.com/canonical/ubuntu-desktop-installer/actions/runs/5276064005/jobs/9542284049?pr=2117


d-loose (approved):
I was just about to complain :D
LGTM :+1:

PR #2121, opened 2023-06-15
refactor(storage): rename page/route/model

jpnurmi:
The storage page is the entry point of the storage sub-wizard and is shown by default when the storage route is enabled. Later on, it will be made possible to configure specific storage sub-routes, too.

See also:
- #1844


d-loose (approved):
LGTM :+1:

PR #2127, opened 2023-06-16
feat(wizard): add proper support for nested wizards

jpnurmi:
This PR makes nested wizards work out of the box by automatically looking up the root wizard when navigating past the boundary of a sub-wizard.

The problem with passing `root: true` for accessing the root wizard from a sub-wizard is that it creates an assumption on the wizard structure. We don't want individual pages to hard-code such assumptions about what comes next because it prevents the respective routes from being configurable.

Instead, we need a declarative set of independent onLoad-guarded routes that can be wired up with route configuration as we did with the top-level wizard.


d-loose (approved):
LGTM :+1:

PR #2133, opened 2023-06-19
feat: add SessionService and implementations

### LINKED ISSUE #2111: Session service
Should have methods:
- `reboot`
- `shutdown`

Two implementations, one based on `SubiquityClient` and another on [ubuntu_session](https://pub.dev/packages/ubuntu_session).

d-loose:
Close #2111 
UDENG-742


packages/ubuntu_welcome/lib/services/xdg_session_service.dart
```diff
@@ -0,0 +1,19 @@
...
+import 'package:ubuntu_session/ubuntu_session.dart';
+
+class XdgSessionService implements SessionService {
+  XdgSessionService([@visibleForTesting UbuntuSession? session])
+      : _session = session ?? UbuntuSession();
+
+  final UbuntuSession _session;
+  @override
+  Future<void> reboot() {
+    return _session.reboot();
```
jpnurmi, line 12: does UbuntuSession have an argument that describes whether the reboot is "immediate" as subiquity calls it?

d-loose: No, this looks like some [subiquity-related parameter](https://github.com/canonical/subiquity/blob/fa29b9088bac7c55b9dc0588c63a18b18455fd78/subiquity/server/controllers/shutdown.py#L66-L72)

jpnurmi: the installer sends "immediate" reboot requests from the "turn off rst" and and "turn off bitlocker" pages. what if we used the systemd session manager that would allow passing interactive=false?

d-loose: According to the [man page](https://www.freedesktop.org/software/systemd/man/org.freedesktop.login1.html#Security):
> The interactive boolean parameters can be used to control whether polkit should interactively ask the user for authentication credentials if required.

So, setting `interactive=false` would just stop polkit from trying to get the user authentication to reboot/shutdown if necessary

jpnurmi: oh, not an interactive confirmation but an interactive authentication? the problem is that when called from the "turn off xxx" pages before installation, subiquity requires the immediate flag or else it won't reboot until installation is finished

d-loose: Ah, I see - if it's needed we could add the parameter and have it do nothing in the xdg case.

jpnurmi (approved):
LGTM :+1:

jpnurmi:
The Jira link already closed the ticket but will you follow up with a PR to take it into use?

PR #2135, opened 2023-06-19
refactor: use SessionService to reboot

d-loose:
Uses the new `SessionService` rather than `SubiquityClient` to reboot in ubuntu_desktop_installer.
WSL setup still uses subiquity directly, since it's not yet using any other abstract services either.

UDENG-742


jpnurmi (approved):
LGTM :+1:

PR #2136, opened 2023-06-20
feat(identity): add auto-login support to XdgIdentityService

jpnurmi:
Close: #2097

PR #2138, opened 2023-06-20
feat(identity): set account password when using the XDG backend

jpnurmi:
To avoid a newly created user account being locked due to not having a password. This implies that the password must be passed unencrypted to the service layer because Subiquity expects an encrypted password whereas XDG AccountsService encrypts it on the fly in SetPassword().

Required for:
- #2136


d-loose (approved):
LGTM :+1:

PR #2139, opened 2023-06-20
fix(storage): ignore disallowed storage targets

jpnurmi:
There's a new feature in Subiquity that it not only reports allowed guided storage targets, but also includes information about disallowed storage targets. Make sure to only include allowed storage targets when determining what type of guided storage targets are available until we implement proper support for disallowed targets.

There are several issues with the screenshot test but this is one reason for random failures recently.


d-loose (approved):
LGTM :+1:

PR #2144, opened 2023-06-21
refactor(screenshot-test): machine config & storage type

jpnurmi:
- Use a suitable machine config that leads to the desired sub-route.
- Select the appropriate storage type instead of blindly jumping to a sub-route and assuming it exists.


d-loose (approved):
LGTM :+1:

PR #2153, opened 2023-06-21
Move LVM to the main guided partitioning screen

### LINKED ISSUE #1310: Move LVM to the main guided partitioning screen
ubuntu-desktop-installer   0+git.7b30974e    734

Since ZFS and ZFS + encryption has been removed, there is only one advanced feature left (LVM). The LVM and optional encryption should be moved to the main guided partitioning page.

dejvizelo:
Fixes #1310 

image


jpnurmi:
Hey @dejvizelo, thanks for taking the time to contribute! :tada:

Personally, I fully agree that the LVM and encryption settings shouldn't be hidden in "Advanced features". The reason such a change didn't get done by now is that we'd like to take the opportunity to redesign the whole page.

The installer backend aka. Subiquity is able to use LVM and encryption even when installing alongside an existing OS, not just when erasing the entire disk. Therefore, we'd like to hold off on the PR until the design team has had an opportunity to take a closer look.

PR #2156, opened 2023-06-22
refactor(flavor): provide ubuntu_flavor

jpnurmi:
- makes use of the [ubuntu_flavor](https://github.com/ubuntu-flutter-community/ubuntu_flavor) package that can detect 8/10 of the official Ubuntu flavors out of the box
- replaces the clumsy inherited widget with a simpler top-level provider because we don't need different information in different parts of the tree
- removes yet another arbitrary non-wizard widget from the shared ubuntu_wizard package


d-loose (approved):
LGTM :+1:

PR #2157, opened 2023-06-22
refactor: move common utils out of ubuntu_wizard

jpnurmi:
The name "utils" is bad but in this case, it actually describes pretty well what this random bunch is. Ideally, such a dumpster package shouldn't exist and these things would find natural homes. Many of them are a bit small for being separate standalone packages, though.


d-loose (approved):
LGTM :+1:

PR #2158, opened 2023-06-22
refactor: eliminate unnecessary dependencies to InheritedLocale

jpnurmi:
The plan is to move `InheritedLocale` into `ubuntu-wsl-setup` and use a locale provider in the installer to get rid of the `ubuntu_utils` dependency from `ubuntu_wizard`.

See also:
- #2157


d-loose (approved):
LGTM :+1:

PR #2159, opened 2023-06-22
refactor(locale): demote inherited locale to wsl & provide in udi

jpnurmi:
Demote `InheritedLocale` to `ubuntu-wsl-setup` and use a locale provider in the installer to get rid of the `ubuntu_utils` dependency from `ubuntu_wizard`.

See also:
- #2157


d-loose (approved):
LGTM :+1:

Issue #2143, opened 2023-06-20
Screenshot test occasionally hangs or fails


jpnurmi:
The screenshot test calls `runInstallerApp()` directly without using `tester.runApp()` which means it hangs sometimes just like the other integration tests did before. However, before we can use `tester.runApp()`, the existing errors and exceptions that we haven't cared about would have to be fixed or prevented in one way or another because after switching to `tester.runApp()` the errors will no longer be swallowed.

- #2104

jpnurmi:
There are even some errors that suggest that we're deleting the .subiquity data directory while subiquity continues to run:

```
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following SubiquityException was thrown running a test (but after the test had completed):
setSource(ubuntu-desktop) returned error 500
Traceback (most recent call last):
  File
"/home/jpnurmi/Projects/canonical/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/common/api/server.py",
line 160, in handler
    result = await implementation(**args)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File
"/home/jpnurmi/Projects/canonical/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/server/controllers/source.py",
line 153, in POST
    await self.configured()
  File
"/home/jpnurmi/Projects/canonical/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/server/controllers/source.py",
line 145, in configured
    await super().configured()
  File
"/home/jpnurmi/Projects/canonical/ubuntu-desktop-installer/packages/subiquity_client/subiquity/subiquity/server/controller.py",
line 108, in configured
    with open(self.app.state_path('states', self.name), 'w') as fp:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '.subiquity/run/subiquity/states/Source'

When the exception was thrown, this was the stack:
#0      SubiquityClient._receive (package:subiquity_client/src/client.dart:71:7)
2
<asynchronous suspension>
(elided one frame from package:stack_trace)
════════════════════════════════════════════════════════════════════════════════════════════════════
```

Such failures may be related to:
- https://github.com/ubuntu/yaru_window.dart/issues/33

PR #2172, opened 2023-06-27
fix(screenshot_test): allow tests to settle before tearing down

jpnurmi:
Most `testFooPage()` helpers end with `tester.tapNext()` or similar. Letting the pipeline settle before tearing down tests helps with some of the obscure errors occasionally happening during `screenshot_test`.

Ref: #2143


d-loose (approved):
LGTM :+1:

PR #2173, opened 2023-06-27
fix(screenshot_test): simulate window close at tear down

jpnurmi:
Unlike other integration tests that run through full installation flows until the window is closed, the screenshot test navigates into specific screens just to take screenshots of them. As such, there's nothing triggering the app to stop Subiquity etc. unless we simulate window close.

This should prevent a series of obscure screenshot test errors because Subiquity is no longer left running while the data directory would be, at the same time, cleaned up in the beginning of each test to ensure a clean environment.

Ref: #2143


d-loose (approved):
LGTM :+1:

PR #2175, opened 2023-06-28
refactor(theme): split ThemeService out of DesktopService

jpnurmi:
The two will be used in different contexts. While one is about inhibiting potentially dangerous actions (auto-mounting, power saving, ...) during the installation phase, the other is about applying the desired look'n'feel during the welcome phase. Currently, the theme service only supports theme brightness but later on, it will gain support for accent colors too.


d-loose (approved):
LGTM :+1:

PR #2176, opened 2023-06-28
fix(ad): catch potential exception for long-lasting join result

jpnurmi:
The AD join result is requested ahead of time from the AD page. The result is returned later at some point during the installation after the necessary tools have been installed etc.


d-loose (approved):
LGTM :+1:

PR #2177, opened 2023-06-28
feat(theme): add view model

jpnurmi:
As a rule of thumb, we should never access services directly from the widget tree but nobody ever bothered adding a view model for the theme page due to its simplicity. However, the complexity will increase a bit after adding support for accent colors. As a bonus, consistently following the same structure for all pages helps with testing, too.


d-loose (approved):
LGTM :+1:

PR #2183, opened 2023-06-29
fix(screenshot_test): use tester.runApp()

jpnurmi:
I'm sure there are still some occasional failures to tackle but this will make sure the CI will no longer swallow errors and hang, at least. :)

Close: #2143


d-loose (approved):
LGTM :+1:

PR #2188, opened 2023-06-30
Don't stage directly libllvm, it's pulled in via library depends

seb128:
The version currently listed is wrong for core22 and unused

It's not the first time we forget to update the version when switching to a new core serie. Currently 15 is being used by the software stack but we are still pulling 12 for nothing. Instead of having a version manually specify let's get it through the dri components depends. 

A local build shows that libllbvm15 is being pulled it as expected, the removal of the extra libllvm makes the snap 20mb smaller


kenvandine (approved):
Good catch!

PR #2189, opened 2023-06-30
refactor(utils): remove color utils

jpnurmi:
Use yaru color extensions instead.

- https://github.com/ubuntu/yaru.dart/pull/354


d-loose (approved):
LGTM :+1:

PR #2191, opened 2023-06-30
l10n: remove unused {disk} argument

jpnurmi:
We changed from `{disk}{partition}` to `{sysname}` to fix the formatting of NVME partitions ("nvme0n11" vs. "nvme0n1p1") but weren't able to change translations right before the release.

See also:
- #1821


d-loose (approved):
LGTM :+1:

jpnurmi:
I hope Weblate won't explode... :)

PR #2210, opened 2023-07-04
chore: prepare ubuntu_provision package

jpnurmi:
Just the usual package boilerplate with common dependencies and empty localizations etc. prepared. Individual pages will be moved in separate steps.

Ref: #2194


d-loose (approved):
LGTM :+1:

PR #2212, opened 2023-07-05
refactor(test): replace expectPage() with localization-agnostic helpers

jpnurmi:
The `expectPage()` helper was doing too much. It would pump until a certain type of page was available and test that a localized page title was found with a callback that assumed `AppLocalizations`. Split it into two helper methods, one for pumping until a certain type of page is available, and another for finding a localized page title in a way that does not assume `AppLocalizations` but allows passing `ProvisionLocalizations`, `TimezoneLocalizations`, or any other localization type.

Ref: #2194


d-loose (approved):
LGTM :+1:

PR #2219, opened 2023-07-05
refactor(provision): move flavor provider

jpnurmi:
Used by several pages, some of which will be moved to ubuntu_provision.

Ref: #2194


d-loose (approved):
LGTM :+1:

PR #2220, opened 2023-07-06
refactor(l10n): match translation entries with page names

jpnurmi:
Ref: #2194


jpnurmi:
Brb, need to lock Weblate to avoid conflicts.

PR #2225, opened 2023-07-06
refactor(test): port away from problematic WidgetTester.lang

jpnurmi:
The approach doesn't work when there are multiple localizations (`ubuntu_desktop_installer`/`ubuntu_welcome` vs. `ubuntu_provision`) in the game. Get rid of the helper and use the usual `FooLocalizations.of()` instead. It's a bit verbose but easy to understand and straightforward to change from `AppLocalizations` to `AnyOtherLocalizations` while moving pages.


d-loose (approved):
LGTM :+1:

PR #2227, opened 2023-07-07
refactor(provision): move theme

jpnurmi:
Ref: #2194


packages/ubuntu_provision/lib/src/theme/theme_page.dart
```diff
@@ -112,7 +112,7 @@ class _ThemeOptionCard extends StatelessWidget {
                 padding: EdgeInsets.zero,
                 child: ClipRRect(
                   borderRadius: BorderRadius.circular(8),
-                  child: Image.asset(assetName),
+                  child: Image.asset(assetName, package: 'ubuntu_provision'),
```
jpnurmi, line 115: i think we just lost asset lookup support for ubuntu flavors because now that the assets are in a separate ubuntu_provision package, we must explicitly specify the package name... but i think we can fix this later in ProxyAssetBundle when it's time to look at flavor support again

d-loose (approved):
LGTM :+1:

PR #2228, opened 2023-07-07
fix(loading): auto-dispose the view model

jpnurmi:
The wizard never returns to the loading view so it's not useful to keep the view model alive.


d-loose (approved):
LGTM :+1:

PR #2229, opened 2023-07-07
refactor(provision): move mascot avatar

jpnurmi:
Used by the locale page which is on its way to `ubuntu_provision`. If the image size wasn't specifically targeted for the installer, we could even promote it to `ubuntu_widgets`...

Ref: #2194


d-loose (approved):
LGTM :+1:

PR #2230, opened 2023-07-10
refactor(provision): move common services

jpnurmi:
This PR promotes common service interfaces and generic non-Subiquity based service implementations from `ubuntu_desktop_installer` to `ubuntu_provision`.

Ref: #2194


d-loose (approved):
LGTM :+1:

PR #2234, opened 2023-07-10
chore: drop pubspec overrides

jpnurmi:
The dependency overrides were added to make it possible to use `ubuntu_desktop_installer` as a Git dependency and `subiquity_client` as a path dependency for maximum flexibility in [ubuntu-flavor-installer](https://github.com/canonical/ubuntu-flavor-installer) but keeping track of compatible sha1's turned into a maintenance nightmare. Use `ubuntu_desktop_installer` as a path dependency through a recursive Git submodule instead, which can be automatically updated with the help of Dependabot or Renovate and also automatically brings in a compatible version of Subiquity.

Ref: #2194
Ref: canonical/ubuntu-flavor-installer#13


d-loose (approved):
LGTM :+1:

PR #2243, opened 2023-07-12
refactor(installer): demote config to post-inst service

jpnurmi:
While the current "config" service does indeed write a config file, it's a very minimal Bash-compatible config file specifically for passing values to the post-inst scripts. Rename it to "post-install" service instead to make room for a more generic TOML/YAML config service that will be used for configuring pages for any provisioning phase.

Ref: #2194


d-loose (approved):
LGTM :+1:

Issue #2252, opened 2023-07-18
mantic: broken automatic partition setup


jbicha:
- Ubuntu Desktop 23.10 "mantic" 2023-07-18

Test Case
-------------
I am using Ubuntu 23.10 as my host. In GNOME Boxes, I created a 23.10 guest using today's ISO.

I clicked all the default options until I get to this confirmation screen with missing details in the devices and partitions sections.

![Screenshot from 2023-07-18 13-52-36](https://github.com/canonical/ubuntu-desktop-installer/assets/876259/5966cb5d-5960-4d50-94c0-f37802a9e489)

Afterwards, I clicked Install and finished the rest of the dialog but it looked like the installer couldn't actually install anything so I was just left at the slideshow stage. No other errors or glitches were shown.

jpnurmi:
Hey @jbicha, do you still have access to the system? Any chance to get the logs from `/var/log/installer`?

jpnurmi:
Never mind, I'm just too used to bumping up the VM specs! :man_facepalming: I was able to reproduce the issue with the default 25.0 GiB storage.

jpnurmi:
We need to add handling for the `TOO_SMALL` reason introduced in canonical/subiquity#1694.
```
2023-07-18 18:48:14.733330 DEBUG subiquity_client: GET http://localhost/storage/v2/guided?wait=true
2023-07-18 18:48:14.734688 DEBUG subiquity_client: ==> getGuidedStorageV2(true) {"status": "DONE", "error_report": null, "configured": null, "targets": [{"disk_id": "disk-vda", "allowed": [], "disallowed": [{"capability": "DIRECT", "reason": "TOO_SMALL", "message": null}, {"capability": "LVM", "reason": "TOO_SMALL", "message": null}, {"capability": "LVM_LUKS", "reason": "TOO_SMALL", "message": null}], "$type": "GuidedStorageTargetReformat"}, {"allowed": ["MANUAL"], "disallowed": [], "$type": "GuidedStorageTargetManual"}]}
```

jpnurmi:
With the default 25 GiB disk gnome-boxes offers for Ubuntu 23.10, Subiquity refuses guided reformat (`TOO_SMALL`) even though manual installation with the simplest possible partition layout installs fine.

- `GET /storage/v2/orig_config`
  ```json
[...]

PR #2255, opened 2023-07-19
fix(installer): respect subiquity's guided storage targets

jpnurmi:
Don't offer options that Subiquity doesn't allow. This doesn't fix the problem that we're currently not able to perform a guided reformat of a 25GiB disk (gnome-boxes default for 23.10) but at least such options are no longer offered that Subiquity explicitly says are not allowed.

Ref: #2252


d-loose (approved):
LGTM :+1:

PR #2282, opened 2023-08-04
Fix typo in shown command to retrieve recovery key

ogayot:


jpnurmi (approved):
Thanks! LGTM :+1: 

We can safely ignore the CI failures because the experimental changes in the canary branch were supposed to be short-lived and were never properly run through the CI.

P.S. We're in the process of bringing TPM support to the main branch.

PR #2289, opened 2023-08-08
Merge remote-tracking branch 'upstream/main' into ubuntu-core-desktop

jpnurmi:


jpnurmi:
@kenvandine Is the `ubuntu-core-desktop` branch still used? This merge brings all the latest changes, including the minimal dd-target handling, into that branch.

The only difference to the main branch is that this branch comes with a `ubuntu-desktop-installer.core-desktop` snap command alias that equals to running `ubuntu-desktop-installer --pages=locale,keyboard,source,storage`.

If we provide a provisioning config on the file system instead, we can switch to the main branch and drop the `.core-desktop` alias.
```yaml
# /usr/share/ubuntu-provision.yaml
bootstrap:
  pages: [locale, keyboard, source, storage]
```

kenvandine:
> @kenvandine Is the `ubuntu-core-desktop` branch still used? This merge brings all the latest changes, including the minimal dd-target handling, into that branch.
> 
> The only difference to the main branch is that this branch comes with a `ubuntu-desktop-installer.core-desktop` snap command alias that equals to running `ubuntu-desktop-installer --pages=locale,keyboard,source,storage`.
> 
> If we provide a provisioning config on the file system instead, we can switch to the main branch and drop the `.core-desktop` alias.
> 
> ```yaml
> # /usr/share/ubuntu-provision.yaml
> bootstrap:
>   pages: [locale, keyboard, source, storage]
> ```

We never got that far, I think it would be better to just try to use the same installer for both and lay that config file down during image generation

PR #2295, opened 2023-08-10
ci: clean up

jpnurmi:
- Drop the labeler because there's not much to label here anymore because the development has moved to ubuntu-desktop-provision.
- Throw away the broken semantic PR title check


d-loose (approved):
LGTM :+1:

